PR #4538, opened 2012-12-14
Integration of RFC86: scale dependent token replacements

tbonfort:
c.f. http://mapserver.org/development/rfc/ms-rfc-86.html


havatv:
I could not find anything about this in the layer documentation.  Should it be documented?

PR #4825, opened 2013-12-14
Implement msPostGISLayerGetExtent (#3585)

szekerest:


szekerest:
@pedros007 I wonder how your layer definition is looking like in this case

pedros007:
My data statement for the layer is maybe a bit verbose?  My schema has a `base_tiles` table with a geometry.  I have a projects table composed of many tiles.  The tiles table is a join table to base_tiles.  The WFS request that causes Postgres to report `ERROR: syntax error at or near "0" at character 29 STATEMENT: SELECT ST_Extent(geom) FROM 0` is:

http://localhost/mapserv?service=WFS&VERSION=1.1.0&REQUEST=getfeature&typename=tiles&srsname=EPSG:4326&OUTPUTFORMAT=geojson&project_id==1&partition_quadkey=='021320320'"

The error goes away if I add `      "ows_extent"            "-180 -90 180 90"` to my METADATA section.

The quadkeys referenced in the query are [Bing Maps tiling scheme](https://msdn.microsoft.com/en-us/library/bb259689.aspx).  Partitions are level 9, tiles are level 12. 

```LAYER
    NAME tiles
    TYPE POLYGON
    METADATA
      "ows_title"             "tiles"
      "ows_description"       "Tile Extents"
      "gml_types"             "auto"
      "gml_include_items"     "all"
      "geojson_include_items" "all"
    END
    STATUS OFF
        INCLUDE "postgis.inc.map"
        DATA "geom FROM (
          SELECT
            base_tiles.geom,
            base_tiles.id as base_tile_id,
            base_tiles.quadkey,
            substring(base_tiles.quadkey from 0 for 10) as partition_quadkey,
            base_tiles.x,
            base_tiles.y,
            base_tiles.level,
            tiles.project_id,
            tiles.id as tile_id,
            tiles.partition_id as partition_id,
            tile_stacks.status
          FROM base_tiles
            INNER JOIN tiles
            ON base_tiles.id = tiles.base_tile_id
            INNER JOIN tile_stacks
            ON tile_stacks.tile_id = tiles.id
          WHERE ST_Intersects(base_tiles.geom, !BOX!) and project_id %project_id% and substring(base_tiles.quadkey from 0 for 10) %partition_quadkey%
          ) AS subquery USING UNIQUE base_tile_id USING SRID=4326"
        VALIDATION
          # %project_id% must be one or more integers
          'project_id' '^=[0-9]+$'
          'default_project_id' 'IS NOT NULL'
          'partition_quadkey' "^='[0-3]{9}'$"
          'default_partition_quadkey' "like '%'"
        END
END
```

szekerest:
It seems that the table name is not correctly identified with this complex expression (the innermost "from 0" is considered to define the table name). I'll see how that could be work arounded.

pedros007:
Thanks, @szekerest .  I opened #5365 to track this.

PR #4857, opened 2014-01-30
RFC108 - Heatmap/Interpolation Layer creation

tbonfort:
This issue tracks the development of RFC-108: http://mapserver.org/development/rfc/ms-rfc-108.html


Yogurt4:
Hi. I was just wondering if this feature could be used to plot a data grid. (And if not, whether it could be developed.)
We have various distribution maps. However, instead of colouring by density of scattered points, we have measurement points on a rectangular grid. The values can come from soil chemistry, air pollution, humidity, etc.
What we need is an interpolation between the measurement points, like the isothermal or isobar plot in the weather forecast.

Or, imagine the SRTM (Surface Radar Terrain Model) data from NASA. It is also a rectangular grid of elevation data. Using this data one could draw an elevation map in MapServer.

PR #4922, opened 2014-05-07
Sld outlinewidth support

yjacolin:
Adding support for outlinewidth and take care of different default value between width (1.0) and outlinewidth (0.0).


yjacolin:
this is not an issue but a PR waiting toi be finalized (yes 1 year after its start :/)

Issue #4974, opened 2014-08-19
Tracking issue for RFC 91


sdlime:
This is a tracking issue for merging RFC 91 branches into master.

http://mapserver.org/fr/development/rfc/ms-rfc-91.html

tbonfort:
We can close this one now, yay!

PR #4997, opened 2014-09-18
RFC91: fix expressions/filters set from SLD (#4974)

tbonfort:
- surround spatial expressions with () so they are treated as expressions
- no need to translate expressions for DB backends, RFC91 takes care of that for us


tbonfort:
RFC91 is now fixed, closing

PR #5002, opened 2014-09-23
UTF grid compression

fblackburn:
Here's some part I added to the UTF grid implementation to reduce the size of JSON file returned by map server. It remove data and key entry for char index that are not appearing in the grid itself. I know it is late for map server 7.0 but it would be a nice to have.


tbonfort:
applied to master in f152773

Issue #4569, opened 2013-01-21
chart runtime set PROCESSING "CHART_SIZE_RANGE


stetou:
I asked on the mailing list, but since there is no answer I suppose it is not possible...
I use this for my pie charts
PROCESSING "CHART_SIZE_RANGE = itemname minsize maxsize minval maxval" 
We should be able to modify  minval maxval at runtime so we always have nice proportional pie chart. With dynamic data, we can't hard code those values in mapfile

By example, the data of my pie chart come from a postgis view. I use variable substitution (year per example) in FILTER. I would need then to adjust minval and maxval. Programmatically I can get those values and I would like to be able to modify PROCESSING "CHART_SIZE_RANGE

PR #5005, opened 2014-09-26
Add substitutions on processing options (#4569)

thhomas:


tbonfort:
added along with #4877, thanks

Issue #4980, opened 2014-08-27
WMS KML generates incorrect output for MULTILINESTRING


phopfgartner:
When exporting a MULTILINE, the generated KML is not valid.

Basically, starting with the second line in the ulti, all LineString Element are collected as childs in a LineString Element, like:

```
        <MultiGeometry>
          <LineString>
            <coordinates>
        0.00000000,0.00000000
        897.00000000,0.00000000
        897.00000000,290.00000000
        </coordinates>
          </LineString>
          <LineString>
            <LineString>
              <coordinates>
        0.00000000,580.00000000
        897.00000000,580.00000000
        897.00000000,586.00000000
        </coordinates>
            </LineString>
          </LineString>
        </MultiGeometry>
```

This is with MapServer 6.4.1 on CentOS 6.

Peter

PR #5017, opened 2014-10-15
fix issue #4980

phopfgartner:
This change solves the above issue.

I had to take the code from KmlRenderer::getGeomParentNode(), breaking the locality of the attribute GeomNode.
I'd say, that the method is not really clear in it's intention, since the method name suggests, that the parent is returned, bit code returns the child, so this was the best I was able to achieve without major changes.


tbonfort:
@phopfgartner is this ready to merge ?

Issue #5029, opened 2014-11-05
Legend and scalebar are not scaling when the resolution changes


szekerest:
When setting the resolution to a different value than defresolution the embedded scalebar and legend should also scale to provide the same look on the map at different resolutions. At least the following parameters should take care of the resolution factor:

For the scalebar: width, height, font size, horizontal and vertical margins, stroke width
For the legend: font size, keysize, keyspacing, horizontal and vertical margins

Not all of these parameters can be set externally (ie. in mapfiles)

PR #5035, opened 2014-11-08
Fix scaling the legend and scalebar when the resolution is changing (#5029)

szekerest:


tbonfort:
The autotests are failing with this patch:
```
# renderers => legend_resolution_point_4.png
compare: image widths or heights differ `../expected/legend_resolution_point_4.png' @ error/compare.c/CompareImageCommand/957.
# renderers => legend_resolution_point_2.png
compare: image widths or heights differ `../expected/legend_resolution_point_2.png' @ error/compare.c/CompareImageCommand/957.
```
it would seem that the images are created with a size different than what was requested.

szekerest:
@tbonfort I noticed the failure, but I might think that it could be a correct result however. Images at a higher resolution should be proportionally larger in order to keep the same look.

tbonfort:
Let me disagree. If the user requests a 50x50 image he should get back a 50x50 image. If he wants a similar hi-res image, he should request a 100x100 image and a higher dpi.

szekerest:
I understand what you mention, but in this case we are talking about the keysize (what is expected), but the autotest check is about the legend image size, which is not the same as the keysize. For example the actual image size for legend_resolution_point_4.png seems to be 199x90 and that won't necessary be the same after the patch, since we also want keysize/keyspacing/hMargin/vMargin to grow proportionally along with the resolution factor. 
It is also a theoretical question whether we would want to let the user calculate all these modified parameters externally, or let mapserver to do that so setting the resolution factor should behave something like the symbolscale setting currently. That means the specified keysize would be the same as the pixel size only if the resolution/defresolution =1
But anyway this patch should be refined indeed.

PR #5053, opened 2014-12-18
Add missing max/minScaleDenom to xld/t files

seuros:


tbonfort:
rebased and applied in f79aacc39fcd302f92 , thanks @seuros

PR #5057, opened 2015-01-06
Handle PROJ_LIB environment variable.

### LINKED ISSUE #4930: PHP MapScript ignores PROJ_LIB environment variable
Loading a MapFile with PHP MapScript ignores Apache SetEnv PROJ_LIB value.

$map = new MapObj($file);
The MapObject has a setconfigoption method, unfortunatly it cannot be called before the constructor. 

A solution would be to add a loadMapFile method to the MapObj which would allow to 

$map = new MapObj();
$projLib = getenv("PROJ_LIB");
$map.setConfigOption("PROJ_LIB", $projLib);
$map.loadMapFile($file);

mkofahl:
According to the docs (http://mapserver.org/errors.html), PROJ_LIB may be set as environment variable, too. This pull request makes it working and fixes #4930.

### DIFF
--- a/mapproject.c
+++ b/mapproject.c
@@ -1246,6 +1246,18 @@ static const char *msProjFinder( const char *filename)
 #endif /* def USE_PROJ */
 
 /************************************************************************/
+/*                       msProjLibInitFromEnv()                         */
+/************************************************************************/
+void msProjLibInitFromEnv()
+{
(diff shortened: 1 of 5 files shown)


tbonfort:
applied in e17e69ba1567869e , thanks @mkofahl

PR #5058, opened 2015-01-15
add JSONP support

aeichner:
This is a small change that creates a JSONP wrapper around OGR output. Uses a format option "JSONP"  that works when "simple" output format is used. Should be considered a workaround until OGR has it's own datasource creation option.

### DIFF
--- a/mapfile.c
+++ b/mapfile.c
@@ -6856,6 +6856,7 @@ void msApplyDefaultSubstitutions(mapObj *map)
   /* output formats (#3751) */
   for(i=0; i<map->numoutputformats; i++) {
     applyOutputFormatDefaultSubstitutions(map->outputformatlist[i], "filename", &(map->web.validation));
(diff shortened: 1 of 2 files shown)


tbonfort:
rebased and applied in 76c8c66 , thanks @aeichner . Could you please add an autotest for this one?

rouault:
Thi is a bit of a hack (it can/should only be used with OGR GeoJSON driver and if FORM=simple), although admitedly a simple&efficient one. 
What about the documentation : should we advertize it  or hide it ?

aeichner:
I agree that this isn't a clean solution. The OGR's JSON driver should get an appropriate datasource creation option instead. If officially documented it should be noted that users must be prepared that it might change when a solution reaches OGR.

aeichner:
@tbonfort A basic test has been added https://github.com/mapserver/msautotest/pull/29

tbonfort:
Needs to be updated to support cgi as well as fastcgi, c.f. mapserver/msautotest#29 and #4858

PR #5065, opened 2015-02-10
WFS needs to project Map Extent when setting the WFS/OWS SRS

msmitherdc:
WFS Sets the default SRS to the first OWS SRS, however, it doesn't project the default MapExt when it does so.


msmitherdc:
@tbonfort this is ready to merge to master

dmorissette:
Fixed in  	cf065f3

PR #5067, opened 2015-02-11
remove the ArcSDE C API driver from MapServer

hobu:
As motioned on the mapserver-dev list, we're removing ArcSDE native support for MapServer 7.0 due to it not being widely used, it being hard to get the SDK, and it has not been actively tested in a number of releases.

http://lists.osgeo.org/pipermail/mapserver-dev/2015-February/014349.html

(All the whitespace noise is due to my having a plugin turned on that removes end-of-line whitespace. If this is unsatisfactory, I can issue a new PR).


tbonfort:
Thanks @hobu . If it's not too hard on you, I'd prefer a version without the whitespace noise, as it tends to mess up versioning annotations.

Issue #5001, opened 2014-09-20
Make specifying a native filter more explicit in 7.0


sdlime:
@tbonfort, @rouault, @dmorissette, I'd like to suggest a breaking change for 7.0 as a result of the RFC 91 work. Presently if you use a database driver and it sees filter->string set it takes that as SQL. From setting via a mapfile the expression is of type MS_STRING. For RFC 91 I added an explicit native_string structure member and changed db drivers to only use that if not NULL. Works ok...

However, digging into the SLD and WMS time code I found that native expressions were being created as type MS_EXPRESSION (e.g. MapServer expression syntax ('[column]' eq 'foo')). That was a problem since those were the types of expressions we want to be able to translate to native SQL. Ugh... I had to force the filter type to MS_STRING in those instances.

It would be nice to be more explicit. OGR uses a "WHERE " qualifier and I think it would make sense to do the same universally. That what the load expression function can strip off the WHERE, set native_string and a specific type (MS_NATIVE_EXPRESSION). It'll be more convenient for MapScript users.

This is a relatively simple change but it can break some stuff: mapfiles that use database drivers and FILTERs (there's little reason to do so) and MapScript scripts that use setFilter() with database drivers. I think those should be relatively limited.

I know the RFC has been a pain in the ass but now is the time...

Steve

sdlime:
Has anyone been able to consider this for other drivers (e.g. Oracle, SQL Server, etc...)? I'd like to know if this work can be done before the sprint. The work on the PostGIS driver really makes things clearer for users and simplifies the query code.

Steve

tbonfort:
Support for NATIVE_FILTER in oracle added in 1842327

PR #5068, opened 2015-02-11
Remove the ArcSDE C API driver from MapServer

hobu:
This is a second attempt at removing the SDE driver, without the whitespace noise #5067 

Note that maplexer.c will need to be regenerated, as this seemed to be really noisy on my diff.


sdlime:
We should remove this *after* the changes to mapsde.c related to issue #5001 are added - just in case we add it back. --Steve

PR #5069, opened 2015-02-12
Fix NULL pointer dereference

aeichner:
Fix NULL pointer dereference in mapdraw.c:msPrepareImage() in case an image could not be initialized.


tbonfort:
applied in 1fdcb04 , thanks

Issue #5077, opened 2015-03-05
No Methods in MapScript to set colo(u)r RGBA


ejn:
Currently the colorObj does not provide any methods capable of setting the alpha value - only setRGB().

PHP- and SWIG MapScript should be extended with a setRGBA() method.

Current workaround is to use updateFromString on the parent object.

tbonfort:
please go ahead and fix, seems pretty straightforward from what I've checked in the swig files. Please also update the fromHex() method so it also accepts #rrggbbaa forms of hex strings.

ejn:
Also missing is:
in SWIG a toHex with alpha channel (lets call it toHexAlpha to avoid changing the signature of the existing toHex method)
in PHP toHex, setHex and toHexAlpha

I'll throw all those into a PR coming up (fairly) shortly

tbonfort:
@ejn: can we close this one now #5078 is done?

ejn:
@tbonfort - yes, should all be sorted now.

PR #5078, opened 2015-03-05
MapScript colorObj SWIG/PHP alignment and methods with alpha

ejn:
Changes in the SWIG bindings:
- add the methods setRGBA() and toHexWithAlpha()
- extend the method setHex() to accept #rrggbbaa hex values

Changes in the PHP bindings:
- add the methods setRGBA() and toHexWithAlpha()
- adds the methods toHex() and setHex() which were already in the SWIG bindings

Fix for #5077 

Still TODO:
- PR to update documentation
- Tests?

### DIFF
--- a/mapscript/php/color.c
+++ b/mapscript/php/color.c
@@ -48,6 +48,23 @@ ZEND_ARG_INFO(0, green)
 ZEND_ARG_INFO(0, blue)
 ZEND_END_ARG_INFO()
 
+ZEND_BEGIN_ARG_INFO_EX(color_setRGBA_args, 0, 0, 4)
+ZEND_ARG_INFO(0, red)
+ZEND_ARG_INFO(0, green)
+ZEND_ARG_INFO(0, blue)
+ZEND_ARG_INFO(0, alpha)
+ZEND_END_ARG_INFO()
+
+ZEND_BEGIN_ARG_INFO_EX(color_setHex_args, 0, 0, 1)
+ZEND_ARG_INFO(0, rgba)
+ZEND_END_ARG_INFO()
+
+ZEND_BEGIN_ARG_INFO_EX(color_toHex_args, 0, 0, 0)
+ZEND_END_ARG_INFO()
+
+ZEND_BEGIN_ARG_INFO_EX(color_toHexWithAlpha_args, 0, 0, 0)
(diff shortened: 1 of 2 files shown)


ejn:
Will check the test failures and update the PR tomorrow.

ejn:
It looks like the failing tests after the updated PR are in gdal and the renderers, neither of which should be at all affected by this PR.

@tbonfort Can you make sense of that? I've not got a working test environment here at the minute which isn't make life easier.

tbonfort:
no worries on the failed test, the remote wms server we're using must be down or changed... 

just wondering about the need to add setRGBA from fromHexWithAlpha, can't we just extend the current setRGB (with a default alpha=255) and fromHex and not add new signatures?

ejn:
@tbonfort Thanks for checking the test failure - I was reasonably sure it wasn't my fault!

There is no fromHexWithAlpha - the existing fromHex (in the case of SWIG, it wasn't in PHP at all) just accepts either a 3- or 4-byte hex string.

For toHex() and toHexWithAlpha() I guess toHex() - up until now only in SWIG - could be extended with an optional flag argument withAlpha (default false for backwards compatibility). Same with setRGB()/setRGBA().

I'm +/-0 on optional arguments or separate methods - I find a separate method more explicit, but optional arguments are probably cleaner.

Not sure what the usual practice has been up to now in Mapscript? Either way I can update the PR if necessary.

tbonfort:
I'd prefer just setRGB, toHex and fromHex, with setRGB taking a default fourth alpha argument, fromHex being extended to parse #rrggbb and #rrggbbaa , and toHex spitting out #rrggbb or #rrggbbaa depending on wether opacity is full or not. We might also want to refactor the code at https://github.com/mapserver/mapserver/blob/branch-7-0/mapfile.c#L405 into a new function that could be called by mapscripts.

ejn:
@tbonfort I rebased this PR to the current branch-7-0 to kick Travis into life.

Since the extended method signatures are documented (mapserver/docs#114 got merged before this PR) I guess this one now counts as a bugfix too? ;-O

ejn:
@tbonfort Thanks! Please also merge mapserver/msautotest/#30

PR #5085, opened 2015-03-31
Removed double free in WFS layer.

mkofahl:
When using an WFS layer, the params object will be freed twice, in msPrepareWFSLayerRequest() and msPrepareWFSLayerRequest().


tbonfort:
applied in 9da77c11f7e6

PR #5086, opened 2015-04-02
Segfault when using ttf symbols with symbolscalednom and minsize=0

tbonfort:
MapServer will crash at very small scales, when a ttf symbol is used together with symbolscalednom, but without minsize, which is default 0.

```
MAP
		SIZE 1000 500
		EXTENT -90 -90 90 90
		PROJECTION
				"init=epsg:4326"
		END

		FONTSET 'glyph.fonts'

		SYMBOL
				NAME "test"
				TYPE TRUETYPE
				FONT "arial"
				CHARACTER '&#97;'
		END

		LAYER
				TYPE LINE
				STATUS DEFAULT

				FEATURE
						POINTS
								0 0
								20 20
						END
				END

				NAME 'ttflayer'
				STATUS DEFAULT
				PROJECTION
						'init=epsg:4326'
				END

				SYMBOLSCALEDENOM 100

				CLASS
						STYLE
								SYMBOL "test"
								WIDTH 2
#								MINSIZE 1
						END
				END
		END
END
```

### DIFF
--- a/maprendering.c
+++ b/maprendering.c
@@ -247,8 +247,8 @@ imageObj *getTile(imageObj *img, symbolObj *symbol,  symbolStyleObj *s, int widt
           if(UNLIKELY(!face)) { status = MS_FAILURE; break; }
           msUTF8ToUniChar(symbol->character, &unicode);
           unicode = msGetGlyphIndex(face,unicode);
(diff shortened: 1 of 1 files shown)


mkofahl:
Thank you for fixing this @tbonfort

Issue #5081, opened 2015-03-19
GetCapabilities request is extremely slow when “OWS_LAYER_GROUP” is used


aabhayas:
GetCapabilities request is extremely slow when the Mapfile has nested layers using “OWS_LAYER_GROUP”. This bug exists from the version 6.2 of Mapserver.

Following link has more details of this bug.

http://osgeo-org.1560.x6.nabble.com/Re-Getcapabilities-extremely-slow-in-version-6-4-1-td5191374.html

I have uploaded two map files and two map part files which can be used to reproduce this bug.

http://osgeo-org.1560.x6.nabble.com/file/n5194857/layer_group.map
http://osgeo-org.1560.x6.nabble.com/file/n5194857/layer_group.part
http://osgeo-org.1560.x6.nabble.com/file/n5194857/no_layer_group.map
http://osgeo-org.1560.x6.nabble.com/file/n5194857/no_layer_group.part

Map file layer_group.map uses layer_group.map.part which has LAYER_GROUPS defined.

Map file no_layer_group.map uses no_layer_group.map.part which has no LAYER_GROUPS defined.

I use the following command line in my server to time the getcapabilities requests.

time /web/cgi-bin/mapserv -nh "QUERY_STRING=map=/web/ms4w/apps/tutorial/htdocs/layer_group.map&SERVICE=WMS&VERSION=1.3.0&REQUEST=GetCapabilities"

time /web/cgi-bin/mapserv -nh "QUERY_STRING=map=/web/ms4w/apps/tutorial/htdocs/no_layer_group.map&SERVICE=WMS&VERSION=1.3.0&REQUEST=GetCapabilities"

tbonfort:
fixed with #5094

tomkralidis:
Is/should this be backported to 6.4.x?

aabhayas:
Hi

I think we should backport this to 6.4

Regards
Anuruddha Abhayasinghe
[Description: Description: Description: Description: http://web.bom.gov.au/eiab/publishing/images/bureau_logo.gif]
Geospatial Data and Services
Delivery Section / Build Branch
Bureau of Meteorology
GPO Box 1289 Melbourne VIC 3001
Level 6, 700 Collins Street, Docklands VIC 3008
Tel: +61 3 9669 4589 | aabhayas@bom.gov.au<mailto:aabhayas@bom.gov.au>
www.bom.gov.au<http://www.bom.gov.au/>

tomkralidis:
[...]

PR #5094, opened 2015-04-29
This patch is to fix the issue "GetCapabilities request is extremely slo...

aabhayas:
...w when “OWS_LAYER_GROUP” is used"

https://github.com/mapserver/mapserver/issues/5081.

Changed the code to create a unique set of groups and then compare them with layer names to avoid multiple passes through layer names.

### DIFF
--- a/mapwms.c
+++ b/mapwms.c
@@ -326,9 +326,15 @@ int msWMSApplyTime(mapObj *map, int version, char *time, char *wms_exception_for
 */
 void msWMSPrepareNestedGroups(mapObj* map, int nVersion, char*** nestedGroups, int* numNestedGroups, int* isUsedInNestedGroup)
 {
-  int i, j, k;
+  int i, k;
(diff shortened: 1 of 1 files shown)


Schpidi:
thx for the patch. I'll take a look hopefully soon...

aperi2007:
@tomkralidis 
@tbonfort 

Hi,
sorry , but I'm not understand well the github issue interface. So don't understand if the patch for issue 5094 was added to mapserver 70 or not.
I read in github that the 24/july 2015 this patch was removed from 7.0 release.

tbonfort:
I'm not sure why @tomkralidis removed the 7.0 milestone from this issue. From what I see, this patch is present in 7.0.0 and 6.4.2

tbonfort:
updated previous comment: 6.4.3 is the first 6.4.x release to include this patch

PR #5098, opened 2015-05-11
Fix segfault in PHP MapScript implementation of saveQuery

mkofahl:
Calling `saveQuery` from php MapScript with result parameter set to `MS_TRUE` or `MS_FALSE`, php may segfault due to illegal assignmet of a variable parsed as long to an int variable.


sdlime:
Pulled into master as well... --Steve

PR #5099, opened 2015-05-12
A freed geometry should be set to null in msGEOSFreeGeometry()

mkofahl:
`msGEOSFreeGeometry` checks whether the object to be freed exists, but does not set the freed object to null. If `msGEOSFreeGeometry` is - accidently - called twice on the same object, Mapserver will crash.


tbonfort:
seems good to me. Is this ready for merging ?

sdlime:
I would say yes...

mkofahl:
yes

sdlime:
Applied in both master and branch-7-0. Thanks!

Issue #5104, opened 2015-06-01
GetFeature with 'date' filter doesn't work on MS7


ger-benjamin:
In MS7 (7.0.0~beta1) I've tried to do some GetFeatures queries with a "date" type but it crashes systematically.
The sended SQL query (to Postgres 9.1, PostGIS 1.5) is not well formed :-/

**The datatype**
````
...
<element name="date_validation" minOccurs="0" type="date"/>
...
````

**The GetFeature request**

````
http://myserv.com/mapserv?&VERSION=1.0.0&SERVICE=WFS&REQUEST=GetFeature&TYPENAME=my_layer&Filter=<Filter><PropertyIsLessThan><PropertyName>date_validation</PropertyName><Literal>2015-05-28T00:00:00</Literal></PropertyIsLessThan></Filter>
````

**The response**
````
<ServiceExceptionReport xmlns="http://www.opengis.net/ogc" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.2.0" xsi:schemaLocation="http://www.opengis.net/ogc http://schemas.opengis.net/wfs/1.0.0/OGC-exception.xsd">
<ServiceException code="NoApplicableCode" locator="mapserv">
msWFSGetFeature(): WFS server error. FLTApplyFilterToLayer() failed msPostGISLayerWhichShapes(): Query error. Error executing query. Check server logs
</ServiceException>
</ServiceExceptionReport>
````

**The corresponding Postgres logs**
````
...WHERE (("date_validation" <= 2015.000000-5.000000-29.0000000.0000000.0000000.000000))
````

I've tested the same request on a 6.0.1-3.2 MapServer and it works correctly.
Perhaps this issue is the same as the #5102

**Comment by sdlime** [comment]
Any chance of getting a small test case? I don't think this is exactly #5102. Probably a conversion within the PostGIS driver or perhaps something specific to WFS.

**Comment by rouault** [comment]
Steve, I can reproduce with one of the msautotest mapfile :
```
[...]

PR #5105, opened 2015-06-01
FLTGetBinaryComparisonCommonExpression(): escape date literals with b…

rouault:
…ackquote (#5104)


rouault:
@sdlime Does that look good to you ?

sdlime:
Yes, I think so. I was wondering if this might be the case. Perhaps I was a bit overzealous when I removed some driver-specific code from mapogcfiltercommon.c. --Steve

rouault:
@sdlime Not sure what the policy is regarding branches : after committing into branch-7-0, should I "git merge branch-7-0" into master ?

sdlime:
You're way better at git than I... ;-) Can you just cherry pick this modification without merging the whole branch?

PR #5113, opened 2015-06-28
handle the case where include_dirs is set in distutils.cfg

nijikon:
This fixes build with homebrew. It should probably be also applied to 6.4 and 7.0.


tbonfort:
applied to branch-6-2, branch-6-4, branch-7-0 and master

PR #5116, opened 2015-07-08
Typo fixes

sebastic:
While updating the Debian package for 7.0.0-beta2, the lintian QA tool reported a couple of typos.


sebastic:
The 'consistent' typo is fixed in msautotest in https://github.com/mapserver/msautotest/pull/32

PR #5118, opened 2015-07-10
Bump SOVERSION to 2 to account for removed symbols since 6.4.

sebastic:
Quite a number of symbols that were introduced in MapServer 6.2.1 are no longer availble in 7.0.0-beta2, to account for these removed symbols to library SOVERSION needs to be incremented.


sdlime:
@tbonfort, ok to merge?

sebastic:
As mentioned on mapserver-dev, this change was merged (https://github.com/mapserver/mapserver/commit/5154eb8dbf63de9b952504c2b9cc47ec65e43532)

PR #5119, opened 2015-07-10
Install headers and export cmake package for external projects.

sebastic:
This patch is included in the mapserver Debian package to allow mapcache
to build with mapserver support.

All headers are installed in usr/include/mapserver, and the cmake files
in usr/share/mapserver/cmake.


tbonfort:
@sebastic OK in principle for integrating the patch, however can you check/fix why the continuous build is failing with it applied?

sebastic:
I've added the missing CMake configuration files to fix this incomplete PR.

tbonfort:
applied to branch-7-0 in 2549b439c325b , thanks @sebastic

rouault:
@sebastic I've finally managed to enable the AppVeyor buils and launched a build of master and it errors out with : 
```
CMake Error at CMakeLists.txt:977 (INSTALL):
install Library TARGETS given no DESTINATION!
 ```
( https://ci.appveyor.com/project/MapServer/mapserver/build/job/h53c6gkmn6v5pktu )

Any idea how to fix that ? (I assume this is linked to the CMake changes of this ticket)

The previous build (one month ago) succeeded : https://ci.appveyor.com/project/MapServer/mapserver/build/1.0.1/job/7vk28s03xxy0qsle

sebastic:
Looks like Windows needs an ARCHIVE install target too (see: [[CMake] Install path problem on MinGW](http://www.cmake.org/pipermail/cmake/2008-September/023876.html))

Can you try [commit f86ccbd](https://github.com/sebastic/mapserver/commit/f86ccbd000a0672d9e29756db2a57b4d9290c5c3) from my fork?

rouault:
@sebastic Can you submit it as a PR ? It should trigger AppVeyor

sebastic:
I can, and will, after I get the results from the Debian package build in which I've include that change for testing. If installing the archive target in the shlib component has adverse effects on the Linux builds I'll need to change the component first.

rouault:
@sebastic Oh I just figured out I can myself create the PR : https://github.com/mapserver/mapserver/pull/5164 (perhaps because I've been given admin rights on mapserver repo, I dunno)

rouault:
@sebastic Looks good on Windows. https://ci.appveyor.com/project/MapServer/mapserver/build/1.0.4

sebastic:
Looking good on Linux too, although I noticed that the headers are now installed in `/usr/lib` too, that's a separate issue to fix.

sebastic:
The fix for the header installation in `/usr/lib` is available in #5165, it also includes the ARCHIVE target change from #5164.

rouault:
@sebastic Do you know which branches this applies ? master only or also 7.0 ?

sebastic:
My PR is for master, but the change should be backported to branch-7-0 too.

Both master & branch-7-0 have the same issue (installing the headers in `/usr/lib` too).

Issue #4942, opened 2014-06-18
Label outline gets scaled with changing map scale


ejn:
According to the [docs for OUTLINEWIDTH in STYLE](http://mapserver.org/de/mapfile/style.html#index-39);

>  The thickness of the outline will not depend on the scale.

The [corresponding docs in LABEL](http://mapserver.org/mapfile/label.html#index-20) don't specify any behaviour in this regard, but the behaviour as specified in STYLE would seem a reasonable expectation.

Currently however, the outline for labels gets scaled up with the map scale - see images below - with seemingly no way to prevent this.

It doesn't appear to be the fault of the font (c.f #4774) since I've tested it with various fonts (including major ones such as Arial).

I've put up the self-contained minimalist test case used for generating the images at https://github.com/faegi/testcases/tree/master/mapserver-outlinewidth - tested using a recent build from master.

~1:500 000: ![test500](https://cloud.githubusercontent.com/assets/1106824/3310339/48cc3a40-f6b2-11e3-8e0a-d757f431ce59.png)
~1:50 000: ![test50](https://cloud.githubusercontent.com/assets/1106824/3310338/4604ad9c-f6b2-11e3-8c18-bd595804dc2b.png)
~1:5 000: ![test5](https://cloud.githubusercontent.com/assets/1106824/3310326/23522090-f6b2-11e3-8c5a-b0dea4b3d918.png)

ejn:
Appears to only occur when SYMBOLSCALEDENOM is set on the layer.

Does anyone have time to check this before the 7.0 release? I had a quick look through but couldn't see where this was going wrong.

tbonfort:
untested patch attached here, can you give it a spin?

```diff
diff --git a/maprendering.c b/maprendering.c
index bf96a3f..f7c5c50 100644
--- a/maprendering.c
+++ b/maprendering.c
@@ -1023,7 +1023,7 @@ int msDrawTextSymbol(mapObj *map, imageObj *image, pointObj labelPnt, textSymbol
     c = &ts->label->color;
   if(MS_VALID_COLOR(ts->label->outlinecolor))
     oc = &ts->label->outlinecolor;
-  ow = ts->label->outlinewidth * ts->scalefactor;
+  ow = ts->label->outlinewidth * (ts->textpath->glyph_size / ts->label->size);
   if(!renderer->renderGlyphs) return MS_FAILURE;
   return renderer->renderGlyphs(image,ts->textpath,c,oc,ow);
```

[...]

PR #5107, opened 2015-07-13
Label outline not changing when OUTLINEWIDTH value changed

tbonfort:
I have a mapfile that has several layers that use SIZEUNITS METERS for styling the layers.  Specific to this issue is that the layers are labelled.  The text seems to be sized properly in meters, but the outline on the text is not.  It seems to always be the same no matter what value I set OUTLINEWIDTH to.

Sample mapfile: https://www.dropbox.com/s/rg9bph3awkmujr2/outlinewidth.map?dl=0

This seems to have started happening after the fix for #4942.

I checked out 3e7ecf7d1ca3512c6c66f30ebed739cdd484a841, and the problem did not happen, but checking out d0043f27538ff70c7911fa22707d4839ccfeae80 makes it happen.  It also happens with an up-to-date branch-7-0

### DIFF
--- a/maprendering.c
+++ b/maprendering.c
@@ -1023,7 +1023,8 @@ int msDrawTextSymbol(mapObj *map, imageObj *image, pointObj labelPnt, textSymbol
     c = &ts->label->color;
   if(MS_VALID_COLOR(ts->label->outlinecolor))
     oc = &ts->label->outlinecolor;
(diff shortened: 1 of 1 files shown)


tbonfort:
@cnieman can you confirm all is OK with this patch?

cnieman:
I just did a quick test and it looks OK.

It's not exactly the same as my build before the fix for #4942, but I assume that things are a bit wrong in that build for the outline widths.

PR #5123, opened 2015-07-20
Allow access to layer [metadata_*] tags in legacy headers & footers

ejn:
Due to these parts being processed in BROWSE mode, no layer metadata
could be accessed - this commit allows access to layer metadata tags
so long as there is a current (non-null) resultlayer, regardless of the
mode in which processLine() is being run

### DIFF
--- a/maptemplate.c
+++ b/maptemplate.c
@@ -3923,18 +3923,8 @@ static char *processLine(mapservObj *mapserv, char *instr, FILE *stream, int mod
     outstr = msReplaceSubstring(outstr, "[lrn]", repstr);
     outstr = msReplaceSubstring(outstr, "[cl]", mapserv->resultlayer->name); /* current layer name */
     /* if(resultlayer->description) outstr = msReplaceSubstring(outstr, "[cd]", resultlayer->description); */ /* current layer description */
-  }
 
-  if(mode != QUERY) {
-    if(processResultSetTag(mapserv, &outstr, stream) != MS_SUCCESS) {
-      msFree(outstr);
-      return(NULL);
-    }
-  }
-
-  if(mode == QUERY) { /* return shape and/or values  */
-
-    /* allow layer metadata access in a query template, within the context of a query no layer name is necessary */
+    /* allow layer metadata access when there is a current result layer (implicitly a query template) */
     if(&(mapserv->resultlayer->metadata) && strstr(outstr, "[metadata_")) {
       for(i=0; i<MS_HASHSIZE; i++) {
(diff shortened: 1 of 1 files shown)


ejn:
@tbonfort Is this one too late for the MS7 release? It's sort of a bugfix - from the documentation it was never clear that the [metadata_XXX] tags only worked in the main template and not the headers and footers, and functionally I see no reason why they shouldn't work there.

tbonfort:
I'll let @sdlime comment, I don't know this part of the code...

sdlime:
Been a long time since I've look through those sections. Seems benign but @ejn how does the resultlayer get set if not in query mode? --Steve

ejn:
@sdlime It's all a bit counter-intuitive: for the legacy templates then in msReturnNestedTemplateQuery() the function msReturnPage() and from there processLines() gets called 3 times for each layer - for the main template with mode=QUERY (maptemplate.c#L4372), but for the header and footer templates with mode=BROWSE (maptemplate.c#L4349, maptemplate.c#L4384). The calls for the web template are also performed with mode=BROWSE.

My intuition was to set all the calls at the layer level to QUERY mode, but on closer inspection everything except the body are done with mode=BROWSE because it makes no sense to attempt to replace shape-level tags outside of the template body and calling processLines() with mode=QUERY is basically used as a shorthand to say "replace feature level tags".

Since map->resultlayer gets set at the top of the layer loop (maptemplate.c#L4326) and nulled at the end (maptemplate.c#L4388) then it will only be set whilst processing queries (legacy header/template/footer), but not for the web header and footer, and not when not processing queries. I'm therefore pretty confident that the layer metadata replacement block isn't going to get called at inappropriate times in its new place - and in my opinion it wasn't getting called when it was appropriate up until now.

Sorry - long explanation. TL;DR is that although we're processing a query, processLine() gets called with mode=BROWSE and we therefore need another way to tell that layer-level tags should be processed.

sdlime:
Ok, this rings a bell. There are so many query modes I was a bit lazy when that stuff was written. Actually we probably don't need a mode at all and could just check the map->resultlayer value but that can happen later if necessary. @tbonfort I have no objections to this pull request being added to the 7.0 release. It's pretty innocuous... --Steve

ejn:
@sdlime Thanks for looking at this. I think some kind of mode flag is necessary (or at least helpful) to work out whether to go into the feature-level block or not. Whether that mode is called BROWSE / QUERY or something else is another question. Once you work out what's going on then it's OK as it is now.

Issue #5124, opened 2015-07-22
Bug in Oracle case insensitive equals filter


msmitherdc:
I'll get a pull request right away for this to get it into 7.0

sdlime:
Is this specific to Oracle or should we look at PostgreSQL, etc… as well?

msmitherdc:
Its specific to Oracle

tbonfort:
fixed in f67fbaf6a0270

PR #5125, opened 2015-07-22
Bug in Oracle case insensitive equals filter - Issue 5124

msmitherdc:
https://github.com/mapserver/mapserver/issues/5124

Make sure IEQ comparison wraps the comparison string with start (^) and end ($) to ensure exact match


tbonfort:
applied to branch-7-0 in f67fbaf6a0270

PR #5129, opened 2015-08-01
Bugfix: mapfile.xsd does not validate itself.

JuergenWeichand:
Bugfix: maxScaleDenom/minScaleDenom have been placed in the wrong xs:element, so the XSD did not validate itself.


tbonfort:
applied to branch-7-0 in f916cad6c6827b086, thanks

PR #5133, opened 2015-08-04
Fix most remaining warnings raised on Travis

rouault:


rouault:
All above commits squahsed as 0fc850b71423323fbc163531395bb34509df6744

PR #5145, opened 2015-08-19
Fix memory leaks on error

gogglesguy:
Fix a couple of memory leaks on error conditions


rouault:
cherry picked into master as f103b421dee8051cb91cbf50d90587fccf03e626

tbonfort:
applied to 7.0 in 75d092a8eab8b, thanks

PR #5146, opened 2015-08-20
Fix memory leaks in mapdrawgdal

gogglesguy:
This fixes various memory leaks in msDrawRasterLayerGDAL, most notably the rasterBufferObject
for the mask (if either msDrawRasterLayerGDAL_RawMode or msDrawRasterLayerGDAL_RawMode is called).
Also cleansup any leaks whenever an error condition has been encountered.

### DIFF
--- a/mapdrawgdal.c
+++ b/mapdrawgdal.c
@@ -58,13 +58,13 @@ LoadGDALImages( GDALDatasetH hDS, int band_numbers[4], int band_count,
                 int *pnNoData1, int *pnNoData2, int *pnNoData3 );
 static int
 msDrawRasterLayerGDAL_RawMode(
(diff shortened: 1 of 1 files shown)


rouault:
A less invasive approach would probably to have a 

```
rasterBufferObj s_mask_rb;
if( .... ) {
      [...]
      mask_rb = &s_mask_rb;
      memset(mask_rb, 0, sizeof(s_mask_rb);
      [...]
}
```

That way you don't have to care about deallocating it at every return site.

gogglesguy:
Yes it would. I assumed the original authors knew what they were doing and really wanted it allocated.

rouault:
Looking at the structure, it is rather small, so I think it is OK to allocate it on the stack.

gogglesguy:
Alright, will come up with an alternative patch.

PR #5148, opened 2015-08-21
Fix buffer overflow / memory leak in mapuvraster.c

gogglesguy:
Using uvlinfo->height instead of uvlinfo->width will either get you a buffer overflow or memory leak depending on whether width or height is larger.

### DIFF
--- a/mapuvraster.c
+++ b/mapuvraster.c
@@ -91,6 +91,7 @@ static int msUVRASTERLayerInitItemInfo(layerObj *layer)
   uvRasterLayerInfo *uvlinfo = (uvRasterLayerInfo *) layer->layerinfo;
   int   i;
   int *itemindexes;
+  int failed=0;
 
   if (layer->numitems == 0)
     return MS_SUCCESS;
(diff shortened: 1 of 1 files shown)


sdlime:
@dmorissette, something you can look at?

gogglesguy:
So checking for -1 for itemindexes on a initialized array was wrong. Returning from the item loop early in msUVRASTERLayerInitItemInfo would result in uninitialized values in itemindexes and eventually propagated to values as well. Perhaps because msUVRASTERLayerGetItems was never called?

gogglesguy:
Although the testsuit fails (it no longer crashes, yay!), i believe the output images look fine. I'm sure @dmorissette can take a look at this :)

dmorissette:
I hate to be blocking this but don't think I'll be able to look into this anytime soon. As I'm not the author of the code anyway, any other committer would be as good as me to reproduce/test and validate the fix

tbonfort:
applied/squashed in d6d3432b8892c7, thanks

dmorissette:
Thank you @tbonfort

PR #5149, opened 2015-08-24
Fix memory leak when setting format options from mapfile

gogglesguy:
Before setting any formatoption from mapfile, make sure to free previous formatoptions as allocated by the default formats.

This slightly modifies the behaviour. Previously it would always overwrite the numformatoptions and thus  preventing any use of the default format options. Now it preserves the default format options if none are specified. I'm not sure what behaviour is required. It should be easy to change though if needed.

### DIFF
--- a/mapfile.c
+++ b/mapfile.c
@@ -5060,8 +5060,9 @@ static int loadOutputFormat(mapObj *map)
             format->renderer = MS_RENDER_WITH_RAWDATA;
         }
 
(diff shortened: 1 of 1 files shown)


tbonfort:
I think this patch needs to be reworked to use ```msSetOutputFormatOption``` calls instead of freeing/recreating the  formatoption list from scratch. That will overwrite an existing key and keep the existing ones.

tbonfort:
closing, see #5255

PR #5150, opened 2015-08-24
fix memory leak in mapproject.c

gogglesguy:
Forgotten msFree when removing the last projection argument.


tbonfort:
applied in 433428e5f972ec , thanks

PR #5151, opened 2015-08-24
Fix a couple of forgotten msFreeShape in msDrawVectorLayer

gogglesguy:


tbonfort:
applied in 8cf63dcc98dac1 thanks

PR #5152, opened 2015-08-24
Fix leak in msWMSPrepareNestedGroups

gogglesguy:


tbonfort:
applied in 1b08236298701d52b1 thanks

PR #5161, opened 2015-09-08
Fix segfault with expression bindings on database backends

mkofahl:
In some situations, attribute bindings may cause a segmentation fault, when executing multiple MapScript operations, e.g. `queryByPoint()` followed by `draw()` on the same layer. This is because the expression is tokenized only once - when uninitialized. In case of an attribute binding, the token bindval index points to the sql column-index built from the list of required attributes, in case of `queryByPoint()` all attributes are requested. A possibly following `draw()` will reduce the number of attributes to be queried from the database, so the column index in the resultset changes, but the expression token is still the same.

The fix is to free the expression object, when `msLayerFreeItemInfo()` is called, which is currently done at `LayerClose()` only.

### DIFF
--- a/maplayer.c
+++ b/maplayer.c
@@ -65,6 +65,13 @@ void msLayerFreeItemInfo(layerObj *layer)
       return;
   }
   layer->vtable->LayerFreeItemInfo(layer);
+
+  /*
+   * Layer expressions with attribute binding hold a numeric index pointing
+   * to an iteminfo (node->tokenval.bindval.index). If iteminfo changes,
+   * an expression may be no longer valid. (#5161)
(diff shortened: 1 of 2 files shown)


mkofahl:
@tbonfort Looks good to me now.

tbonfort:
ping @sdlime

tbonfort:
@sdlime can you please check this is OK for 7.0.1 ?

sdlime:
Thomas: Missed this one. Ok to include IMHO... --Steve

tbonfort:
Applied to branch-7-0 in 7413952a17f82

PR #5164, opened 2015-09-12
Add ARCHIVE install target for Windows builds.

rouault:

PR #5165, opened 2015-09-12
Don't install mapserver target twice.

sebastic:
The mapserver target only needs to be installed when BUILD_DYNAMIC is true.

For convenience the mapserver_static target is assigned to the staticlib component.

This change also includes the ARCHIVE target from #5164.


sebastic:
The discussion in #5119 raises the question if I should rebase this PR on branch-7.0 instead of master to ease merging?

tbonfort:
Bas,
Branch-7-0 would be perfect!

sebastic:
The rebased PR is available in #5166.

PR #5166, opened 2015-09-18
Don't install mapserver target twice (rebased for branch-7-0)

sebastic:
GitHub doesn't like recreating branches apparently, this rebase should have replaced the commits for #5165 instead of a new PR.

The mapserver target only needs to be installed when BUILD_DYNAMIC is true.

For convenience the mapserver_static target is assigned to the staticlib component.

This change also includes the ARCHIVE target from #5164.


sebastic:
This change was already cherry-picked as 3c10ad1 into branch-7-0 & master as mentioned in #5164.

Issue #5167, opened 2015-09-19
Mapserver hangs if the text is empty


krzbor:
Mapserver (7.0.0) hangs if the text is empty:
LAYER
	#NAME         je_100_n
	GROUP         granice
	METADATA
	  "wms_title"    "Powiaty - nazwy"   ##required‚‚
	END
	PROJECTION
		"init=epsg:2180"   ##required
	END

	DATA         je100O
	STATUS       off
	TYPE         polygon
	LABELITEM    "NAZWA" #  <--------- hangs if the text is empty
	CLASS
		LABEL
			ENCODING WINDOWS-1250
			FORCE TRUE
			COLOR 200 0 0
			TYPE TRUETYPE
			FONT arial
			SIZE 9
			ANTIALIAS FALSE
			POSITION cc
			PARTIALS FALSE
			STYLE
				GEOMTRANSFORM 'labelpoly'
				COLOR 255 255 255
			END
		END # end of label
	END
END

tbonfort:
unable to reproduce with a simple testcase. please provide a full testcase (data + full mapfile + request)

krzbor:
It is difficult for me to extract a small piece of data and file maps. The server hangs when I try to view the area where the attribute value is empty. I changed a blank text on the minus sign (inside DBF file) and everything was OK.

mkofahl:
@krzbor Are you able to compile branch-7-0 with patch 5570f5d3c960ab109bc5574fb9de028eb3624609 yourself? I noticed the same behaviour (infinitve loop until http server timeout) when labelitem met database null-values. A simplified test-case worked however. I found a small typo in `msAddLabelGroup` which solved the problem for mee.

krzbor:
5570f5d patch solved my problem. Thank you mkofahl

ejn:
Merged #5169 fixes this according to OP. Closing.

PR #5169, opened 2015-09-21
Avoid infinite loop during labeling, possibly related to #5167.

mkofahl:


tbonfort:
applied in c00e84a36a83d , thanks @mkofahl

PR #5176, opened 2015-10-04
Make return compatible with 64bits checks.

tigerfoot:
Modify as suggested to make gcc check happy in 64 bits world.


tigerfoot:
Sorry close and will redo it without dyslexia in issue number

Issue #5174, opened 2015-10-04
7.0 libjavamapscript compile error with swig 3.0 javamodule.i


tigerfoot:
javamodule.i line 4 use obsolete swig comment
```
# Uncomment this if you wish to hace enums wrapped in an interface compatible
# with that generated by swig 1.3.21 (tests wont compile, though)
#%include enumsimple.swg
```
Simple fix is using new comment syntax
```
/* 
   Uncomment this if you wish to hace enums wrapped in an interface compatible
   with that generated by swig 1.3.21 (tests wont compile, though)
   %include enumsimple.swg
*/
```

PR #5178, opened 2015-10-04
Use new swig comment Fix issue #5174

tigerfoot:
Simply use new swig comments


rouault:
@tigerfoot just submit a PR based on branch-7-0  (branch-7-0 is then merged to master regularly)

tigerfoot:
On Sunday 04 October 2015 02.53:09 Even Rouault wrote:
> @tigerfoot just submit a PR based on branch-7-0  (branch-7-0 is then merged to master regularly)
> 
Sorry I've seen your remark too late

The way origin is set is no more compatible with old way of mapserver and not common 
compared to what other project do origin === master

Issue #5137, opened 2015-08-11
Building mapserver 7.0 checker on openSUSE complain mapserver 64bit-portability-issue in mapthread.c


tigerfoot:
E: mapserver 64bit-portability-issue /home/abuild/rpmbuild/BUILD/mapserver-7.0.0/mapthread.c:191

I've no idea how to fix it.
Mapserver is configured with the following option.

```
%cmake \
  -DCMAKE_INSTALL_PREFIX=%{_prefix} \
	-DCMAKE_SKIP_RPATH=ON \
	-DCMAKE_C_FLAGS_RELEASE="%{optflags} -fno-strict-aliasing" \
	-DCMAKE_CXX_FLAGS_RELEASE="%{optflags} -fno-strict-aliasing" \
	-DCMAKE_VERBOSE_MAKEFILE=ON \
	-DCMAKE_BUILD_TYPE="Release" \
	-DCMAKE_SKIP_INSTALL_RPATH=ON \
	-DCMAKE_SKIP_RPATH=ON \
	-DWITH_CAIRO=TRUE \
	-DWITH_CLIENT_WFS=TRUE \
	-DWITH_CLIENT_WMS=TRUE \
	-DWITH_CURL=TRUE \
	-DWITH_FCGI=TRUE \
  -DWITH_FRIBIDI=FALSE \
  -DWITH_HARFBUZZ=FALSE \
	-DWITH_GDAL=TRUE \
	-DWITH_GEOS=TRUE \
	-DWITH_GIF=FALSE \
	-DWITH_ICONV=TRUE \
	-DWITH_JAVA=TRUE \
	-DWITH_KML=TRUE \
	-DWITH_LIBXML2=TRUE \
	-DWITH_OGR=TRUE \
	-DWITH_MYSQL=TRUE \
	-DWITH_PERL=TRUE \
	-DCUSTOM_PERL_SITE_ARCH_DIR="%{perl_vendorarch}" \
	-DWITH_PHP=TRUE \
	-DWITH_POSTGIS=TRUE \
	-DWITH_PROJ=TRUE \
	-DWITH_PYTHON=TRUE \
	-DWITH_RUBY=TRUE \
	-DWITH_SOS=TRUE \
	-DWITH_THREAD_SAFETY=TRUE \
	-DWITH_WCS=TRUE \
	-DWITH_WMS=TRUE \
	-DWITH_WFS=TRUE \
	-DWITH_XMLMAPFILE=TRUE \
[...]

PR #5179, opened 2015-10-04
Make return compatible with 64bits checks fix issue #5137

tigerfoot:
Fix the problem with 64 bits checker


tbonfort:
closed with #5137

PR #5194, opened 2015-11-07
Mapfile that is encoded as UTF-8 with BOM is not accepted

tbonfort:
Mapserver 7.0 does not accept mapfiles if they start with BOM. It seems that BOM has no effect in UTF-8 encoding http://www.unicode.org/faq/utf_bom.html#bom5 but some software still write it by default to text files, for example Notepad++
If mapfile starts with BOM Mapserver gives this error:
msLoadMap(): Unknown identifier. First token must be MAP, this doesn't look like a mapfile. 
Issue can be tested with the attached mapfile that github did not accept it as "bom.map" so I renamed it into "bom.map.txt". It is easy to create by hand as well because the whole mapfile is just
MAP
END

Test query:
http://localhost/cgi-bin/mapserv.exe?map=bom.map

I think that making Mapserver to accept starting BOM could not break anything. Minimum action would be to mention the denial of BOM in the migration guide and in the mapfile documentation,

[bom.map.txt](https://github.com/mapserver/mapserver/files/29364/bom.map.txt)


tbonfort:
@jratike80 can try out the provided patch?
@sdlime can you review?

jratike80:
Sorry, I can test only with precompiled binaries like the Windows development builds in gisinternals.com. The issue is well defined and I trust that you understand what you did.

sdlime:
@tbonfort, patch looks reasonable to me. I wonder if it's really necessary but since it's easy... Any reason we'd support other BOMs?

jratike80:
@sdlime, would you rather explain to Windows users why they can't create a proper mapfile with Notepad or why "UTF-8" selected from the menu of Notepad++ is not the correct one but they must select "UTF-8 without BOM" and so on? I think that this fix is friendly for the users and developers as well.

tbonfort:
@sdlime no use in supporting other boms as we require mapfiles to be utf8 encoded

tbonfort:
rebased to branch-7-0 in e50c067a1577efc9

PR #4731, opened 2015-11-10
Reminder to fix EPSG parsing code to make it thread safe

tbonfort:
While adding Java mapscript to the test target I have received a segfault
in the WxS threaded tests that come with Java mapscript.

In an attempt to fix it I have, for now, blocked concurrent access to **msWMSGetCapabilities**, but this is probably too coarse.

The original pull request with discussion is here:
https://github.com/mapserver/mapserver/pull/4728

This ticket is a reminder to fix the EPSG parsing code, with a hint that the cause of the segfaults could be the use of the static variable **epsgCode** mentioned here:

http://lists.osgeo.org/pipermail/mapserver-dev/2013-August/013509.html

Once the EPSG parsing code has been fixed the locks around msWMSGetCapabilities must be removed.

### DIFF
--- a/mapcontext.c
+++ b/mapcontext.c
@@ -1302,7 +1302,8 @@ int msSaveMapContext(mapObj *map, char *filename)
 int msWriteMapContext(mapObj *map, FILE *stream)
 {
 #if defined(USE_WMS_LYR) && defined(USE_OGR)
-  const char * version, *value;
+  const char * version;
+  char *pszEPSG;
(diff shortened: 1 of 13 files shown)


tbonfort:
note to future @tbonfort: never take on a fix on the OWS code, how ever trivial it may seem at first, that code is brutal...
The attached patches apply to branch-7-0 given it's a bugfix, however considering their extent they may be better suited for the master branch. Any comments about that @sdlime or @dmorissette ?

tbonfort:
applied to master, this will be in 7.2

PR #5195, opened 2015-11-10
Compiling without iconv fails.

tbonfort:
The function msLayerEncodeSahpeAttributes calls iconv_open in mapstring.c, line 2206 even if iconv is not being used.  The #ifdef USE_ICONV  block maybe should extend through the entire function instead of ending at line 2201 with a new ifdef USE_ICONV block starting at line 2213?


tbonfort:
@offroadbis provided patch should fix the issue

offroadbis:
First time using. I am not sure that I wanted a pull request.  That fixed the problem. Thanks.

PR #5209, opened 2015-12-11
WMS dimension filter for tileindex

edigiacomo:
WMS dimensions in raster layers, similar to [`TIME` support for raster layers](http://mapserver.org/ogc/wms_time.html#example-of-wms-t-with-postgis-tile-index-for-raster-imagery).

### DIFF
--- a/mapwms.c
+++ b/mapwms.c
@@ -685,6 +685,7 @@ int msWMSApplyDimensionLayer(layerObj *lp, const char *item, char *value, int fo
 {
   int result = MS_FALSE;
   char *pszExpression=NULL;
+  int tlpindex = -1;
 
   if (lp && item && value) {
     /*for the value, we support descrete values (2005) */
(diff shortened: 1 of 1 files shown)


sdlime:
Any WMS maintainers available to review this patch? Seems reasonable but it's not an area of the code I am familiar with. --Steve

@dmorissette

tbonfort:
@edigiacomo I found no place in the code where a ```TIME``` filter was being explicitely set on an underlying tileindex layer as it seems that you are doing in the generic dimension case here. Can you point me where in the code you are "copying" this from ?

edigiacomo:
See `msRASTERLayerSetTimeFilter` in `maprasterquery.c` (it's the `LayerSetTimeFilter` of a raster layer).

tbonfort:
Applied to branch-7-0 in 7daec58ba4651779b45 , thanks

PR #5214, opened 2015-12-16
Update mapserver.h

kellyelton:
Made mapserver.h compatible with VSC 1900. VSC 1900 has snprintf built in, and will throw an error if you define it.


sdlime:
Thank you... merged.

PR #5213, opened 2016-01-04
Symbol orientation problem using GEOMTRANSFORM "vertices" with ANGLE AUTO

tbonfort:
We tried to use GEOMTRANSFORM "vertices" with ANGLE AUTO to display a directional arrow on each vertex of a GPS track and noticed that the current version of calcMidAngle() sometimes results in symbols with the reversed orientation.

Here is an illustration of the problem using the msautotest/renderers/geomtransform-vertex.map  in which the vertex symbols from the original test case are replaced by arrow heads. The blue dot marks the beginning of the line, and you'll notice that the arrow heads along the lines are not always in the right direction:
ttt-before2

The correct result would be this:
ttt

Here is a new version of calcMidAngle() that produces the corrected image above. The mathematicians around the room are welcome to propose a more optimal method.

```
double calcMidAngle(pointObj *p1, pointObj *p2, pointObj *p3)
{
  pointObj p1n;
  double dx12, dy12, dx23, dy23, l12, l23;

  /* We treat both segments as vector 1-2 and vector 2-3 and 
   * compute their dx,dy and length 
   */
  dx12 = p2->x - p1->x;
  dy12 = p2->y - p1->y;
  l12 = sqrt(dx12*dx12 + dy12*dy12);
  dx23 = p3->x - p2->x;
  dy23 = p3->y - p2->y;
  l23 = sqrt(dx23*dx23 + dy23*dy23);

  /* Normalize length of vector 1-2 to same as length of vector 2-3 */
  if (l12 > 0.0)
  {
    p1n.x = p2->x - dx12*(l23/l12);
    p1n.y = p2->y - dy12*(l23/l12);
  }
  else
    p1n = *p2;  /* segment 1-2 is 0-length, use segment 2-3 for orientation */

  /* Return the orientation defined by the sum of the normalized vectors */
  return calcOrientation(&p1n, p3);
}
```

I'll try to produce a proper pull request with the fix and the necessary updates to msautotest but since I'm a still git-challenged, it may take some time before I get to that so I thought I should file this here in case another dev wants to grab and commit the fix.

### DIFF
--- a/mapgeomtransform.c
+++ b/mapgeomtransform.c
@@ -83,12 +83,30 @@ double calcOrientation(pointObj *p1, pointObj *p2)
 
 double calcMidAngle(pointObj *p1, pointObj *p2, pointObj *p3)
 {
-  double theta1,theta2;
-  theta1 = atan2(p1->x-p2->x,p1->y-p2->y);
-  if(theta1<0) theta1 += MS_2PI;
-  theta2 = atan2(p3->x-p2->x,p3->y-p2->y);
-  if(theta2<0) theta2 += MS_2PI;
-  return MS_RAD_TO_DEG*((theta1+theta2)/2.0);
+  pointObj p1n;
+  double dx12, dy12, dx23, dy23, l12, l23;
+
+  /* We treat both segments as vector 1-2 and vector 2-3 and 
+   * compute their dx,dy and length 
+   */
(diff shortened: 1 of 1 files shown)


tbonfort:
rebased into branch-7-0 in 9b93019cb7ad2e8603

Issue #5216, opened 2016-01-04
Alpha band not always detected


rouault:
Currently, in the absence of a BANDS processing directive, msDrawRasterLayerGDAL() only checks that if the 4th band is the alpha band when the raster has 4 or more bands. But for some products, the 4th band might be the last one, so we'd better check the color interpretation of all bands >= 4th one.

tbonfort:
@rouault do we want this in 7.0.1 ?

rouault:
As far as I'm concerned, I don't need it in 7.0.1. I guess it could be backported quite safely though

jratike80:
Was this ever included in some release?

PR #5217, opened 2016-01-04
msDrawRasterLayerGDAL(): find alpha band even if not 4th one (#5216)

rouault:
Currently, in the absence of a BANDS processing directive,
msDrawRasterLayerGDAL() only checks that if the 4th band is the alpha
band when the raster has 4 or more bands. But for some products, the 4th
band might be the last one, so we'd better check the color interpretation
of all bands >= 4th one.

### DIFF
--- a/mapdrawgdal.c
+++ b/mapdrawgdal.c
@@ -333,24 +333,32 @@ int msDrawRasterLayerGDAL(mapObj *map, layerObj *layer, imageObj *image,
 
   /*
    * Set up the band selection.  We look for a BANDS directive in the
-   * the PROCESSING options.  If not found we default to red=1 or
-   * red=1,green=2,blue=3 or red=1,green=2,blue=3,alpha=4.
+   * the PROCESSING options.  If not found we default to grey=1, grey=1,alpha=2,
+   * red=1,green=2,blue=3 or red=1,green=2,blue=3,alpha>=4.
    */
 
   if( CSLFetchNameValue( layer->processing, "BANDS" ) == NULL ) {
+    const int gdal_band_count = GDALGetRasterCount( hDS );
     red_band = 1;
 
-    if( GDALGetRasterCount( hDS ) >= 4
-        && GDALGetRasterColorInterpretation(
-          GDALGetRasterBand( hDS, 4 ) ) == GCI_AlphaBand )
-      alpha_band = 4;
+    if( gdal_band_count >= 4 )
+    {
+        /* The alpha band is not necessarily the 4th one */
+        for( i = 4; i <= gdal_band_count; i ++ ) {
+            if( GDALGetRasterColorInterpretation(
+                GDALGetRasterBand( hDS, i ) ) == GCI_AlphaBand ) {
+                alpha_band = i;
+                break;
+            }
+        }
+    }
 
-    if( GDALGetRasterCount( hDS ) >= 3 ) {
+    if( gdal_band_count >= 3 ) {
       green_band = 2;
       blue_band = 3;
     }
 
-    if( GDALGetRasterCount( hDS ) == 2
+    if( gdal_band_count == 2
         && GDALGetRasterColorInterpretation(
           GDALGetRasterBand( hDS, 2 ) ) == GCI_AlphaBand )
       alpha_band = 2;


mapdrawgdal.c
```diff
@@ -333,24 +333,32 @@ int msDrawRasterLayerGDAL(mapObj *map, layerObj *layer, imageObj *image,
...
-    if( GDALGetRasterCount( hDS ) >= 4
-        && GDALGetRasterColorInterpretation(
-          GDALGetRasterBand( hDS, 4 ) ) == GCI_AlphaBand )
-      alpha_band = 4;
+    if( gdal_band_count >= 4 )
+    {
+        /* The alpha band is not necessarily the 4th one */
+        for( i = 4; i <= gdal_band_count; i ++ ) {
+            if( GDALGetRasterColorInterpretation(
+                GDALGetRasterBand( hDS, i ) ) == GCI_AlphaBand ) {
```
tbonfort: Are we sure that all GDAL drivers will correctly identify which band is the raster band? Should we set band==4 as the default if no band was explicitely marked as being the alpha one (with the current patch, band 0 will be used in that case)

rouault: > Are we sure that all GDAL drivers will correctly identify which band is the raster (edit: you mean alpha ?) band?

Not necessarily, but in most cases they should... There are different situations : 
* the alpha band is correctly identified as such, in which case this patch improves the situation by finding it even if not the 4th one. The past situation was that a R,G,B,unknown,Alpha raster was treated as RGB only. With this patch, the Alpha band will be taken into account
* there was an alpha band but the driver fails to identify it as such (R,G,B,Unknown). The patch doesn't change the situation. MapServer only considers RGB
* there was no alpha band but the driver wrongly identifies it as alpha. It it was a 4 band raster - R,G,B,A(wrong) - then the patch doesn't change the behaviour. If it was a 5 or more band, R,G,B,unkown,A(wrong) then the patch will pick up the 5th band as alpha, whereas previously it was ignored.

So I'd say the benefits of the patch outweights the drawbacks . The only "regression" would be a 5 band raster with a driver wrongly identifying a band >= 5 as alpha. rather uncommon. In that situation, people can either put an explicit BANDS directive as a workaround. Or sometimes create a .aux.xml file or wrap it in a VRT to fix the interpretation.

> Should we set band==4 as the default if no band was explicitely marked as being the alpha one 

I don't think that would be a good idea. You can have R,G,B,Unknown(for example Near Infra Red) products.

> (with the current patch, band 0 will be used in that case)

Not sure to follow you regarding that. "alpha" is initialized to 0, which means no alpha band, and in code later, this is properly dealt as the absence of an alpha band

tbonfort: From what I understood, (R,G,B,Unknown) was treated as (R,G,B,A) and would now be treated as (R,G,B). If that was the previous behavior I believe it should not be modified; if the previous behavior was to create (R,G,B) then I am fine with the proposed change.

rouault: > From what I understood, (R,G,B,Unknown) was treated as (R,G,B,A) and would now be treated as (R,G,B)

I don't see anything in the code that would have done that and my below test also confirms that R,G,B,Unknown was treated as RGB.

Given test.map :
```
MAP

NAME TEST
STATUS ON
SIZE 400 300
EXTENT 0.5 0.5 399.5 299.5
IMAGECOLOR 255 255 0

IMAGETYPE png24_t

OUTPUTFORMAT
  NAME png24_t
  DRIVER "GDAL/PNG"
  IMAGEMODE RGBA
  TRANSPARENT ON
END

LAYER
  NAME rgba
  TYPE raster
  STATUS OFF
  DATA data/rgba.tif
END

LAYER
  NAME rgbu
  TYPE raster
  STATUS OFF
  DATA data/rgbunknown.tif
END


END # of map file
```

placed in msautotest/gdal. Where data/rgba.tif is the file in git and data/rgbunkown.tif is produced with the following operation :
```
gdal_translate data/rgba.tif data/rgbunknown.vrt -of vrt
edit data/rgbunknown.vrt  to remove <ColorInterp>Alpha</ColorInterp>
gdal_translate data/rgbunknown.vrt data/rgbunknown.tif  -co alpha=no
```

Then 'shp2img -m test.map -l rgba -o out_rgba.png' and 'shp2img -m test.map -l rgbu -o out_rgbu.png' produce different images. And that, before and after my patch.

PR #5215, opened 2016-01-05
WFS request with chinese keywords does not have any result

tbonfort:
My WFS request is the following code:
```
          <?xml version='1.0' encoding='UTF-8'?>
            <wfs:GetFeature  service='WFS' version='1.0.0' 
            xmlns:wfs='http://www.opengis.net/wfs' 
            xmlns:gml='http://www.opengis.net/gml' 
            xmlns:ogc='http://www.opengis.net/ogc' 
           xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance' 
            xsi:schemaLocation='http://www.opengis.net/wfs http://schemas.opengis.net/wfs/1.0.0/wfs.xsd'>           
 <wfs:Query typeName='district' srsName='EPSG:3830'>
            <ogc:Filter>
            <ogc:PropertyIsEqualTo>
            <ogc:PropertyName>TField</ogc:PropertyName>
            <ogc:Literal>高新区</ogc:Literal>
            </ogc:PropertyIsEqualTo>
           </ogc:Filter>
           </wfs:Query>
           </wfs:GetFeature>
```
I want to search the feature that the value of field named TField is 高新区. I set the the layer's encoding to GBK, ant the above request returns the message that No matching record(s) found. however, if I set the keyword to english words or number, then I can get the result. The code above works well on Geoserver WFS, but can't got result from mapserver' WFS. Additional, "高新区" is Chinese words.

### DIFF
--- a/maplayer.c
+++ b/maplayer.c
@@ -330,6 +330,14 @@ int msLayerNextShape(layerObj *layer, shapeObj *shape)
     if(rv != MS_SUCCESS) return rv;
 
     filter_passed = MS_TRUE;  /* By default accept ANY shape */
+    
+    /* attributes need to be iconv'd to UTF-8 before any filter logic is applied */
+    if(layer->encoding) {
+      rv = msLayerEncodeShapeAttributes(layer,shape);
+      if(rv != MS_SUCCESS)
+        return rv;
+    }
+    
     // if(layer->numitems > 0 && layer->iteminfo) {
       filter_passed = msEvalExpression(layer, shape, &(layer->filter), layer->filteritemindex);
     // }
@@ -344,11 +352,6 @@ int msLayerNextShape(layerObj *layer, shapeObj *shape)
       return rv;
   }
 
-  if(layer->encoding) {
-    rv = msLayerEncodeShapeAttributes(layer,shape);
-    if(rv != MS_SUCCESS)
-      return rv;
-  }
   
   return rv;
 }


tbonfort:
Can you try the attached patch? It would be nice if you could provide a sample shapefile so we can include this check in our test suite

bruce6135:
I can not attach a  patch in the comment, so I upload the file into oneDrive. 
The link is :
https://onedrive.live.com/redir?resid=660C89D3266533C6!109&authkey=!AIAeACYiK62s0rY&ithint=file%2czip
https://onedrive.live.com/redir?resid=660C89D3266533C6!110&authkey=!AOoFQdF6ItRjpv4&ithint=file%2cmap
If you can not download, you can you name hymyjl2010@163.com and pwd Bruce6228 to logon.
 My data's EPSG is 4326.
When I enter the URL like this:"http://localhost:88/cgi-bin/mapserv.exe?map=D:/ms4w/Apache/htdocs/map/query.map&SERVICE=WFS&VERSION=1.0.0&REQUEST=getFeature&TYPENAME=city&Filter=<Filter><PropertyIsEqualTo><PropertyName>NAME</PropertyName><Literal>塔河县</Literal></PropertyIsEqualTo></Filter>", then show the following:
<gml:boundedBy>
<gml:null>missing</gml:null>
</gml:boundedBy>
The information in the log file is :[Tue Jan 05 23:07:33 2016].399000 CGI Request 1 on process 4908
[Tue Jan 05 23:07:33 2016].400000 FLTLayerApplyPlainFilterToLayer():  ("[NAME]" ="塔河县")
[Tue Jan 05 23:07:33 2016].438000 msQueryByFilter(): Search returned no results. No matching record(s) found.
[Tue Jan 05 23:07:33 2016].438000 mapserv request processing time (msLoadMap not incl.): 0.039s
[Tue Jan 05 23:07:33 2016].438000 msFreeMap(): freeing map at 05BAA990.

tbonfort:
Pushed the fix to branch-7-0, along with an autotest in mapserver/msautotest@b52c73c . Without the current fix, the autotest result is
```
<gml:boundedBy>
<gml:null>missing</gml:null>
</gml:boundedBy>
```
instead of the correct feature.

Thanks for the testcase @bruce6135 , you can delete the data now, I was able to create the autotest with an exisiting dataset

bruce6135:
Congratulations for you fixed this problem,  and thanks for you help. I will try to compile in my computer to fix this problem.

bruce6135:
Sombody helped me to find another method. In QGIS, saving the layer as another layer and set the new layer's code to UTF-8.The new layer's property sheet is now in UTF-8 code and shows correctly.  Then I can query chinese words with WFS service. The disadvantage of this method is that UTF-8  is not contain all the chinese words, by save as the layer, some chinese words will be garbled and must be reenter with the correct content. However, those garbled words usually is not often used and account for a small proportion of chinese words.

tbonfort:
I doubt this has something to do with UTF-8 not supporting those glyphs. You are probably giving qgis an incorrect version of gbk (with respect to the version actually used by your dataset) when asking it to convert it to utf8.
ps: always use utf-8 and you'll never have to deal with encoding issues again :)

PR #5220, opened 2016-01-12
WMS caps: add queryable attribute for non-leaf layers

rouault:
@sdlime @jratike80 

Related to discussion
https://lists.osgeo.org/pipermail/mapserver-users/2016-January/078538.html

Currently MapServer only sets the queryable attribute on leaf layers.
This changeset also adds it non-leaf layers of group/layer_group. A layer
is queryable if and only if all of its sublayers are queryable.

Would require adjusting a few msautotest results. Not sure if it is branch-7-0 material or master only. msautotest doesn't seem to include tests with several level of nesting.

### DIFF
--- a/mapwms.c
+++ b/mapwms.c
@@ -2625,7 +2625,20 @@ void msWMSPrintNestedGroups(mapObj* map, int nVersion, char* pabLayerProcessed,
         groupAdded = 1;
       }
     } else {
-      msIO_printf("%s    <Layer>\n", indent);
+      int queryable = MS_TRUE;
+      for (j = 0; j < map->numlayers; j++)
+      {
+        if( !pabLayerProcessed[j] &&
+            msWMSIsSubGroup(nestedGroups[index], level, nestedGroups[j], numNestedGroups[j]) )
+        {
+            if( !msIsLayerQueryable(GET_LAYER(map, j)) )
+            {
+                queryable = MS_FALSE;
+                break;
+            }
+        }
+      }
+      msIO_printf("%s    <Layer queryable=\"%d\">\n", indent, queryable);
       msIO_printf("%s      <Title>%s</Title>\n", indent, nestedGroups[index][level]);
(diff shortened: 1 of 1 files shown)


tbonfort:
FWIW, I think a layer should become queryable if at least one of its sublayers is queryable (the actual query would only return results from the queryable sublayers)

rouault:
@tbonfort My understanding of the discussion was that a layer is currently queryable only if all its sublayers are queryable. But I didn't check the code. Or is your remark about modifying the current behaviour ? Do the WMS specs say something about the expected behaviour ?

tbonfort:
@rouault as per https://lists.osgeo.org/pipermail/mapserver-users/2016-January/078541.html it would seem that geoserver does what I would assume to be the default behavior, i.e. layer queryable if at least one sublayer queryable. If that's not easily doable I'm fine with starting with your proposed behavior.

landryb:
Fwiw, i stumbled upon this behaviour, and would welcome the change. If one of the sublayers is queryable, just mark the layer as queryable. Right now i'm grouping a TILEINDEX layer and a regular layer, and since a TILEINDEX layer is forcedly non-queryable (cf mapquery.c#L122) the group layer isnt queryable.. while i'd like queries to be able to return values from the 'regular layer'.

tigerfoot:
+1

sdlime:
@rouault, what do you want to do this one? Seems like a positive change although I see a Travis is failing.

rouault:
Note: this PR shouldn't be merged. From above comments, a better behaviour would be to make a layer queriable as soon as one of its sublayers is also queriable.

sdlime:
@rouault, can I close this?

rouault:
I'm clossing this pull request, because it doesn't implement the desired behaviour, but opened https://github.com/mapserver/mapserver/issues/5570 to track the issue though

PR #5226, opened 2016-01-20
Transfer geometries in binary from Postgres.

pvalsecc:
Use a trick to have the other fields (when any) still transfered in
text form.

### DIFF
--- a/mappostgis.c
+++ b/mappostgis.c
@@ -81,6 +81,12 @@
 #define HAS_Z   0x1
 #define HAS_M   0x2
 
+#if TRANSFER_ENCODING == 256
+#define RESULTSET_TYPE 1
+#else
+#define RESULTSET_TYPE 0
+#endif
+
 #ifdef USE_POSTGIS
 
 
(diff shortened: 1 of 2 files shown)


pvalsecc:
The Travis failure is because of those two files that have a bad (according to pep8) indentation:
* python/plugins/processing/algs/grass/GrassAlgorithm.py
* python/plugins/processing/algs/grass7/nviz7.py

I didn't touch those files with my commit and autopep8 on my machine does not agree with the one on Travis.

tbonfort:
@pvalsecc the failure is
```
# query => query_postgis_test012.txt
--- ../expected/query_postgis_test012.txt	2016-01-20 13:18:30.446291031 +0000
+++ query_postgis_test012.txt	2016-01-20 13:20:03.918291031 +0000
@@ -1,3 +0,0 @@
-Marshall:MARS 
-
-
```

pvalsecc:
Thanks Tomas, big mix up on my side... I should not do two PRs at the same time ;-)

I'll investigate that.

pvalsecc:
Can somebody have a look at this PR, please?

sdlime:
Looks ok to me but I'm no expert on the PostGIS code. @tbonfort or @pramsey, any comment?

pramsey:
If it regresses, I'm OK w/ it

tbonfort:
@pvalsecc I think this should be a change for 7.2 rather than 7.0 ?
I'm no expert on the subject at hand. Are there any issues you think may arise if we apply this?

pvalsecc:
@tbonfort, it's as you guys want. I don't mind. I worked on 7.0 because it's the version I was using in prod and I didn't want to risk some incompatibilities with my mapfiles. But I guess transplanting the patch shouldn't be a problem.

pvalsecc:
I've created another PR based on master: https://github.com/mapserver/mapserver/pull/5238

tbonfort:
@pvalsecc If you can rework this to be able to select the transfer mode at runtime rather than compile time, and that the default remains to the current behavior, then we could apply this to the stable 7.0 branch. If not, let's close this issue and focus only on 7.2 in #5238.

pvalsecc:
@tbonfort, I don't see the point. Closing this PR.

PR #5228, opened 2016-01-26
Fix 'shoud' typo, replace with 'should'.

sebastic:
The issue was reported by the lintian QA tool as part of the Debian package build.


tbonfort:
applied in 84c9d79, thanks

PR #5230, opened 2016-02-01
Changes related to label placement issues Tamas found related to narrow rectangles.

sdlime:
Ignore this comment...


sdlime:
A simpler set of changes would apply to the 7.0 and 6.4 branches, just uncommenting a couple of lines in the msPolygonLabelPoint() function. The layer processing option is a new feature and would not be added those older releases, just master.

sdlime:
I'm going to redo this minus the change to ignore the distance check - at least as written. It's just too disruptive especially with MapScript since you have to pass a boolean derived from the layer object.

Issue #5193, opened 2015-11-02
GEOMTRANSFORM (buffer([shape],n)) + GetLegendGraphic -> Error 500


m431m:
Hi,

It appears that the ```Geomtransform``` directive in ```STYLE``` object is not working with _GetLegendGraphic_ requests. It returns an error 500: ‘_End of script output before headers: mapserv_’.

There is no problem in _GetMap_.

I tried using **master** branch.

Regards,

m431m

tbonfort:
please include a testcase (mapfile + request)

m431m:
Yep,

A minimal MAP :
```
MAP
  NAME "test"
  EXTENT -180 -90 180 90
  SIZE 600 400
  STATUS ON
  UNITS DD
  PROJECTION
    "init=epsg:4326"
  END
  WEB
    METADATA
      ows_enable_request "*"
    END
  END
  LAYER
    NAME "my_layer"
    TYPE POLYGON
    FEATURE
      POINTS -100 -60 -100 60 100 60 100 -60 -100 -60 END
    END
    STATUS ON
    TILEITEM "location"
    UNITS METERS
    CLASS
    NAME "legend"
      STYLE
        COLOR 255 0 0
        GEOMTRANSFORM (buffer([shape],-1))
      END
    END
  END
END
```
And the _GetLegendGraphic_ request (basic):
http://hostname/cgi-bin/mapserv?map=/path/to/mapfile.map&version=1.3.0&service=WMS&request=GetLegendGraphic&sld_version=1.1.0&layer=my_layer&format=image/png&STYLE=default

jratike80:
This is how bug reports should be made!

I could repeat Internal server error within 5 minutes. Comment out GEOMTRANSFORM and GetLegendGraphic request is successful. Unfortunately I can't fix anything.

tbonfort:
assigning to @sdlime , the parser seems to be borking here:
```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7a72308 in msDrawTransformedShape (map=0x60d260, image=0x61a800, shape=0x7fffffffd830, style=0x616850, scalefactor=1)
    at /home/tbonfort/dev/mapserver/mapgeomtransform.c:193
193	      switch (tmpshp->type) {
```
```
(gdb) bt
#0  0x00007ffff7a72308 in msDrawTransformedShape (map=0x60d260, image=0x61a800, shape=0x7fffffffd830, style=0x616850, scalefactor=1)
[...]

PR #5233, opened 2016-02-04
Fixed case when drawing legend shapes with style geomtransforms. (#5193)

sdlime:
This should fix issue #5193. If OK, it should be applied to the 6.4 and 7.0 branches. --Steve

### DIFF
--- a/mapgeomtransform.c
+++ b/mapgeomtransform.c
@@ -180,6 +180,15 @@ int msDrawTransformedShape(mapObj *map, imageObj *image, shapeObj *shape, styleO
 
       p.shape = shape; /* set a few parser globals (hence the lock) */
       p.expr = &(style->_geomtransform);
+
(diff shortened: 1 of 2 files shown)


sdlime:
@tbonfort, what is the preferred method to apply a pull request like this to older branches (7.0 and 6.4). Should I have prepared the pull request against 7.0 instead? --Steve

tbonfort:
@sdlime it should ideally be based on 6.4. Don't bother too much, it can be easily back ported if not.

tbonfort:
PS an autotest for this would be useful in the long run

sdlime:
Any objection to the solution?

tbonfort:
I can have a look if you need when I'm back from holiday next week. I'm not familiar with the expression parser so you probably know better than me anyways.

sdlime:
I'll back port to the 7.0 and 6.4 branches by hand... --Steve

Issue #5234, opened 2016-02-05
Labeling thin polygons sometimes fails.


sdlime:
@szekerest reported instances where labeling where thin, simple polygons (think parcels) results in label points outside a polygon. It happens because a edge/distance computation in msPolygonLabelPoint() done after a center of gravity computation drops the label point computation into a second algorithm that uses scan lines to place the label. There is a bug in that second computation that can result in bad label placement. 

The fix is easy, simply uncomment a couple of lines in that function, look for:

```
  /* lp->y = y; */
  /* lp->x = x; */
```

and the issue is resolved. In cases like this it might be even better to skip the distance check or at least make that more robust, but that's another ticket. I tried allowing a configuration via a layer processing option but since label placement is independent of layer properties it's not trivial to make that value available, especially with MapScript.

tbonfort:
pull request in #5235

PR #5235, opened 2016-02-05
Fixed label placement computation issue with simple, thin polygons. (#5234)

sdlime:


sdlime:
The commit breaks a bunch of tests that place labels but not in a way that represents a true regression. @tbonfort or @rouault , how do you think it should be handled? I checked output from bunch that failed and the failure is due to expected and mostly subtle changes in label placement. These sorts of changes are always going to result in lots of errors.

tbonfort:
merged into branch-7-0 in 5c2411e182e3

PR #5238, opened 2016-02-11
Transfer geometries in binary from Postgres.

pvalsecc:
Use a trick to have the other fields (when any) still transfered in
text form.


tbonfort:
applied to master in f6c862e , thanks @pvalsecc

PR #5237, opened 2016-02-11
Labels leak through MASK

tbonfort:
It seems as if there is a border problem when masking out labelled layers. Sometimes labels at the very edge of the map leak through even when they should definitely be masked away.

Example:
Input dataset:
image

The above dataset masked with a mask which is a union of the polygons at the centre of the map:
image
Geometries are correctly masked, but note the label "Nørre Broby".

Using v6.4.1

### DIFF
--- a/maplabel.c
+++ b/maplabel.c
@@ -372,6 +372,9 @@ int msAddLabelGroup(mapObj *map, int layerindex, int classindex, shapeObj *shape
           return MS_SUCCESS;
         }
 #endif
+      } else {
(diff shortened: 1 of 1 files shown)


tbonfort:
applied to branch-6-4 in 4196ac3. This was happening on polygons that where close to the image border when the label-point was computed to be outside of the image extent.

tbonfort:
also applied by merging to branch-7-0 in 6c13c20 and master in 27fb2d1356da9

AsgerPetersen:
Thank you @tbonfort. Highly appreciated!

PR #5244, opened 2016-02-17
Add macro definitions for MIN/MAX if not defined

wraeth:
Definitions for MIN() and MAX() only seem to exist in some environments (Visual Studio, [from what I've found](https://github.com/OpenRCT2/OpenRCT2/issues/293)), breaking linking on some platforms.

This adds simple definitions if the macros aren't defined.


wraeth:
Closing this PR to rebase against master and re-submit.

Issue #5232, opened 2016-02-04
MapServer as WMS client, spaces in layer names on upstream server not supported.


bblammo:
For more background see these threads:

http://osgeo-org.1560.x6.nabble.com/Allowing-spaces-in-layer-names-WMS-WFS-td5213960.html
https://lists.osgeo.org/pipermail/mapserver-users/2014-August/076870.html
https://lists.osgeo.org/pipermail/mapserver-users/2014-August/076886.html

Here is some minimal info on using the Census (Arc) server WMS,  The server is using spaces in both the NAME and TITLE tags, which MapServer doesn't seem to want to push through.

So, as an example, using this WMS:

    http://tigerweb.geo.census.gov/arcgis/services/TIGERweb/tigerWMS_ACS2015/MapServer/WMSServer?

Specifically, two layers, “Census Tracts” and “Census Tracts Labels”

I can open each of these service calls in just about everything, except  Mapserver as a WMS client.  I’m assuming the problem is in how MapServer is sending it’s request.

I've tried encoding in the MapServer request string, as a "+", and " " and  as "%02".  These last two seem to work when directly hitting the WMS, but MapServer is not translating any of them through it's request.

Here is a test mapfile I was using:

##Mapfile

       MAP
            LAYER
                 NAME "Census Tracts"
                 TYPE RASTER
                 STATUS DEFAULT
                 CONNECTION
       "http://tigerweb.geo.census.gov/arcgis/services/TIGERweb/tigerWMS_ACS2015/MapServer/WMSServer?"

                 CONNECTIONTYPE WMS
                 METADATA
                      "wms_srs" "EPSG:4326"
                      "wms_name" "Census Tracts"
                      "wms_server_version" "1.1.1"
                      #"wms_server_version" "1.3.0"
                      "wms_format" "image/png24"

                      "ows_title" "Census Tracts"
                      #"gml_include_items" "all"
                 END

                 PROJECTION
                      "init=epsg:4326"
                 END

                 TEMPLATE "dummy"
            END #layer

       END

jrahkonen:
ArcGIS server requires that space is url-encoded as "%20" but it does not accept space encoded as "+". There has been a ticket in the ESRI ticket tracker about accepting both "%20" and "+" because also the http standard accepts them both but the resolution was "won't fix". I believe I can find more information from my old mails if needed because I reported the issue through ESRI Finland.

jmckenna:
[...]

PR #5246, opened 2016-02-18
modify encodeurl logic for spaces (issue #5232)

jmckenna:
No feedback yet on the testing, but I can see from server stats that 3 people have fully downloaded the experimental build with this change.  This is the change referenced in the thread: https://lists.osgeo.org/pipermail/mapserver-users/2014-August/076879.html

### DIFF
--- a/mapstring.c
+++ b/mapstring.c
@@ -421,7 +421,7 @@ void msStringToLower(char *string)
 void msStringInitCap(char *string)
 {
   int i;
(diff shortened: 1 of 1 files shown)


tbonfort:
This needs a PSC vote I think, especially if we are going to apply this to a point-release.

tbonfort:
closed with #5232

PR #5247, opened 2016-02-19
Add macro definitions for MIN/MAX if not defined

wraeth:


tbonfort:
Thanks for the patch, but this needs to be reworked a little: Occurrences of MAX/MIN should be replaced by MS_MAX/MS_MIN which are always defined (in mapserver.h). You can submit a new PR against branch-7-0. thanks

PR #5248, opened 2016-02-19
Correct compile error with -Werror=format-security

wraeth:
Build fails when built with -Werror=format-security


tbonfort:
thanks!

PR #5249, opened 2016-02-19
msLibXml2GetFirstChildNs() only used with libxml2

wraeth:
This makes use of libxml2 and fails when built with `-DWITH_LIBXML2=OFF`


tbonfort:
backported to branch-7-0 in ab96f8a . thanks!

PR #5251, opened 2016-02-20
Correct calls to MAX/MIN to MS_MAX/MS_MIN

wraeth:
As per https://github.com/mapserver/mapserver/pull/5247#issuecomment-186087171 this renames instances of MAX() and MIN() to MS_MAX() and MS_MIN() respectively. Compiling with this patch goes without issue.


tbonfort:
tweaked a little and rebased into branch-7-0 in f0dedb6 . thanks @wraeth

PR #5253, opened 2016-02-23
Fix format-security errors with swig and java and ruby

Jenselme:
Can compile java and ruby mapscript extension with gcc's
`-Werror=format-security` option.


tbonfort:
applied to branch-7-0 in 90f00f77631e , thanks

Jenselme:
@tbonfort that was quick, thanks!

PR #5208, opened 2016-02-23
Empty response is returned for GeoJSON when no features

tbonfort:
When using GeoJSON as output format with the OGR GeoJSON driver, an empty response (`Content-Length: 0`) is returned when no features match a query.

For GML an empty feature collection is returned, I would expect the same for GeoJSON. GeoServer does it this way and OpenLayers 3 also expects an empty feature collection.

### DIFF
--- a/mapogroutput.c
+++ b/mapogroutput.c
@@ -869,7 +869,7 @@ int msOGRWriteFromQuery( mapObj *map, outputFormatObj *format, int sendheaders )
     int  reproject = MS_FALSE;
     int  nFirstOGRFieldIndex = -1;
 
-    if( !layer->resultcache || layer->resultcache->numresults == 0 )
+    if( !layer->resultcache )
       continue;
 
     /* -------------------------------------------------------------------- */


tbonfort:
The proposed patch fixes the issue, it now emits the layer even the query returned no features. @sdlime: ok?

sdlime:
Looks ok to me... --Steve

juliensam:
With the test mapfile above, I get the following response:
```
{
"type": "FeatureCollection",
"crs": { "type": "name", "properties": { "name": "urn:ogc:def:crs:EPSG::3857" } },
"features": [

]
}
```
Merging

rouault:
I hope all drivers generate valid files with 0 feature written in the layer... I guess most should behave OK.

sdlime:
All OGR drivers, correct?

rouault:
Yes OGR drivers. But anyway the change should be OK.

tbonfort:
autotests added. @juliensam please don't merge pull-requests before discussions concerning the implications have happened.

juliensam:
sorry about this. Noted for the future.

PR #5014, opened 2016-02-23
Min/Max scale values incorrectly formatted (python Mapscript)

tbonfort:
If I set a min or max scale in Mapscript (python) using

```
k.maxscaledenom = 5000000000.0
```

And then generate the mapfile using

```
m.save()
```

The resulting value is formatted as:

```
MAXSCALEDENOM 5e+09
```

Which means the annotation layer isn't rendered until I replace it with:

```
MAXSCALEDENOM 5000000000
```

Is there anyway to control the format of these values when using save()?

Thanks


tbonfort:
applied to branch-6-4 branch-7-0 and master

PR #5255, opened 2016-02-24
Fix memory leak when loading FORMATOPTIONS

tbonfort:
supercedes #5149


tbonfort:
This also fixes an issue where GD fallback was creating 24 bit pngs instead of 8bit ones, given that the formatoptions that we setup in code to handle this fallback where being removed by the parser.

PR #5207, opened 2016-02-26
Feature id is ignored for GeoJSON output format

tbonfort:
I configured a GeoJSON output format using the OGR GeoJSON driver. When I specify the feature id field using `"wfs_featureid" "id"`, the id property is not set in the output.

According to the [GeoJSON specification](http://geojson.org/geojson-spec.html#feature-objects) the id should be set on the feature object:

> If a feature has a commonly used identifier, that identifier should be included as a member of the feature object with the name "id".


tbonfort:
@rouault any comments on this change ?

rouault:
Hum, calling OGR_F_CreateFeature() with a non-default FID might not work in some circumstances. I see for example that the MapInfo driver starting with GDAL 2.0 can handle user-defined FID, but only if the FID is in the range [1, max_fid]. That is to say it can fill holes, but not extend the file. Previous versions ignored completely the FID, so that would work.
There's not a way to know if a driver will ignore completely FID (which is OK), accept any of them and use them as intended, or some others might have restrictions like explained above.

tbonfort:
rebased into branch-7-0 in 4af7028

tsauerwein:
Thanks! :+1:

PR #5259, opened 2016-03-01
Avoid calling msProjectionsDiffer over and over again

tbonfort:
As reported on the mailing list in http://osgeo-org.1560.x6.nabble.com/Performance-issue-with-MS-7-x-td5253419.html , there is a performance issue that became visible once the code to ```msProjectionsDiffer``` became more complicated with eeb1e786. This PR aims to set ```layer->project``` when necessary and then avoid calling ```msProjectionsDiffer``` over and over (for each feature)
cc @pmauduit @yjacolin

### DIFF
--- a/mapchart.c
+++ b/mapchart.c
@@ -250,10 +250,8 @@ int getNextShape(mapObj *map, layerObj *layer, double *values, int *nvalues, sty
   status = msLayerNextShape(layer, shape);
   if(status == MS_SUCCESS) {
 #ifdef USE_PROJ
-    if(layer->project && msProjectionsDiffer(&(layer->projection), &(map->projection)))
+    if(layer->project)
       msProjectShape(&layer->projection, &map->projection, shape);
-    else
-      layer->project = MS_FALSE;
 #endif
 
     if(msBindLayerToShape(layer, shape, MS_DRAWMODE_FEATURES|MS_DRAWMODE_LABELS) != MS_SUCCESS)
@@ -363,6 +361,9 @@ int msDrawPieChartLayer(mapObj *map, layerObj *layer, imageObj *image)
       return MS_FAILURE;
     }
   }
+#ifdef USE_PROJ
+  layer->project = msProjectionsDiffer(&(layer->projection), &(map->projection));
+#endif
   /* step through the target shapes */
   msInitShape(&shape);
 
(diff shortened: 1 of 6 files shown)


yjacolin:
we will test this soon. Thanks!

pmauduit:
Thanks @tbonfort ! This is way better after your commit (https://github.com/tbonfort/mapserver/commit/7cb9582a291bb65628a153239ef5feb900ac1b2d):

here are some metrics about my own tests, always using the same query:

```
/path/to/mapserv QUERY_STRING="MAP=./mymapfile.map&SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&FORMAT=image%2Fpng&TRANSPARENT=true&LAYERS=may_layer&CRS=EPSG%3A3857&STYLES=&WIDTH=2036&HEIGHT=745&BBOX=-13071343.332991421%2C885446.5356554817%2C6848757.734351791%2C8174481.552929889"

```

Time needed before your commit (`HEAD^` on your branch):

```
real	0m5.282s
user	0m5.095s
sys	0m0.188s
```
Time needed after:

```
real	0m1.985s
user	0m1.846s
sys	0m0.132s
```

Callgrind analysis before your commit:
callgrind-before-tbonfort

~38.5 billions CPU cycles needed, most of the time wasted in projections init (pj_init_ctx)

Callgrind after the commit:

callgrind-tbonfort

~11.8 billions CPU cycles needed.

The previous version used by our customer was still faster than the 2 studied here (only 6.5 billions CPU cycles were needed for the previous query in Mapserver 6.4), but the current optimization is clearly better than nothing.

tbonfort:
I wold say that something else has changed then, because with this fix we will end up calling msProjectionsDiffer much less often than in 6.x . Can you show the full callgrind output for 6.4 ?

pmauduit:
Using a compiled version of MS on branch-6.4, I actually have the same result as the one after your commit:

callgrind-64

But below is the screenshot I obtained last week when the same query was performed on the official debian jessie package ; there might be some optimizations done at compilation time by the debian packaging team.

callgrind-641-vanilla

For the record, here is the full debian package version:

```
# apt-cache policy cgi-mapserver
[...]
     6.4.1-5+b3 0
        500 http://httpredir.debian.org/debian/ jessie/main amd64 Packages
```

pmauduit:
@tbonfort do you think it can be merged, or do you still need to work on this PR ? (or any extra infomation from my side ?)

tbonfort:
rebased in branch-7-0 in 6acfaf410dfb7

PR #5264, opened 2016-03-23
SVG symbol gets clipped when rotated

tbonfort:
Non-square SVG symbols are clipped when they are rotated.

Here is an example where a small box of 40x20 pixel is rotated by 45 and 90 degrees:

out

Here is the Mapfile and SVG symbol:
https://gist.github.com/olt/2a345fe22653cd021c29

Tested this with current master, 7.0.1 and 6.4.3 (`shp2img -m clipped-symbol.map -o out.png`).

### DIFF
--- a/mapcairo.c
+++ b/mapcairo.c
@@ -174,7 +174,7 @@ static cairoFaceCache* getCairoFontFace(cairoCacheData *cache, FT_Face ftface) {
   cur->options = cairo_font_options_create();
   cairo_font_options_set_hint_style(cur->options,CAIRO_HINT_STYLE_NONE);
   return cur;
-} 
+}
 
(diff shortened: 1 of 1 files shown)


jmckenna:
I've tested this fix on Ubuntu through librsvg with branch-7.0 + this change.  Is this ok to merge?
ttt

jmckenna:
thanks @olt !

tbonfort:
A PR which breaks tests isn't ready for inclusion, please wait for confirmation before merging PRs you are not affiliated with. Fix added in 45b3917f89

jmckenna:
for the record i had begun looking at the 2 errors reported by msautotest.  thanks for the fix.

PR #5269, opened 2016-03-31
Heap errors with msEvalTextExpressionJSonEscape in maputfgrid.cpp

geographika:
When debugging I get invalid heap errors at the following line:

`msFree(r->data->table[i].datavalues);`

This is caused by the following line, which assigns a string on the stack rather than heap - hence the error when calling free()

`r->data->table[r->data->counter].datavalues = msEvalTextExpressionJSonEscape(&r->utflayer->utfdata, p);`

This pull request stops the error, but may not be the cleanest solution. 

Pull request also includes comment updates.

### DIFF
--- a/maputfgrid.cpp
+++ b/maputfgrid.cpp
@@ -285,7 +285,7 @@ band_type addToTable(UTFGridRenderer *r, shapeObj *p)
   utfvalue = encodeForRendering(utfvalue);
 
   /* Data added to the table */
-  r->data->table[r->data->counter].datavalues = msEvalTextExpressionJSonEscape(&r->utflayer->utfdata, p);
+  r->data->table[r->data->counter].datavalues = msStrdup(msEvalTextExpressionJSonEscape(&r->utflayer->utfdata, p));
 
   /* If UTFITEM is set in the mapfile we add its value to the table */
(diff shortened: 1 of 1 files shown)


rouault:
Could you give a way of reproducing the issue you see and/or a Valgrind log ? As far as I can see msEvalTextExpressionJSonEscape() already returns something that is allocated by msStrdup(), so it looks like your proposed patch would cause a memory leak. Perhaps that's covering an issue elsewhere

geographika:
@rouault - I think the errors in the above were caused by the issue outlined in https://github.com/mapserver/mapserver/pull/5277 - as with the updated pull request these errors no longer occur. 
The reallocation commit can therefore be ignored.
The [updated comments](https://github.com/mapserver/mapserver/pull/5269/commits/4e0d69b123c233bfafa049d1560f1b9690a66220) however could still be applied?

PS Congratulations on the OSGeo award!

rouault:
Commit https://github.com/mapserver/mapserver/pull/5269/commits/4e0d69b123c233bfafa049d1560f1b9690a66220 cherry-picked as https://github.com/mapserver/mapserver/commit/2f547e17c669b1cc8ef5d42e472988498903a899

Issue #4748, opened 2013-09-02
Python3 support


fbriol:
Hello,

I changed the Python wrapper to make it compatible with Python3. The patch file used is the fallowing:

``` diff
diff --git mapscript/python/CMakeLists.txt mapscript/python/CMakeLists.txt
index 33945b2..987c1bd 100644
--- mapscript/python/CMakeLists.txt
+++ mapscript/python/CMakeLists.txt
@@ -7,8 +7,8 @@ FIND_PACKAGE(PythonInterp)
 # interpreter that was found beforehand, and defaults to the system
 # python. We first try to find python.h and libpython.so ourselves
 # from the hints given by distutils and sys
-execute_process ( COMMAND ${PYTHON_EXECUTABLE} -c "from distutils.sysconfig import get_python_inc; print get_python_inc(True)" OUTPUT_VARIABLE PYTHON_INC OUTPUT_STRIP_TRAILING_WHITESPACE)
-execute_process ( COMMAND ${PYTHON_EXECUTABLE} -c "import sys; print sys.prefix" OUTPUT_VARIABLE PYTHON_PREFIX OUTPUT_STRIP_TRAILING_WHITESPACE)
+execute_process ( COMMAND ${PYTHON_EXECUTABLE} -c "from distutils.sysconfig import get_python_inc; print(get_python_inc(True))" OUTPUT_VARIABLE PYTHON_INC OUTPUT_STRIP_TRAILING_WHITESPACE)
+execute_process ( COMMAND ${PYTHON_EXECUTABLE} -c "import sys; print(sys.prefix)" OUTPUT_VARIABLE PYTHON_PREFIX OUTPUT_STRIP_TRAILING_WHITESPACE)

 find_path(PYTHON_INCLUDE_PATH Python.h
   HINTS ${PYTHON_INC}
@@ -37,7 +37,7 @@ set_target_properties(${SWIG_MODULE_pythonmapscript_REAL_NAME} PROPERTIES PREFIX
 set_target_properties(${SWIG_MODULE_pythonmapscript_REAL_NAME} PROPERTIES OUTPUT_NAME _mapscript)


-execute_process ( COMMAND ${PYTHON_EXECUTABLE} -c "from distutils.sysconfig import get_python_lib; print get_python_lib(True)" OUTPUT_VARIABLE PYTHON_SITE_PACKAGES OUTPUT_STRIP_TRAILING_WHITESPACE)
[...]

PR #5290, opened 2016-06-04
Add MapScript support for Python 3.x

klassenjs:
I haven't had time to test this much, but this allows MapScript to
build and run against Python 3.4. The changes appear to continue to
work with Python 2.7.  I kind of accidentally figured this out while playing
around so it may be incomplete, but I figured I'd share it as it is at least a
good start.

If there are multiple Python versions installed on your machine,
the Python version that is selected is whatever CMake finds first
or the path set with -DPYTHON_EXECUTABLE=

Changes include:
- print ""                     ---> print("")
- raise E, "text"              ---> raise E("text")
- PyString_FromStringAndSize() ---> PyBytes_FromStringAndSize()
- Remove references to PyFile*

### DIFF
--- a/mapscript/python/CMakeLists.txt
+++ b/mapscript/python/CMakeLists.txt
@@ -7,8 +7,8 @@ FIND_PACKAGE(PythonInterp)
 # interpreter that was found beforehand, and defaults to the system
 # python. We first try to find python.h and libpython.so ourselves
 # from the hints given by distutils and sys
-execute_process ( COMMAND ${PYTHON_EXECUTABLE} -c "from distutils.sysconfig import get_python_inc; print get_python_inc(True)" OUTPUT_VARIABLE PYTHON_INC OUTPUT_STRIP_TRAILING_WHITESPACE)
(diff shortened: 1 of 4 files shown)


jmckenna:
There is also an open ticket for Python3 support: ticket #4748

jmckenna:
I tried this, but I also needed the other setup.py changes made in #4748 (dict, readme)

soleboxy:
when i try it out with python3 i get the usual 
ImportError: dynamic module does not define init function (PyInit__mapscript)

... even tried to manually edit all the hex_version>= 0x03000 conditions to always be true in mapscriptPYTHON_wrap.c

sdlime:
So where are we at with this pull request? It's a little dated and should be based against master and not the 7.0 branch. --Steve

claudep:
I recreated this pull request against master in #5561, also including some setup.py changes from #4748.

sdlime:
Beautiful - anyone else want to comment or test? Otherwise I say we merge it and deal with issues as part of the release process. --Steve

klassenjs:
I didn't actually end up using this, so I don't have any specific comments/test cases.  Glad to see @claudep pick it up.

(This PR was fallout of experimenting with options for moving beyond MapScript/PHP 5 for GeoMoose 3, but the final decision was to use a combination of standard WFS calls and client side processing rather than MapScript.)

sdlime:
I'm going to close with with the newer pull request in play. --Steve

PR #1715, opened 2016-06-28
Line pattern in image mosaic when reprojecting

tbonfort:
**Reporter: ml.dje@geocontent.de**
**Date: 2006/03/15 - 12:58**
**Trac URL:** http://trac.osgeo.org/mapserver/ticket/1715

```
Creating a mosaic from seamless tiled images creates a regular pattern of lines
in the background color along the edges of the images. This only happens if the
image is requested in a projection other than the source projection.

Further:
1. only versions 4.8.x of mapserver are affected (4.6.1 works fine)
2. the pattern is only visible at certain combinations of image resolution, map
extent and map size.

Regarding 1: The gdal version is different (1.2 vs. 1.3). Maybe this is somehow
related.
Regarding 2: It seems that the effect preferably occurs if the images are scaled
down.
```

### DIFF
--- a/mapresample.c
+++ b/mapresample.c
@@ -1383,14 +1383,21 @@ int msResampleGDALToMap( mapObj *map, layerObj *layer, imageObj *image,
   nLoadImgYSize = MS_MAX(1, (int) (sSrcExtent.maxy - sSrcExtent.miny)
                       * (dfNominalCellSize / sDummyMap.cellsize));
 
-  /*
-  ** Because the previous calculation involved some round off, we need
-  ** to fixup the cellsize to ensure the map region represents the whole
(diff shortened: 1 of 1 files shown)


rouault:
@tbonfort How did you reproduce the issue ? I tried with https://trac.osgeo.org/mapserver/raw-attachment/ticket/1715/test.zip and 
"shp2img -m test_31468_large.map -o out_31438_large.png" and "shp2img -m test_31468.map -o out_31438.png" and I can't see any line artifact in the mosaic (with or without your patch, and the results are binarily identical in both cases)

rouault:
https://github.com/rouault/mapserver/commit/026847b1c962cdd666d28cfbe38ac40fff54ffbb merged in branch-7-0

PR #5306, opened 2016-07-30
Add support for DescribeFeatureType with gml_types=auto for the MSSQL…

szekerest:
… driver (#5305)

### DIFF
--- a/mapmssql2008.c
+++ b/mapmssql2008.c
@@ -38,6 +38,7 @@
 #include <assert.h>
 #include "mapserver.h"
 #include "maptime.h"
+#include "mapows.h"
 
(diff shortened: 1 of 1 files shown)


sethgirvin:
@szekerest - this is great work! I can get rid of 15Kb from my map files and it fixes a bug I had with MapInfo where NULL values caused problemts with WFS layers (the nillable property fixes this). 
Thanks!

geographika:
@szekerest - this is great work! I can get rid of 15Kb from my map files and it fixes a bug I had with MapInfo where NULL values caused problemts with WFS layers (the nillable property fixes this). 
Thanks!

PR #5277, opened 2016-08-27
strdup() is deprecated under Windows

tbonfort:
While stepping through the MapServer source code in a Visual Studio debugger, every call to [msStrdup](https://github.com/mapserver/mapserver/blob/branch-7-0/mapstring.c#L2121) caused the debugger to crash with heap errors. 

I saw in the code comments that `msStrdup` was originally copied from GDAL. Looking at the current GDAL source and issues I found that [CPLStrdup()](https://github.com/OSGeo/gdal/blob/bca12f7655fac8bc2734c1a8af1cad4f272450ac/gdal/port/cpl_conv.cpp#L261) had been updated for the same issue I am having with memory errors in Windows. 

See - https://github.com/OSGeo/gdal/pull/51 for details and discussions.

Could the MapServer source also be changed to match GDAL and allow debugging on Windows? I've not noticed any issues in production, but the VS debugger is much stricter with memory leaks. 

I think only [msStrdup](https://github.com/mapserver/mapserver/blob/branch-7-0/mapstring.c#L2128) need be changed as `strdup` is always wrapped in `msStrdup`, although `free` is used in several places, and should perhaps be `msFree`?


tbonfort:
@geographika please test this PR and comment

geographika:
@tbonfort - many thanks for looking into this. 

I got the branch-7-0 MapServer branch and ran the following command to get your changes:

`git pull git://github.com/tbonfort/mapserver.git issues/5277-free`

I then rebuilt (in debug mode with no optimizations). The debugger got stuck in a loop, after checking some of the changes I noticed there was a self-calling function in `mapfile.c` included as part of the pull request: 

```
#ifdef USE_MSFREE
void msFree(void *p)
{
  if(p) msFree(p);
}
#endif
```

I changed it to  `if(p) free(p);` rebuilt and was able to step through the entire process of creating a WFS response from start to end, so it appears the issues will be resolved with this patch. 

It will be a huge help to try and debug any other issues on Windows - so thanks again for looking at this.

geographika:
I have run through a few more code paths, which have triggered the same exceptions as I found originally. I was incorrect in my first post that all `strdup` calls were wrapped in `msStrdup` - there are several cases where `strdup` is used on its own e.g.

https://github.com/mapserver/mapserver/blob/branch-7-0/maputil.c#L1508
https://github.com/mapserver/mapserver/blob/branch-7-0/maputil.c#L1481

Would there be any side effects in replacing all these with `msStrdup` throughout the codebase?

tbonfort:
Seth, you can open a pr with the changes. There should be no side effect.

jmckenna:
I've stepped through branch-7-0 with Visual Studio 2015 with the Debugger, with no issues.  Thanks @geographika !

PR #5268, opened 2016-08-27
Legend MapServer 7.0.1 Label SIZE is ignored when using SIZEUNITS meters

tbonfort:
When you use SIZEUNITS meters instead of pixel in your LAYER the SIZE of your LEGEND is very small as long as you define only size. When you define MINSIZE and MAXSIZE the size is ok.

```
LEGEND
	STATUS ON
	KEYSIZE 30 30
	KEYSPACING 5 3  
	LABEL
		TYPE TRUETYPE
		FONT 'arial' 
		SIZE 10
		COLOR 119 179 0
	END 
END
```

* example using SIZE 10
legend_mapserv7_size10

* Example using SIZE 100 (legend comes back with a big white extra area at the left and bottom)
legend_mapserv7_size100

* when you add MINSIZE and MAXSIZE the Legend SIZE is ok

```
LEGEND
	STATUS ON
	KEYSIZE 30 30
	KEYSPACING 5 3  
	LABEL
		TYPE TRUETYPE
		FONT 'arial' 
		SIZE 10
                MINSIZE 10
                MAXSIZE 10
		COLOR 119 179 0
	END 
END
```

legend_mapserv_7_size_minsize_maxsize

### DIFF
--- a/maplegend.c
+++ b/maplegend.c
@@ -523,7 +523,7 @@ int msLegendCalcSize(mapObj *map, int scale_independent, int *size_x, int *size_
 
       if(*text) {
         initTextSymbol(&ts);
-        msPopulateTextSymbolForLabelAndString(&ts,&map->legend.label,msStrdup(text),lp->scalefactor*resolutionfactor,resolutionfactor, 0);
+        msPopulateTextSymbolForLabelAndString(&ts,&map->legend.label,msStrdup(text),resolutionfactor,resolutionfactor, 0);
         if(UNLIKELY(MS_FAILURE == msGetTextSymbolSize(map,&ts,&rect))) {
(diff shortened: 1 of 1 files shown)


jmckenna:
@astroidex can you confirm that thomas' fix posted here, works in your case?

jmckenna:
@astroidex can you confirm that thomas' fix posted here, works in your case? Or, can you post the sample mapfile + data, so I can test?

jmckenna:
@astroidex can you confirm that thomas' fix posted here, works in your case? Or, can you post the sample mapfile + data, so I can test?  (I'd like to get this included in the MapServer 7.0.2 release this week)

tbonfort:
applied to branch-7-0

PR #5012, opened 2016-08-27
Issue rendering compound polygons (PostGIS)

tbonfort:
Some arcs end up drawn as full circle. See the center of the attached image, the red circle.
pg_arcs
test-data, same eâs shown in image (sql dump) can be found here:
www.tydac.ch/download/compound_test.zip


jmckenna:
Tested successfully using @ezala's ovals, in PostgreSQL 9.5.4 and PostGIS 2.2.2

ttt

PR #5312, opened 2016-08-27
Fix travis builds

tbonfort:
shamelessly taken from gdal


tbonfort:
applied to branch-7-0 in 5ef69c9

PR #5313, opened 2016-08-30
Change all instances of strdup to msStrdup

geographika:
See issue #5277 - fixes Windows memory issues in conjunction with using `msFree`

I left the following in mapstring.c alone. I presume this shouldn't be called anyway if only `msStrdup` is used?

```
#ifndef HAVE_STRDUP
char  *strdup(char *s)
{
  char  *s1;
  ...
```


tbonfort:
- There's still a call to strdup in mapcopy.h
- The calls to strdup in mapscript should also probably be changed
- As you noted, you can remove the mapserver provided strdup() function and the HAVE_STRDUP cmake check

geographika:
Superseded by #5321

PR #5318, opened 2016-09-02
php mapscript build error with php 5.6.25

tbonfort:
5.6.23 builds fine, 5.6.25 fails with:

```
/usr/obj/ports/mapserver-7.0.1/mapserver-7.0.1/mapscript/php/error.c:71: error: called object 'zend_error' is not a function
```

pointing at https://github.com/mapserver/mapserver/blob/branch-7-0/mapscript/php/error.c#L71

this seems caused by this upstream change: http://git.php.net/?p=php-src.git;a=commitdiff;h=791a98eb1c66d2340b4e897ab60e4a6700435b5b

When unrolling the IF_GET_STRING macro, something blows.. tried with gcc 4.2.1 and clang 3.8.1, using libestdc++ 4.2.1 in both cases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: php
 php:
   - 5.5
-#  - 5.6
+  - 5.6
 
 env:
   global:
--- a/mapscript/php/error.c
+++ b/mapscript/php/error.c
@@ -31,6 +31,7 @@
 
 #include "php_mapscript.h"
 
+#if PHP_VERSION_ID >= 50625
 #undef ZVAL_STRING
 #define ZVAL_STRING(z, s, duplicate) do {       \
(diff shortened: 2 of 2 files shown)


landryb:
Im not sure that's a correct fix but this macro maze is beyond every horror i saw so far.

tbonfort:
@landryb it fixes the broken tests we were having with 5.6.25 on travis CI, thanks!

tbonfort:
This PR needs to be checked against php <=5.6.24 before merge

tbonfort:
applied to branch-7-0 in d5cc5840683bd

sdlime:
@jmckenna, should this be back ported to the 6.4 branch? --Steve

PR #5321, opened 2016-09-03
Change strdup to use MsStrdup throughout codebase

geographika:
See issue #5277 - fixes Windows memory issues in conjunction with using msFree

The new pull request hopefully addresses the comments from @tbonfort in a previous pull request: 

https://github.com/mapserver/mapserver/pull/5313#issuecomment-243698082

> - There's still a call to strdup in mapcopy.h
> - The calls to strdup in mapscript should also probably be changed
> - As you noted, you can remove the mapserver provided strdup() function and the HAVE_STRDUP cmake check

The removal of the replacement ``strdup`` function  in ``mapstring.c`` may mean if any new code doesn't use ``msStrdup`` then it will crash on Windows (or at least in debugger mode - ``strdup`` still exists even if it is deprecated).


jmckenna:
@geographika can you also create a pull request for this to mapserver:branch-7-0 ?

PR #5326, opened 2016-09-20
Label frame offset with leader line

tbonfort:
When using the LEADER for point labels, the GEOMTRANSFORM 'labelpoly' frame around the label gets an offset both vertically and horizontally. See the image below where the bottom-left label has a leader, but the green frame is moved a few pixels down and to the right. The label which rendered without a LEADER has the correct frame around it.

labeloffset

Here is a section of my mapfile for the relevant layer:

    LABEL 
        WRAP "!"
        ALIGN center
        TEXT ('[name]' + "!D116.7 " + '[ident]')
        MAXSCALEDENOM 2600064
        POSITION auto
        PARTIALS false
        SIZE 8 
        COLOR 100 150 100
        TYPE truetype 
        FONT vera
        FORCE false 
        STYLE
                GEOMTRANSFORM 'labelpoly'
                OFFSET 2 2
                COLOR 150 200 150
                OUTLINECOLOR 150 200 150
                WIDTH 1
                OPACITY 100
        END
        STYLE
                GEOMTRANSFORM 'labelpoly'
                COLOR 255 255 255
                OUTLINECOLOR 150 200 150
                WIDTH 1
                OPACITY 95
        END
    END
    LEADER
        GRIDSTEP 40
        MAXDISTANCE 1000
        STYLE
                COLOR 150 200 150
                WIDTH 2
        END
    END

Thanks,
Stephen

### DIFF
--- a/mapdraw.c
+++ b/mapdraw.c
@@ -2269,7 +2269,7 @@ void offsetAndTest(mapObj *map, labelCacheMemberObj *cachePtr, double ox, double
       }
     }
     for(j=0; j<ts->label->numstyles; j++) {
-      if(ts->label->styles[i]->_geomtransform.type == MS_GEOMTRANSFORM_LABELPOINT) {
+      if(ts->label->styles[j]->_geomtransform.type == MS_GEOMTRANSFORM_LABELPOINT) {
         scratch->poly = scratch_line;
         offset_label_bounds(ts->style_bounds[j], scratch, ox, oy);
         status = msTestLabelCacheCollisions(map, cachePtr, scratch, priority, label_idx);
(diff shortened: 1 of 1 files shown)


tbonfort:
I've attached a fix for this. There was also a bug on the offseted label computations, as in your example the red symbol should also be offseted instead of remaining on the original position.
If you don't need a symbol attached to that label you should be OK.
Support for offsetting only part of a label (i.e. in your example the text but not the red dot) is not implemented, you can contact me if that is something you want to fund. The use case implemented is to offset a whole complex symbol, e.g. a meteo forecast icon.

tbonfort:
applied to branch-7-0

jmckenna:
Tested with branch-7-0.  Here are the results, using that sample mapfile (which I see was initially written by me, ha, small world):

MapServer 7.01:

before

MapServer-branch-7-0

after

jmckenna:
Yikes, I prefer the old MapServer 7.0.1 results.  I find the branch-7-0 resulting map image confusing in terms of leader lines.

jmckenna:
Ah ok it is the labelframe, which does display properly in the branch-7-0 imagine.  sorry for the noise.

PR #5335, opened 2016-10-12
Unable to pass FEATURE ITEMS from URL

tbonfort:
We tried to pass a FEATURE via a URL parameter. This works perfectly for attributes WKT and TEXT of the FEATURE.
It doesn't seem to work for attribute ITEMS.

Example:

Layer definition in Mapfile:

<pre>LAYER
    NAME testLayer
    TYPE polygon
    STATUS on
    METADATA
       WMS_TITLE "TestLayer"
       WMS_SRS "EPSG:31468"
    END
    PROCESSING "ITEMS=zaehljahr"
    CLASSITEM "zaehljahr"
    CLASS
      EXPRESSION '0'
      NAME "0"
      STYLE
        OPACITY 60
    COLOR 0 255 0
    ANTIALIAS FALSE
      END
      TEXT "[zaehljahr]"
      LABEL
        TYPE truetype
        COLOR 255 255 255
        FONT arial
        SIZE 12
        ANTIALIAS false
        POSITION lc
        PARTIALS false
        BUFFER 0
      END
    END
    CLASS
      EXPRESSION '1'
      NAME "1"
      STYLE
        OPACITY 60
    COLOR 0 221 0
    ANTIALIAS FALSE
      END
      TEXT "[zaehljahr]"
      LABEL
        TYPE truetype
        COLOR 255 255 255
        FONT arial
        SIZE 12
        ANTIALIAS false
        POSITION lc
        PARTIALS false
        BUFFER 0
      END
    END
END</pre>

Request-URL:
http://localhost/cgi-bin/mapserv.fcgi?map=test.map&VERSION=1.1.1&LAYERS=testLayer&BBOX=4548756.4916980425,5451024.760221503,4548993.064391087,5451261.332914547&WIDTH=300&HEIGHT=300&SRS=EPSG:31468&FORMAT=image/png&STYLES=&TRANSPARENT=FALSE&SERVICE=WMS&REQUEST=GetMap&map.layer[testLayer]=FEATURE WKT "MULTIPOLYGON (((4548889.74 5451201.12, 4548897.895232 5451196.8590681, 4548903.707732 5451211.3590681, 4548925.645232 5451208.3590681, 4548919.020232 5451190.1715681, 4548932.21 5451186.42, 4548903.68 5451193.28, 4548889.74 5451201.12)))" ITEMS "1" END

Error-Message:

<pre>loadfeature(): Unknown identifier. Parsing error near (ITEMS):(line 1)</pre>

Thanks for looking into this.


tbonfort:
Attached patch should resolve the issue. Please confirm.

sdlime:
Since this input is so arbitrary I think we need to also update mapfile.c around line 950 to require a validation entry so it's not enabled by default. --Steve

tbonfort:
@sdlime validation should be for whole FEATURE, not just ITEMS? (I mean, changing/setting the geometry is just as dangerous as changing just the items)

sdlime:
You can't change anything about a feature and can only add them. Anyway, the the parser for the geometry portion of a feature already does specific validation (e.g. is a value a number or valid WKT), I'd think that's sufficient.

itt-de:
Many thanks for the fast fix. Unfortunately I don't have any build-infrastructure, so I cant't confirm until the next official release. We will certainly make good use of this in future.

tbonfort:
applied to branch-7-0 in 93cc1edab4

PR #5346, opened 2016-11-09
Fix PHP MapScript build on Windows with CMake and Visual Studio 2012 Express

AlexanderGabriel:
Hi,

i compiled php mapscript on Windows with Visual Studio 2012 Express and PHP-5.6.27. I used CMake, not the stuff from the documentation (i think, this is really outdated).
I'm not new to Mapserver and MapScript but new to github and not really experienced in this Windows-compiling stuff. I'm not quite sure if i'm doing this right but i'm looking forward to get feedback ;)

I had some Problems but  i think i found good solutions for it. It compiles now on Windows with PHP-5.6.26-ts-x86-vc11. I'd like to provide some patches to improve the build-process for other Windows users in the future.

There is already a pull-request for the patch from tbonfort but it is not merged yet. i included it, because building will not work without.

I tested on linux. phpinfo() shows the extension so it seams i didn't break anything in the build system. maybe there are better tests to do before merging.

Thank you.

Kind regards,
Alex


tbonfort:
Looks good to me, thanks @AlexanderGabriel. It would be really cool if you could create a wiki page here on github explaining the detailed steps you undertook to compile mapserver from source under windows, that we could send windows users to if/when they need to test a fix/patch on their own.

AlexanderGabriel:
That is a good Idea. I can do that. But wouldn't it be better to first merge this pull-request? Or should i point to my rep and the commits in the wiki-page until this got merged?
I've updated "README.WIN32" in / and at /mapscript/php (branch-7-0-UpdateWin32BuildInstructions in my rep) a few minutes ago. The content of the new readmes can be used for the wiki-page.
Should I open a new pull request for just updating docs or make a new pull-request for both, docs and patches to build on windows and we throw away this pull-request?

tbonfort:
Fixes applied to branch-7-0, thanks @AlexanderGabriel. Documentation updates can go in a new PR

Issue #4864, opened 2014-02-13
pole holes: raster reprojections make ozone holes in the poles


dshorthouse:
Example: http://www.simplemappr.net/map/1579

I had a patch for this in MapServer 5, which might be reworked to solve the same artifact in MapServer 6. https://github.com/dshorthouse/SimpleMappr/blob/4cad6dd9724e3d59d5046c0a18b137ca5a10f80f/patches/mapserver_polehole.patch

dshorthouse:
Anyone?

dshorthouse:
This is still an issue with MapServer 7.02

Issue #5348, opened 2016-11-10
Layers not always rendering with particular projections and extents


dshorthouse:
MapServer 7.0.2

For starters, I'm using a [NaturalEarth shp file](http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/10m/cultural/ne_10m_admin_0_map_units.zip), though this isn't directly relevant to the issue because I experience the same issue with the shp layer off; it just helped to include a shp file here so I could verify that mapserver was rendering the images.
 
This works as expected, with both layers, "countries", and "layer_0" rendering:

~~~~
MAP
  EXTENT 54922.8044507052 365450.649162042 4115803.01914984 2393632.20244337
  IMAGETYPE "png"
  NAME "simplemappr"
  SIZE 900 450
  STATUS ON
  UNITS METERS

  OUTPUTFORMAT
    NAME "png"
    MIMETYPE "image/png"
    DRIVER "AGG/PNG"
    EXTENSION "png"
    IMAGEMODE RGB
    TRANSPARENT FALSE
  END # OUTPUTFORMAT

  SYMBOL
    NAME "square"
    TYPE VECTOR
    FILLED TRUE
    IMAGE "square"
    POINTS
      0 1
      0 0
      1 0
      1 1
      0 1
    END
  END

  PROJECTION
    "proj=laea"
    "lat_0=90"
    "lon_0=0"
    "x_0=0"
    "y_0=0"
    "ellps=WGS84"
    "datum=WGS84"
    "units=m"
    "over"
    "no_defs"
  END # PROJECTION

  LAYER
    DATA "/Users/dshorthouse/Sites/SimpleMappr/mapserver/maps/10m_cultural/10m_cultural/ne_10m_admin_0_map_units"
    METADATA
      "wms_srs"	"epsg:4326 esri:102009 esri:102015 esri:102014 esri:102012 esri:102024 epsg:3112 epsg:102017 epsg:102019"
      "wfs_extent"	"-180 -90 180 90"
      "wfs_srs"	"epsg:4326 esri:102009 esri:102015 esri:102014 esri:102012 esri:102024 epsg:3112 epsg:102017 epsg:102019"
      "gml_featureid"	"OBJECTID"
      "wfs_encoding"	"CP1252"
      "wfs_title"	"countries"
      "wfs_typename"	"countries"
      "gml_include_items"	"all"
      "wms_title"	"countries"
    END # METADATA
    NAME "countries"
    PROJECTION
[...]

PR #5360, opened 2016-12-13
Repeat transformation if extent is extreme

dshorthouse:
Repeat transformation for a rectangle interior to the output requested region. If the latter results in a more extreme y extent, then extend extents in source layer projection to southern/northing bounds and entire x extent. This fix resolves issue #4864.


dshorthouse:
This issue is still present in 7.0.4 and the PR continues to solve the issue.

tbonfort:
does this also solve #5348 ?

dshorthouse:
No, it solves #4864.

sdlime:
Any objection to applying this? Looks to be relatively isolated and all tests pass. --Steve

jmckenna:
@dshorthouse thanks for this patch, nice work on #4864 (a long time coming ha).

PR #5362, opened 2016-12-22
Added C-style multi-line comment support to the lexer.

sdlime:
Initial crack at supporting C-style comments. Both single line (e.g. /* foo */) and multi-line comments work. Only casualty is that a regex of the form /*.../ can't be used since those would be recognized as the start of a comment. See ticket #4097.


jmckenna:
This can be merged.

sdlime:
Jeff, I added support for C-style single line comments too. So we have Perl-style and C-style. Give it a shot... --Steve

jmckenna:
Steve it works great!

PR #5370, opened 2017-01-16
Use include paths from php-config when include path not found.

sebastic:
Fixes build failure with PHP 5 PPA packages, introduced by the changes in https://github.com/mapserver/mapserver/commit/54d645c18d0293f7e5821e9d740542408083f955.

OSGeo-Live uses the PHP 5 PPA packages to support PHP MapScript on Ubuntu xenial, but was unable to update the mapserver package due to this build failure.


jmckenna:
Thanks @sebastic

PR #5371, opened 2017-01-17
make wms layer groups requestable

tbonfort:
In the current layer grouping implementation, a layer with wms_layer_group=/g will only let its group "g" appear as requestable (i.e. with a ```<name>g</name>``` in capabilities) if there is also a LAYER in the mapfile with NAME=g.

I'd argue that this is unintuitive as when you are creating a hierarchy of layers you end up being obliged to define dummy/empty layers in order to get that hierarchy to be actually useful. e.g. see https://github.com/mapserver/mapserver/blob/rel-7-0-4/msautotest/wxs/wms_inspire.map#L257 

The WMS spec (wms 1.1 - 7.1.5.4) does provision for the case where a group of layers appears but cannot be called if it has only a ```<title>``` defined but not a ```<name>```. This is what we are implementing currently in the capabilities, but we are not respecting it in GetMap requests (we happily render the group even if it has no ```<name>``` defined).

In order to be coherent with the concept of layer grouping, be coherent with our actual behavior, and simplify writing grouped mapfiles, I propose we modify the current behavior and always add a ```<name>``` tag to group layers.

- [x] patch to add ```<name>``` tag to group layers
- [x] capabilities test
- [x] request test
- [ ] documentation update
- [ ] HISTORY.TXT update
- [ ] migration guide update


tbonfort:
cc @yjacolin

tbonfort:
see also #5220

yjacolin:
@tbonfort do you need a feedback from me?

sdlime:
Is this one ready to merge?

yjacolin:
If we can help in any way, let's us know.

ebelo:
Hi, is this ready to merge? Can we do something so that it gets into master?

yjacolin:
@tbonfort I am working on the doc parts of this feature. I created a new branch up-to-date with master and cherry-pick your 3 commits, added an item in HISTORY.TXT. Are you ok to close this PR and I will open a new one with a link to the doc. Thank for you answer.

yjacolin:
MIGRATION_GUIDE.txt doesn't need to be updated.

yjacolin:
following discussion in Mapserver-dev list, a new PR superseed this one: https://github.com/mapserver/mapserver/pull/5533

PR #5376, opened 2017-01-24
Vector Tile Support

sdlime:
Based on Thomas' vector-tiles branch. Fixed an alignment issue, added integration with mode=tile and updated to version 2.1 or the spec - basically switched out the .proto file. Tested against a couple of the Leaflet mvt plugins. --Steve

TODO:
- [x] Fix buffering issues in tile mode
- [ ] Add autotests
- [ ] Create documentation page
- [x] Create RFC


mapmvt.c
```diff
@@ -278,22 +271,9 @@ int msMVTWriteFromQuery( mapObj *map, outputFormatObj *format, int sendheaders )
   void *buf;
   const char *mvt_buffer = msGetOutputFormatOption(map->outputformat, "EDGE_BUFFER", "10");
   int buffer = MS_ABS(atoi(mvt_buffer));
-  double res;
-  rectObj bbox;
   VectorTile__Tile mvt_tile = VECTOR_TILE__TILE__INIT;
   mvt_tile.layers = msSmallCalloc(map->numlayers,sizeof(VectorTile__Tile__Layer*));
 
-
```
tbonfort: @sdlime, I see why you were getting alignment issues. This block of code needs to be kept, however the query.rect should be extended before the query, as is done in the WFS part: https://github.com/mapserver/mapserver/pull/5376/commits/032e5d3215d88a8519b0ef2405ff9a0c49eab998#diff-8b3f151aa0bfca8681cd277ced6c85cfR2619 . I know this isn't clean, but the buffering logic (similar to metabuffering when tiling) is essential IIRC for correct rendering.

sid-kap:
Hi! I was looking into MapServer and might be interested in using this feature. Is this still being developed, or has the PR been abandoned?

sdlime:
Hi @sid-kap... The pull-request has not been abandoned. Just trying to get it merged. --Steve

sdlime:
@rouault, @tbonfort, @jmckenna: the tests are failing because the protobuf output format is now appearing in a bunch of get capabilities output as a supported format (e.g. <Format>application/x-protobuf</Format>). I can just fix all those results -- or should would make that an optional format?

rouault:
I don't think there are more reasons to make it optional than the other formats

itbeyond:
I am really keen to get this running as soon as possible and have a project just waiting for this functionality. I am not sure if I can help but if there is something I can do to help get this functionality running I am happy to put in some time/effort.

sdlime:
@rouault where the integration tests are crapping out doesn't look to be failing because my pull request (it's generating a PNG), is there a way to examine the Travis or AppVeyor output files? --Steve

rouault:
@sdlime I'm not aware of a way of extracting the files from Travis/AppVeyor. I'd just suggest that you run msautotest from Vagrant, which uses the same library version as Travis. That's what I do usually

sdlime:
@rouault what version of vagrant and virtual box are you using? Stefan's docs reference versions that are a tad old.

rouault:
The ones of Ubuntu 16.04: vagrant 1.8.1 + virtualbox 5.0.40. But the versions of ubuntu 14.04 also worked

sdlime:
@rouault, a local vagrant up dies since it can't find protobufc yet Travis seems to have that installed. Do you know where to affect that? --Steve

rouault:
@sdlime In scripts/vagrant/packages.sh, add libprotobuf-dev or whatever the package is named to the apt-get install final line

itbeyond:
Compiled on Windows X64 platform and can confirm it is all working - there are some issues around Windows support for protobuf-c however I found a precompiled Windows binary that seems to work (See details here: https://github.com/protobuf-c/protobuf-c/issues/168 - download provided at the end of the comments).

I have tested it with the mvt-demo published by @sdlime and will now start working on my data and looking at mapscript operation. Thanks guys great work and perfect timing for getting this running.

itbeyond:
@sdlime I have been doing extensive testing with mvt support against a very large (10K lines) map file and converting the png output to mvt with mixed results. Of course there are heaps of DB changes and updates I am applying however I have just started working with Polyline and cannot get the alignment between tiles to work - I have looked at and tried the EDGE_BUFFER from values of 0 to 50 and still see the same type of alignment issues my OUTPUTFORMAT statement is as follows:
```
OUTPUTFORMAT
    NAME "mvt"
    MIMETYPE "application/x-protobuf"
    DRIVER "mvt"
    EXTENSION "pbf"
	FORMATOPTION "EDGE_BUFFER=50"
  END # OUTPUTFORMAT
```
I am using mode=tile and pulling a road layer from postgres. The png output is fine and is cut in the right place - so using the same mapserve cmd and only changing the outputformat from png to mvt (mvt rendered in inspect mode on Maputnik, shows the same in mapbox gl) produes the images attached. NOTE the mvt layer is only showing major roads hence there is less visible in the mvt output than the png. There errors in the mvt are visible centre and left - there is a 4 way tile join.
mapserver_mvt
mapserver_png

I will endeavour to produce a minimal example however I suspect any major Polyline output will render this way (these all produced with the EDGE-BUFFER=50 however it seems to do nothing different if I set at 0 or default to 10 etc.

itbeyond:
Oh and I just booted the mvt-demo provided by @sdlime and the error exists in this output also as per the attached screenshot. You can see it in the road line and river line.
demo

PR #5381, opened 2017-02-02
Fix generated C# module

NathanW2:
Fixes the name of the C# module.  Should be `mapscript.dll` however the generated code currently looks for `csharpmapscript.dll` so will throw an error when loaded.

Seems it's only the cmake file that has this result.

@szekerest thoughts?


jmckenna:
makes sense to me.  Thanks @NathanW2 !

Issue #5404, opened 2017-03-07
add metadata to optionally not emit root layer name in WMS Capabilities XML


tomkralidis:
A sample GetCapabilities request shows the following:

```xml
...
<Capability>
...
<Layer>
 <Name>foo</Name>
 <Title>bar</Title>
...
```

The root layer element's `foo` value is derived from the mapfile's `MAP.NAME` value.  WMS says any `Layer` with a `Name` element should be `GetMap` capable.  In our case this requests all layers in the WMS, which may prove problematic for a number of reasons:

1. potential performance issue
1. if one layer is broken/not functioning, will result in error
1. even on a successful request, the result is not meaningful/valuable

The workaround is to set `MAP.NAME` to `""`.  The proper fix is to tell MapServer to not emit the root layer name via something like `"wms_rootlayer_name_enabled" "false"`

jratike80:
Hi,


> The proper fix is to tell MapServer to not emit the root layer name via something like wms_rootlayer_name_enabled


For my mind the default value should be to not emit the root layer name because in most cases users do not really want to do that. Same discussion was going on in the Geoserver issue tracker some time ago https://osgeo-org.atlassian.net/browse/GEOS-7935.


-Jukka Rahkonen-

tomkralidis:
cc @dmorissette @rouault for thoughts.

@jratike80 I would tend to agree however I'm not sure how many installations utilize the functionality (imagine a 2 layer mapfile where is it easy to just call the root layer) already, so the concern is breaking existing applications.

tomkralidis:
FYI PR in #5405 pending consensus.

jratike80:
Copy of the the comment that I wrote to PR follows because I think this is where it belongs.

I think that also the WMS documentation http://mapserver.org/ogc/wms_server.html should be updated from this place:

> Map Name and wms_title:
> WMS Capabilities requires a Name and a Title tag for every layer. The Map’s NAME and wms_title metadata will be used to set the root layer’s name and title in the GetCapabilities XML output. The root layer in the WMS context corresponds to the whole mapfile.

Did you consider an alternative to add an explicit new metadata item "wms_name" that would override the name of the mapfile? The same effect than by setting
`"wms_rootlayer_name_enabled" "yes"`
would be obtained then by setting an empty wms_name
`"wms_name" ""`
That might be more user friendly because at the moment and with this change also in the future the WMS title is set in metadata but WMS name in a different and sort of non-logical place.

[...]

PR #5405, opened 2017-03-08
add metadata to optionally not emit root layer name in WMS Capabilities

tomkralidis:
As per #5404.  Pending consensus.

### DIFF
--- a/mapwms.c
+++ b/mapwms.c
@@ -2686,6 +2686,7 @@ int msWMSGetCapabilities(mapObj *map, int nVersion, cgiRequestObj *req, owsReque
   const char *updatesequence=NULL;
   const char *sldenabled=NULL;
   const char *layerlimit=NULL;
+  const char *rootlayer_name=NULL;
   char *pszTmp=NULL;
   int i;
   const char *format_list=NULL;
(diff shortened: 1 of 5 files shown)


jratike80:
I think that also the WMS documentation http://mapserver.org/ogc/wms_server.html should be updated from this place:

> Map Name and wms_title:
> WMS Capabilities requires a Name and a Title tag for every layer. The Map’s NAME and wms_title metadata will be used to set the root layer’s name and title in the GetCapabilities XML output. The root layer in the WMS context corresponds to the whole mapfile.

Did you consider an alternative to add an explicit new metadata item "wms_name" that would override the name of the mapfile? The same effect than by setting
`"wms_rootlayer_name_enabled" "yes"`
would be obtained then by setting an empty wms_name
`"wms_name" "" `
That might be more user friendly because at the moment and with this change also in the future the WMS title is set in metadata but WMS name in a different and sort of non-logical place.

dmorissette:
I agree with @jratike80's comment and suggestion of setting "wms_name" "" instead of creating a new "wms_rootlayer_name_enabled" metadata.

PR #5409, opened 2017-03-15
Add dll version stamp in Windows

NathanW2:
This commits adds a version stamp to the mapserver.dll on Windows like the following: Just a nice touch more than anything else.

image


jmckenna:
Nice addition @NathanW2 !  Thanks!   (I already do something like this for the MS4W community)

NathanW2:
@jmckenna thanks for the merge.  I'm going to improve it so it can also include the git sha as well hopefully because I have found that handy with my other projects when doing dev builds for testing.

NathanW2:
@jmckenna do you want me to make a commit on the 7 branch as well?

jmckenna:
@NathanW2 yes to branch-7-0 also would be good, might as well

PR #5410, opened 2017-03-26
Fixed crash of mapserv due to memory corruption at large values of double variables

sergmorev:
I'll explain the reason for editing with mappostgis.c, starting at line 3640.

 In this case, for values node->tokenval.dblval> = 123456789012 (such values are real in the field),
strlen(snippet)> = 19, and the memory is requested in the program only for 19 bytes. The unsolicited
memory is overwritten.

 It was possible without a check to request 318 bytes, this is enough to store number, when converting

to char [] having the longest string, - numeric_limits <double> :: max (),  but it is a pity to spend
memory when it is likely that most of the values are less of this number.

 Other edits are associated with preventing similar errors. Only there memory is sometimes requested
in other quantities.

 Developers should pay attention to other pieces of software, where it is used sprintf( ..., "%lf",...).


mapserver.h
```diff
@@ -243,6 +243,9 @@ extern "C" {
 #define MS_DEG_TO_RAD .0174532925199432958
 #define MS_RAD_TO_DEG   57.29577951
 
+#define MS_MAX_DOUBLE 1.7976931348623157e+71
```
rouault: Where do those magic values come from ?

mapmssql2008.c
```diff
@@ -2389,7 +2389,10 @@ int process_node(layerObj* layer, expressionObj *filter)
     case MS_TOKEN_LITERAL_BOOLEAN:
     case MS_TOKEN_LITERAL_NUMBER:
       strtmpl = "%lf";
-      snippet = (char *) msSmallMalloc(strlen(strtmpl) + 16);
+      if (layerinfo->current_node->tokenval.dblval>MS_MAX_DOUBLE || layerinfo->current_node->tokenval.dblval<MS_MIN_DOUBLE)
```
rouault: Couldn't we have a char* msPrintLf(double) function that could be reused in all those multiple locations ?

PR #5416, opened 2017-03-29
Memory overwrite ( Calling msDoubleToString () for memory requests )

sergmorev:
Calling msDoubleToString () for memory requests to the string of sprintf (...,"% lf",doubleValue).
This is the answer to the proposal rouault make a function char* msPrintLf(double).


sergmorev:
The test gives errors of type:
-                <gml:pos>-75.000000 45.000000</gml:pos>
+                <gml:pos>-75 45</gml:pos>

I'm not strong in gml, however in my gml both varinata are allowed. In any case, by reference
http://www.gdal.org/drv_gml.html in  <gml:pos> integers are used, not fractional numbers.

sdlime:
Reviewers? Any feedback? The two failing tests are related to PHP.

mapstring.c
```diff
@@ -371,14 +371,22 @@ char *msLongToString(long value)
...
-  size_t bufferSize = 256;
-  char *buffer = (char*)msSmallMalloc(bufferSize);
-
-  if (force_f == MS_TRUE)
-    snprintf(buffer, bufferSize, "%f", value);
-  else
-    snprintf(buffer, bufferSize, "%g", value);
-  return(buffer);
+  char *buffer;
+  if (value==(int)value && value>=INT_MIN && value<=INT_MAX) {
```
rouault: Perhaps drop that integer case ?

jmckenna:
@sergmorev can your rebase this against master (instead of branch-7-0) ?

rouault:
Superseded with https://github.com/MapServer/MapServer/pull/6226

Issue #4918, opened 2014-05-01
GRASS 7.0 NoData handling


HuidaeCho:
GRASS 7 uses NaN as a no data value for floating-point rasters and msGetGDALNoDataValue correctly returns NaN. When comparing pixel values with the NaN no data value, mapdrawgdal.c failes to find no data pixels because it simply uses the equal comparison. Because of this issue, dfScaleMin and dfScaleMax are set to a NaN and, as a result, COLORRANGE & DATARANGE don't work properly and an empty image is returned. A simple example is:

```
MAP
  SIZE 600 600
  EXTENT 2029205.71251009 1483554.15327001 2036686.24774584 1494117.98023943
  LAYER
    TYPE RASTER
    STATUS DEFAULT
    DATA "/home/grassdata/epsg102667/tmp/cellhd/contourareas"
    CLASS
      STYLE
        COLORRANGE 0 0 0 255 255 255 # black to white
        DATARANGE 702 768 # min value to max value
      END
    END
  END
END
```

PR #5419, opened 2017-04-09
Fix SCALE=AUTO (#4927); No data in GRASS (#4918)

### LINKED ISSUE #4927: PROCESSING SCALE doesn't scale colors
PROCESSING "SCALE=*" scales pixel values, but doesn't scale the corresponding colors. For example, originally
```
pixel	color
0		10,10,10
100		0,0,100
1000	0,100,0
2500	100,0,0
```
If I use SCALE=AUTO, the pixel values are scaled such that they are from 0 to 255, but the colors are not scaled:
```
pixel	color
0		10,10,10
10		0,0,0 # it has to be 0,0,100 because pixel value 100 was scaled to 10.
100		0,0,100 # it has to be 0,100,0.
250		0,0,0 # it has to be 100,0,0.
```
Because of the above issue, the map is rendered differently from the original map. Expected results are:
```
pixel	color
0		10,10,10
10		0,0,100
100		0,100,0
250		100,0,0
```

HuidaeCho:
This pull request fixes #4927 & #4918.

### DIFF
--- a/mapdrawgdal.c
+++ b/mapdrawgdal.c
@@ -49,13 +49,18 @@ extern int InvGeoTransform( double *gt_in, double *gt_out );
 #include "gdal_alg.h"
 
 static int
+IsNoData( double dfValue, double dfNoDataValue, int bIsNoDataNan );
+
+static int
 LoadGDALImages( GDALDatasetH hDS, int band_numbers[4], int band_count,
                 layerObj *layer,
                 int src_xoff, int src_yoff, int src_xsize, int src_ysize,
                 GByte *pabyBuffer,
                 int dst_xsize, int dst_ysize,
                 int *pbHaveRGBNoData,
-                int *pnNoData1, int *pnNoData2, int *pnNoData3 );
+                int *pnNoData1, int *pnNoData2, int *pnNoData3,
+                int *scaled, double *scaleMin, double *scaleRatio,
+                unsigned char **isNoDataBuffer );
 static int
 msDrawRasterLayerGDAL_RawMode(
   mapObj *map, layerObj *layer, imageObj *image, GDALDatasetH hDS,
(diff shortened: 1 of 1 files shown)


mapdrawgdal.c
```diff
@@ -1937,9 +1995,8 @@ msDrawRasterLayerGDAL_16BitClassification(
       /*
        * Skip nodata pixels ... no processing.
        */
-      if( bGotNoData && fRawValue == fNoDataValue ) {
+      if( bGotNoData && IsNoData(fRawValue, fNoDataValue, bIsNoDataNan) )
         continue;
-      }
 
       if( CPLIsNan(fRawValue) )
```
tbonfort: isnan has already been checked for here

mapdrawgdal.c
```diff
@@ -1983,6 +2040,16 @@ msDrawRasterLayerGDAL_16BitClassification(
...
 
 /************************************************************************/
+/*                          IsNoData()                                  */
+/************************************************************************/
+static int
+IsNoData( double dfValue, double dfNoDataValue, int bIsNoDataNan )
+{
+  return (bIsNoDataNan && isnan(dfValue)) ||
+         (!bIsNoDataNan && dfValue == dfNoDataValue);
+}
```
tbonfort: wouldn't dfValue==dfNodataValue || isnan(dfValue) be sufficient here?

HuidaeCho: If bIsNoDataNan is true, you cannot simply compare two NaN's like that.

> NaN is unordered: it is not equal to, greater than, or less than anything, including itself. x == x is false if the value of x is NaN. [https://www.gnu.org/software/libc/manual/html_node/Infinity-and-NaN.html](url)

HuidaeCho: Hmm.. I thought about this more and yes your check is safer. If NoData is not NaN, my check returns No when dfValue is NaN (if that ever happens) because NaN == non-NaN is false. Logically, returning No is correct because dfValue (NaN) is different from non-NaN no data value, but proceeding with NaN dfValue further is not a good idea if dfValue gets ever assigned NaN for whatever reasons when NoData is not NaN.

mapdrawgdal.c
HuidaeCho: Yes, CPLIsNan is already there. IsNoData is still required to check if fRawValue is No Data when NoData is not NaN. Then CPLIsNaN is redundant.

jmckenna:
@HuidaeCho can you rebase this against master (instead of branch-7-0) ?

rouault:
Superseded per https://github.com/MapServer/MapServer/pull/6251

Issue #5202, opened 2015-12-01
BBOX is not re-projected properly in MapServer query statement


marionb:
While using MapServer I came across some strange behaviour when doing a POST request on a BBox. From the client the following request is sent:

```
<wfs:GetFeature xmlns:wfs="http://www.opengis.net/wfs" service="WFS" version="1.1.0" maxFeatures="200" xsi:schemaLocation="http://www.opengis.net/wfs http://schemas.opengis.net/wfs/1.1.0/wfs.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <wfs:Query typeName="feature:cameras" xmlns:feature="http://mapserver.gis.umn.edu/mapserver">
        <ogc:Filter xmlns:ogc="http://www.opengis.net/ogc">
            <ogc:BBOX>
                <ogc:PropertyName>geom</ogc:PropertyName>
                <gml:Envelope xmlns:gml="http://www.opengis.net/gml">
                    <gml:lowerCorner>565240 5735152.5</gml:lowerCorner>
                    <gml:upperCorner>565895 5735602.5</gml:upperCorner>
                </gml:Envelope>
            </ogc:BBOX>
        </ogc:Filter>
    </wfs:Query>
</wfs:GetFeature>
```

The returned xml contains no information (```<gml:Null>missing</gml:Null>```) also if srsName="EPSG::3857" is sent with the request.

When checking the query log from PostQGIS the following query I can find the reason for the empty gml:

``` SQL
SELECT ...
FROM
  (SELECT cam.*
   FROM shema.cameras cam
   WHERE (etage = '+0'
          OR etage IS NULL)) AS foo
WHERE geom && ST_GeomFromText('POLYGON((250840.933537713 -5541337.97732029,250840.933537713 -5541337.97732029,250840.933537713 -5541337.97732029,250840.933537713 -5541337.97732029,250840.933537713 -5541337.97732029))',27562)
  AND (st_intersects(geom,ST_GeomFromText('POLYGON ((813376.7942609430756420 83142.7202183806366520, 813359.1764772359747440 83647.8472906547685852, 813645.7609626927878708 83657.8494853072916158, 813663.4024164262227714 83152.7232393100421177, 813376.7942609430756420 83142.7202183806366520))',27562)) = TRUE)
```
In this query the first polygon is not properly transformed from projection 3857 back in to 27562 such that the WHERE clause always returns FALSE.

The following Configurations are used:

[...]

PR #5430, opened 2017-05-14
Add two tests to cover the out-of-extent intersects filter

juliensam:
#5202 : WFS Intersects filters are limited by the extent defined in the mapfile. I added 2 tests that play with the extent of the map to return no results and 2 results.


dmorissette:
For the record, the MapServer WFS interface only returns data that overlaps with the MAP EXTENT by default. So in a case where the MAP EXTENT does not overlap with the INTERSECTS FILTER then it is normal to return no results. This is what those new test cases are checking.
This came up as we were trying (without success) to reproduce the issue reported in #5202.

PR #5451, opened 2017-06-28
Allow using double quoted aliases for Oracle column aliases

micking:
Adding double quotes to the columns of the mapserver generated outer
query safely allows using double quoted column aliases in the data
statement
By that we can have mixed capital column names


msmitherdc:
@sdlime This can be merged in

PR #5452, opened 2017-06-28
Fix enable out of source build path with spaces

micking:
this fix allows to have an out-of-source build on any location - previously it was assumed that the mapscript csharp sources had a fixed path relative to the build
In addition we can now have build path names with spaces


jmckenna:
thanks @micking

PR #5454, opened 2017-06-28
Enabled WFS sortby for Oracle plugin

micking:
Enabled WFS sortby for Oracle plugin, Previously that was only available if Oracle support was statically linked to mapserver.

additionally we now support a quoted sortby parameter to enable sorting by non-capitalized column names

### DIFF
--- a/maplayer.c
+++ b/maplayer.c
@@ -1604,9 +1604,19 @@ char* msLayerBuildSQLOrderBy(layerObj *layer)
   if( layer->sortBy.nProperties > 0 ) {
     int i;
     for(i=0;i<layer->sortBy.nProperties;i++) {
-      char* escaped = msSmallMalloc(strlen(layer->sortBy.properties[i].item) + 3);
-      sprintf(escaped, "\"%s\"", layer->sortBy.properties[i].item);
-      if( i > 0 )
+      char* escaped = msLayerEscapePropertyName(layer, layer->sortBy.properties[i].item);
+      //Enclose property name in double quotes (if it isn't yet) to ensure that mixed case property names are supported
+      if (escaped[0] != '"')
+      {
+        size_t propertyLen = strlen(escaped);
+        escaped = msSmallRealloc(escaped, propertyLen + 3);
+        memmove(escaped + 1, escaped, propertyLen);
+        escaped[0] = '\"';
+        escaped[propertyLen+1] = '\"';
+        escaped[propertyLen + 2] = 0;
+
+      }
+      if (i > 0)
         strOrderBy = msStringConcatenate(strOrderBy, ", ");
       strOrderBy = msStringConcatenate(strOrderBy, escaped);
       if( layer->sortBy.properties[i].sortOrder == SORT_DESC )
--- a/maplayer.c
+++ b/maplayer.c
@@ -1564,7 +1564,7 @@ int msLayerApplyPlainFilterToLayer(FilterEncodingNode *psNode, mapObj *map, int 
 int msLayerSupportsSorting(layerObj *layer)
 {
   if (layer && (
-         (layer->connectiontype == MS_OGR) || (layer->connectiontype == MS_POSTGIS) || (layer->connectiontype == MS_ORACLESPATIAL)
+    (layer->connectiontype == MS_OGR) || (layer->connectiontype == MS_POSTGIS) || (layer->connectiontype == MS_ORACLESPATIAL) || ((layer->connectiontype == MS_PLUGIN) && (strstr(layer->plugin_library,"msplugin_oracle") != NULL))
                )
      )
     return MS_TRUE;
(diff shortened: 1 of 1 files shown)


maplayer.c
```diff
@@ -1604,7 +1604,8 @@ char* msLayerBuildSQLOrderBy(layerObj *layer)
   if( layer->sortBy.nProperties > 0 ) {
     int i;
     for(i=0;i<layer->sortBy.nProperties;i++) {
-      char* escaped = msLayerEscapePropertyName(layer, layer->sortBy.properties[i].item);
+      char* escaped = msSmallMalloc(strlen(layer->sortBy.properties[i].item) + 3);
```
rouault: This change seems wrong. Proper escaping is lost. SQL injection could potentially occur

micking: Tx Even, you are totally right!
But we need to enclose the sort property by double quotes - without this Oracle throws an error if the Oracle column or the alias aren't upper case.
What do you think about using the original `msLayerEscapePropertyName` and then enclosing the escaped string in double quotes (if it isn't yet enclosed), like the following:

```
     char* escaped = msLayerEscapePropertyName(layer, layer->sortBy.properties[i].item);
     //Enclose property name in double quotes (if it isn't yet) to ensure that mixed case property names are supported
      if (escaped[0] != '"')
      {
        size_t propertyLen = strlen(escaped);
        escaped = msSmallRealloc(escaped, propertyLen + 3);
        memmove(escaped + 1, escaped, propertyLen);
        escaped[0] = '\"';
        escaped[propertyLen+1] = '\"';
        escaped[propertyLen + 2] = 0;
      }
```
I've added a corresponding commit to the branch

maplayer.c
```diff
@@ -1604,9 +1604,19 @@ char* msLayerBuildSQLOrderBy(layerObj *layer)
   if( layer->sortBy.nProperties > 0 ) {
     int i;
     for(i=0;i<layer->sortBy.nProperties;i++) {
-      char* escaped = msSmallMalloc(strlen(layer->sortBy.properties[i].item) + 3);
-      sprintf(escaped, "\"%s\"", layer->sortBy.properties[i].item);
-      if( i > 0 )
+      char* escaped = msLayerEscapePropertyName(layer, layer->sortBy.properties[i].item);
+      //Enclose property name in double quotes (if it isn't yet) to ensure that mixed case property names are supported
+      if (escaped[0] != '"')
```
rouault: Why is it needed ? msOracleSpatialEscapePropertyName() should do that. Actually msOracleSpatialEscapePropertyName shouldn't be defined at all and just rely on the default quoting mechanism

micking: `msOracleSpatialEscapePropertyName` is not enclosing a propertyname with double quotes - it just duplicates enclosing double quotes and backslashes, so e.g. it modifies `"PropertyName"` into `""PropertyName""`. Don't ask me if this really makes sense...
So could you imagine that we just modify `msOracleSpatialEscapePropertyName` to do what you think it should do? e.g. modifies `PropertyName` into `"PropertyName"`?

rouault: My suggestion would be to just remove completely msOracleSpatialEscapePropertyName, so that the default implementation that does quoting and escaping mechanism is used.

micking: Tried this (easy fix) and it works, tx for the suggestion. 
I was wondering why there was the special version of msOracleSpatialEscapePropertyName. I'm guessing that some developers were interested in using the "lazy" Oracle naming where `COLUMN` and `column` and `Column` all mean the same as Oracle internally translates all non-double-quoted names to uppercase, only `"column"` really means the lower case variant.
As the side effect of the change users will now be unable to use `column` for an (e.g. WFS) parameter name when in Oracle it's `COLUMN`.

rouault: Ah, good point. Didn't know that Oracle was case sensitive... I'll defer to @msmitherdc which is our Oracle expert...

msmitherdc: I think this is a good change. oracle is case sensitive when the columns are not uppercase.

micking: tx, so I'll undo my initial changes on the escaping stuff and just de-activate msOracleSpatialEscapePropertyName, that's fine. @msmitherdc you created the function - should I remove it?
With regards to git should I create an undo commit for my initial work on the escaping? I guess we cannot hide the first commit as it is already published in the branch used for the pull request?
Or should I drop the branch and create a new "clean" one?

msmitherdc:
@sdlime i think this is ready to merge in.

micking:
I just noticed that this pull request had been merged before I had modified my changes to be in sync with Even's suggestion to remove msOracleSpatialEscapePropertyName and undo my additional code to escape the properties.
I've modified my original branch to reflect this - can this pull request be re-opened and merged again?

rouault:
> can this pull request be re-opened and merged again?

I'd suggest rather to create a new branch from the latest state of branch-7-0, apply your changes and issue a new pull request

micking:
tx, I'll do

PR #5456, opened 2017-07-05
USE_POINT_Z_M default on

DDanielH:
Performance difference between on and off is minimal

### DIFF
--- a/nmake.opt
+++ b/nmake.opt
@@ -159,8 +159,8 @@ IGNORE_MISSING_DATA=-DIGNORE_MISSING_DATA
 # ----------------------------------------------------------------------
 # If you want to use shape Z and M parameter this option must be set.
 # It's OFF by default.
-#USE_POINT_Z_M=-DUSE_POINT_Z_M
-USE_POINT_Z_M=
+USE_POINT_Z_M=-DUSE_POINT_Z_M
+#USE_POINT_Z_M=
 
 
 ########################################################################
--- a/nmake.opt
+++ b/nmake.opt
@@ -158,7 +158,7 @@ IGNORE_MISSING_DATA=-DIGNORE_MISSING_DATA
 # Measures and Z values
 # ----------------------------------------------------------------------
 # If you want to use shape Z and M parameter this option must be set.
-# It's OFF by default.
+# It's ON by default.
 USE_POINT_Z_M=-DUSE_POINT_Z_M
 #USE_POINT_Z_M=
 


bjoernboldt:
If you change the code to "on" as default, you should change the comment too.
From `# It's OFF by default.` to `# It's ON by default.`

DDanielH:
Sorry, i forgot

jmckenna:
@DDanielH could you create this pull request to mapserver:master instead of branch-7-0 ?   (because this changes the functionality of MapServer, it will have to be applied to the 7.2 release)

geographika:
Can anyone provide any information on why in CMakeLists there is the following [comment](https://github.com/mapserver/mapserver/blob/d5c22cd3113ba85f5b1f1c406fe4055f75d04ce3/CMakeLists.txt#L149)?

`
option(WITH_POINT_Z_M "include Z and M coordinates in point structure (advanced, not recommended)" OFF)
`

Is this recommendation now out of date? What were the issues in using it?
The comment was added during the migration from autotools to cmake 6 years ago.

sdlime:
Assuming the cmake config was put together from the old configure build system it’s likely it’s out-of-date. There is at least one spot I know of that could leverage Z/M but doesn’t – that’s the templating code (shpxy tag). There’s a separate ticket for that to introduce an extended shpxy tag or more like add new ones. It’s possible other output drivers (e.g. GML, KML) aren’t complete in this area but if I remember correctly that’s why support was added in the first place, at least for GML. Another place that could be deficient is expression handling. We’d have to check…

I’m curious if the binary distros are built with this option or not.

-Steve

msmitherdc:
The output formats do support Z output but there are requirements on setting options on the output formats. See https://mapserver.org/output/ogr_output.html, specifically ows/wfs_geomtype.

 

Mike

geographika:
@sdlime - The Windows builds at GIS Internals don't have this option set - which then throws errors in various tests when using Python MapScript built with USE_POINT_Z_M ON (which is why it would be good to set this to ON by default if it is stable).

juliensam:
Note that this switch was added a long time ago because having a point structure of 2 int instead of 4 int meant it was better handled in memory and made a "significant" performance improvement. See details here: https://trac.osgeo.org/mapserver/ticket/1244

It was a time of custom build and it was assumed that people were compiling a lot. This is not the case anymore and it is really possible that the performance "improvement" is not even true anymore. We could set this switch to ON by default or even remove the switch altogether if that is the case.

geographika:
@juliensam - thanks for the info. It sounds like at least the CMake warning can be removed, and then possibly set it to ON by default?

This pull request however modifies nmake.opt - which as I understand it is redundant, and so a new pull request should be made and this one closed?

geographika:
I think this can be closed - nmake is no longer used since the switch to CMake. 
Perhaps the nmake.opt file should be removed completely to avoid confusion?

rouault:
> Perhaps the nmake.opt file should be removed completely to avoid confusion?

yes, feel free to submit a PR removing it. I'm supportive

PR #5459, opened 2017-07-07
Fix error to support Oracle compound elements for GTYPE 2007

micking:
Fix error to support Oracle compound elements for GTYPE 2007 (multipolygon) and not only for GTYPE 2003 (polygon)


msmitherdc:
this is good to merge.

jmckenna:
thanks @micking !

PR #5461, opened 2017-07-10
Add PHP7.x support

bjoernboldt:
I added some macros and made some changes to the structs in php_mapcript.h for PHP7. So the code is now compatible with PHP5.x and PHP7.x.

### DIFF
--- a/mapquery.c
+++ b/mapquery.c
@@ -736,7 +736,7 @@ int msQueryByFilter(mapObj *map)
     paging = msLayerGetPaging(lp);
     msLayerClose(lp); /* reset */
     status = msLayerOpen(lp);
-    if(status != MS_SUCCESS) goto query_error;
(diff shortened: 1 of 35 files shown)


sdlime:
So this gives us PHP7 support 2 ways - via php_mapscript and Swig? --Steve

nboisteault:
Thank you very much! Will it land in mapserver 7.0 soon?

sdlime:
I don't know quite yet. Assuming it works - I'm not a PHP user and so not in a position to test - I don't know why not. This probably needs to be taken to the mapserver-dev list for discussion. Might be good if you brought it up? There is also SWIG-based PHP 7.0 effort... --Steve

stoecker:
CMake says "-- Found PHP5-Version 7.2.5 (using /usr/bin/php-config)". Text should be adapted ;-)

geographika:
PHP-NG support added with #5675

PR #5467, opened 2017-07-27
compile the Ms sql plugin

lepirlouit:


jmckenna:
@lepirlouit I have never faced this issue (using MSVC 2012).  What compiler are you using?

geographika:
Is the windows.h file there to make it clear the plugin can only be compiled on Windows?

lepirlouit:
why can the plugin not be compiled on linux ?

geographika:
@lepirlouit - this has been fixed with #5996, and the MSSQL driver is now built on Travis/Linux as part of the MapServer CI tests. Thanks for raising this.

Issue #5477, opened 2017-09-18
Cannot load more than one .kml layer


erimirtis:
I created a map with my region, radar data e lightning data.
It all works well if I have only one kml file with lightning data;
if I try to split the data in different layers only one layer is shown.

1. All layers are working separately. 
2. They don't work together when I load them all in the main mapfile trought an INCLUDE
(as shown in my example files) 
3. or even if I directly write them all in the main mapfile

Thank you for your support.
Any suggestion is appreciated.

[lightning map.zip](https://github.com/mapserver/mapserver/files/1310032/lightning.map.zip)

geographika:
This should probably be sent to the user's mailing list - see http://mapserver.org/community/lists.html#mapserver-users

Your issue seems to be having an additional END which prematurely ends the Mapfile.

```
    CLASS
        STYLE
            SYMBOL "cross" # no need to have an END here as it is a string
            SIZE 8
            WIDTH 2
            COLOR 200 200 200
        END
    END
```

erimirtis:
Thank you for your help (and for pointing me to the mailing list, I didn't know about it!)

jratike80:
No confirmation from the reporter but the additional END is an obvious reason so no bug.

Issue #5475, opened 2017-09-01
Simple fix but I'm new: I’ve installed MapServer but can’t see the layers on the ‘demo’.


neilpritchard:
I’ve installed MapServer but can’t see the layers on the ‘demo’.
I followed the instructions on the MapServer website, to install MapServer on Ubuntu LTS 16.04.3 Server:-

sudo apt-get update
sudo apt-get install apache2
sudo apt-get install libpng-dev
sudo apt-get install libjpeg-dev
sudo apt-get install curl
sudo apt-get install libpcre3-dev
sudo apt-get install libpixman-1-dev
sudo apt-get install libgdal1-dev 
sudo apt-get install libgeos-dev
sudo apt-get install libsqlite3-dev
sudo apt-get install libtiff5-dev
sudo apt-get install software-properties-common python-software-properties
sudo apt-add-repository ppa:ubuntugis/ubuntugis-unstable
sudo apt-get install cgi-mapserver
sudo apt-get install mapcache-cgi 
sudo apt-get install libapache2-mod-mapcache
sudo apt-get install mapserver-bin
sudo apt-get install libapache2-mod-fastcgi
sudo service apache2 restart
sudo ldconfig

Created the following apache configuration to the file in /etc/apache2/sites-available/001-mapserver.conf :-

<VirtualHost *:80>
        ServerName 192.168.116.135

        <IfModule mapcache_module>
           <Directory /var/www/mapserver>
              Require all granted
           </Directory>
           MapCacheAlias /mapcache "/var/www/mapserver/mapcache.xml"
        </IfModule>

        ServerAdmin webmaster@localhost
        DocumentRoot /var/www/mapserver
		
        LogLevel warn
        ErrorLog ${APACHE_LOG_DIR}/mapserver-error.log
        CustomLog ${APACHE_LOG_DIR}/mapserver-access.log combined
</VirtualHost>

sudo ln -s /etc/apache2/sites-available/001-mapserver.conf /etc/apache2/sites-enabled/001-mapserver.conf

sudo mkdir /var/www/mapserver

sudo cp /usr/share/doc/libapache2-mod-mapcache/examples/mapcache.xml /var/www/mapserver/

sudo apachectl restart 

After installing I can see the following list when I call the URI http://192.168.116.135/mapcache/demo/ :-
tms
wmts
gmaps
kml
ve
mapguide
wms

[...]

Issue #5437, opened 2017-05-30
Reprojection error for CONNECTIONTYPE WMS


Simmesimme:
I am using mapserver to reproject raster data from various projections to the [HEALPix projection from proj.4](http://proj4.org/projections/healpix.html). This works well when the layer source is a raster, however I get severe reprojection problems when using an external WMS as source for my layer.

## Some examples

Here is the layer description for the raster source (I omitted all metadata for clarity, if this is important, I can provide it):

```
LAYER
    NAME "bluemarble"
    STATUS ON
    TYPE RASTER
    DATA "bluemarble.vrt"

    PROJECTION
        AUTO
    END
END
```
And here we have the WMS sourced layer:

```
LAYER
    NAME "dlrbasemap"
    TYPE RASTER
    STATUS ON

    CONNECTION "https://geoservice.dlr.de/eoc/basemap/wms?"
    CONNECTIONTYPE WMS

    METADATA
        WMS_NAME            "basemap"
        WMS_SERVER_VERSION  "1.1.1"
        WMS_FORMAT          "image/png"
        WMS_SRS             "EPSG:4326"
        WMS_TITLE           "dlrbasemap"
    END
END
```
In the epsg file I added this line:
```
<900916> +proj=healpix +a=1 <>
```

When I request the raster layer with `...&LAYERS=bluemarble&SRS=EPSG:900916&BBOX=-3.5,-1.6,3.5,1.6` I get the expected result:

![healpix_good_all](https://cloud.githubusercontent.com/assets/829942/26572376/f59df980-4519-11e7-92cb-28ab209b7ab8.png)

When I perform the very same request but with `...&LAYERS=dlrbasemap&...` **I get an empty image**. 
For smaller bounding boxes (`...&LAYERS=bluemarble&SRS=EPSG:900916&BBOX=-0.5,0.6,0.5,1`) I get again correct results for the raster layer but somehow distorted results for the WMS layer.

Raster layer:
![healpix_good](https://cloud.githubusercontent.com/assets/829942/26572731/633df278-451b-11e7-98b2-a55c8cc84a60.png)

WMS layer:
![healpix_bad](https://cloud.githubusercontent.com/assets/829942/26572742/6a43023e-451b-11e7-8b87-693358040bb8.png)

(for example observe how parts of Africa are missing)

## Some thoughts

- Maybe it's just a configuration problem? But that seems unlikely since the reprojection is working more or less, it looks more like a precision problem...
[...]

PR #5478, opened 2017-09-18
Create CONTRIBUTING.md

geographika:
To let users know about the mailing list, and bug submission policy. A link to the file should show up when creating an issue: https://help.github.com/articles/setting-guidelines-for-repository-contributors/

This may prevent some of the recent issues that would have been better added to the mailing lists. For example:

https://github.com/mapserver/mapserver/issues/5477
https://github.com/mapserver/mapserver/issues/5475
https://github.com/mapserver/mapserver/issues/5437

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,37 @@
+# How to contribute to MapServer
+
+Contributions to the MapServer project are most welcome, and can take many forms such as detailed bug reports, documentation, 
+tests, features, and patches. 
+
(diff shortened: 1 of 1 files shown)


sdlime:
Thanks!

geographika:
@sdlime - thanks for adding it! 
For info it looks like the new issue message will only be shown when CONTRIBUTING is in the active branch (currently 7-0 by default for MapServer, whereas CONTRIBUTING.md is in master).

PR #5460, opened 2017-07-09
Update test Mapfile encoding to utf-8

geographika:
These files were in iso-8859-1 (for the German accents in Straßenfläche). All Mapfiles should be in utf-8 according to the docs.

Sample Python code used to convert below for reference (all line breaks etc. preserved). 

```
def change_encoding():
    sourceEncoding = "iso-8859-1"
    targetEncoding = "utf-8"
    files = ["wms_inspire_scenario2_cap130_ger.xml","wms_inspire_scenario2_cap111_ger.xml","wms_inspire_scenario1_cap130_ger.xml"]

    for f in files:
        source = open(os.path.join(r"D:\GitHub\mapserver\msautotest\wxs\expected", f))
        target = open(os.path.join(r"D:\GitHub\mapserver\msautotest\wxs\expected", f + ".utf.xml"), "wb")
        target.write(unicode(source.read(), sourceEncoding).encode(targetEncoding))
```

PR #5479, opened 2017-09-19
Update test Mapfile encoding to utf-8

geographika:
This replaces #5460 as it was not created in a feature branch..apologies. 

These files were in iso-8859-1 (for the German accents in Straßenfläche). All Mapfiles should be in utf-8 according to the docs.

Sample Python code used to convert below for reference (all line breaks etc. preserved). 

```
def change_encoding():
    sourceEncoding = "iso-8859-1"
    targetEncoding = "utf-8"
    files = ["wms_inspire_scenario2_cap130_ger.xml","wms_inspire_scenario2_cap111_ger.xml","wms_inspire_scenario1_cap130_ger.xml"]

    for f in files:
        source = open(os.path.join(r"D:\GitHub\mapserver\msautotest\wxs\expected", f))
        target = open(os.path.join(r"D:\GitHub\mapserver\msautotest\wxs\expected", f + ".utf.xml"), "wb")
        target.write(unicode(source.read(), sourceEncoding).encode(targetEncoding))
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: precise
 language: php
 php:
   - 5.5
--- a/mapcopy.c
+++ b/mapcopy.c
@@ -500,6 +500,7 @@ int msCopyStyle(styleObj *dst, styleObj *src)
   MS_COPYSTELEM(offsetx);
   MS_COPYSTELEM(offsety);
   MS_COPYSTELEM(angle);
+  MS_COPYSTELEM(autoangle);
   MS_COPYSTELEM(minvalue);
   MS_COPYSTELEM(maxvalue);
   MS_COPYSTELEM(opacity);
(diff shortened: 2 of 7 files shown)


geographika:
Just to note the test failed as there were issues with Travis when it was run.

rouault:
@geographika Can you rebase on freshly updated branch-7-0 now that it passes test ?

rouault:
Ah, you merged instead of rebasing. The history is somewhat messed-up now. OK I'm going to pull your branch and fix that

rouault:
Fixed and committed per https://github.com/mapserver/mapserver/commit/a3a481a1d83e1f6749c77172c300902cd5778ed0

geographika:
@rouault - thanks! Sorry got in a real mess with that branch and was going to come back to it later..

PR #5482, opened 2017-09-26
MINSCALE and MAXSCALE are both deprecated

geographika:
Update to MAXSCALEDENOM in test case

### DIFF
--- a/msautotest/gdal/heat.map
+++ b/msautotest/gdal/heat.map
@@ -92,26 +92,26 @@ layer
     type POINT
     data "data/pnts.shp"
     CLASS
-     MAXSCALE 255000000
+     MAXSCALEDENOM 255000000
      STYLE
       SIZE [VAL]
      END
     END
     CLASS
-     MAXSCALE 265000000
+     MAXSCALEDENOM 265000000
      STYLE
       SIZE 0.1
      END
     END
     CLASS
-     MAXSCALE 275000000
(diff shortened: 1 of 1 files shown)


rouault:
I'm wondering if it doesn't make sense to still have them tested ? They are deprecated, but are still supposed to work, so better make sure they still work with the test (although I don't think anyone really madesure that deprecated keywords are still tested...)

geographika:
@rouault - there are plenty of deprecated keywords around still, the reason I added this particular case was because MINSCALE and MAXSCALE are no longer listed in the CLASS docs page - http://mapserver.org/mapfile/class.html

The pull request can be closed until further discussion on the dev lists if that is appropriate?

I'm not too far off allowing deprecated keywords to be highlighted in a mapfile with Python, so will post this to the dev list when it's ready.

rouault:
Actually I see there are a lot of remaining MAXSCALE uses: https://github.com/mapserver/mapserver/search?utf8=%E2%9C%93&q=%22MAXSCALE%22&type=

@sdlime Any opinion if we should remove all uses of deprecated keywords in msautotest, or keep some on purpose to check that we still work with those ? I raised this point, but I'm not at all strong about that. I don't maintain old mapfiles ;-)

jmckenna:
thanks @geographika Merging as you recommended, and further discussion can happen on -dev as you suggested.

PR #5488, opened 2017-10-01
DEBUG 10 is an invalid value

geographika:
Unless ``DEBUG 10`` is an undocumented setting?


jmckenna:
thanks @geographika good catch.

PR #5489, opened 2017-10-04
Add an offset to the scalebar

theduckylittle:
The scalebar is "stuck" to the edges of the map. This can make the
scalebar difficult to read when it is embedded in some mapping applications.

This PR adds the "OFFSET" key word to the scalebar object. The OFFSET X and Y
moves the scalebar away from the map a set number of pixels.

See: https://github.com/geomoose/gm3/issues/161 for an example of the rendering gone astray.


sdlime:
I thought there was an easier way to fine-tune placement already available (beyond POSITION) but I don't recall exactly what it was - I'll have to check before proceeding...

sdlime:
How does the offset behave when combined with position? I mean are the values absolute so the same values wouldn't work for LL vs UR?

theduckylittle:
It's an offset of the position as defined by position.  So if the POSITION is LL then the OFFSET is the number of pixels away from the LL corner.  If the POSITION is UR then it's the number of pixels is the distance from the upper-right corner.  For the positions which specify a center then it's the number of pixels to be offset from the calculated left-hand position of the placement.

sdlime:
Any idea why the CI test is failing?

theduckylittle:
I think it's an issue with different versions of cairo, all my local cairo tests are failing.

theduckylittle:
The test image I included was rendered locally with my version of Cairo, and therefore passes, but with the one in travis, it is failing.

theduckylittle:
I did find there were substantial Cairo version difference between what I'm running locally (Xenial, cairo @ 1.16.1) and what's running on Travis/Vagrant (Precise, cairo @ 1.10.2).  I've updated the image to one that I generated inside of the vagrant environment.

theduckylittle:
Travis now passes with the fixed PNG.

geographika:
Is there any chance of making this a simple attribute rather than a OFFSET...END block?
OFFSET is already an attribute of STYLE and LABEL and used as a simple:

`OFFSET 5 5`

rather than

`OFFSET 5 5 END`

In the interests of consistency and introducing new ambiguities into the Mapfile syntax it would be preferable to have the former - although a switch could introduce more confusion.

theduckylittle:
The actual code is a copy-and-paste of the label code. I believe END is actually just being ignored. 

From `loadLabel`:
```
      case(OFFSET):
        if(getInteger(&(label->offsetx)) == -1) return(-1);
        if(getInteger(&(label->offsety)) == -1) return(-1);
        break;
```

From `loadScalebar`:
```
      case(OFFSET):
        if(getInteger(&(scalebar->offsetx)) == -1) return(-1);
        if(getInteger(&(scalebar->offsety)) == -1) return(-1);
        break;
```

I believe in this case both syntaxes are valid.

geographika:
@theduckylittle - is it ok to change the test to match the non-END syntax (see pull request)? 
The docs actually don't mention END in the example, so this should be the only modification.

theduckylittle:
Doesn't bother me. :-)

Issue #5499, opened 2017-10-26
WFS Get Feature - No result with PropertyIsEqualTo


ger-benjamin:
Hi,

I'm on a custom package of MapServer master (7.x), near here: https://github.com/mapserver/mapserver/commits/1c132ced344ec4f2511c5f56773b9974216e7687

If I do a WFS Get Feature request with a "PropertyIsEqualTo" filter, I get no result at all. Never.

Example of request (on a text column "ele", with value 407, but that's the same result with a text value/column):

`````
<wfs:GetFeature xmlns:wfs="http://www.opengis.net/wfs" service="WFS" version="1.1.0" maxFeatures="200" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/wfs http://schemas.opengis.net/wfs/1.1.0/wfs.xsd"><wfs:Query typeName="feature:bus_stop" srsName="EPSG:21781" xmlns:feature="http://mapserver.gis.umn.edu/mapserver"><ogc:Filter xmlns:ogc="http://www.opengis.net/ogc"><ogc:PropertyIsEqualTo matchCase="false"><ogc:PropertyName>ele</ogc:PropertyName><ogc:Literal>407</ogc:Literal></ogc:PropertyIsEqualTo></ogc:Filter></wfs:Query></wfs:GetFeature>
````

PostGIS log: 

````
2017-10-26 09:56:34 CEST LOG:  statement: BEGIN
2017-10-26 09:56:34 CEST LOG:  statement: SELECT main.tsearch.id AS main_tsearch_id, main.tsearch.label AS main_tsearch_label, main.tsearch.layer_name AS main_tsearch_layer_name, main.tsearch.role_id AS main_tsearch_role_id, main.tsearch.public AS main_tsearch_public, main.tsearch.ts AS main_tsearch_ts, ST_AsBinary(main.tsearch.the_geom) AS main_tsearch_the_geom, main.tsearch.params AS main_tsearch_params 
	FROM main.tsearch 
	WHERE ts @@ to_tsquery('fr', 'pau:*') AND main.tsearch.public IS true ORDER BY ts_rank_cd(ts, to_tsquery('fr', 'pau:*'), 2|8) DESC, main.tsearch.label 
	 LIMIT 1
2017-10-26 09:56:34 CEST LOG:  statement: ROLLBACK
2017-10-26 09:56:35 CEST LOG:  statement: BEGIN
[...]

PR #5512, opened 2017-11-07
postgis - implements the "ilike" / IEQ operator

pmauduit:
This PR is a fix proposal for issue #5499 ; when a OGC filter is posted with a case-insensitive search, the postgis code is unable to translate the filter into a query, raising the following error:

```
msPostGISLayerTranslateFilter(): General error message. Translation to native SQL failed.
```

The fix was simply implemented by removing the `MS_TOKEN_COMPARISON_IEQ` from the "unsupported operator" list in mappostgis.c then adding a "ilike" expression instead of an empty string in the method responsible of transforming the token to SQL.

Tests: directly by faking onto our dev server the libmaserver.so with a custom `LD_LIBRARY_PATH`, ensuring the previous postgis function was actually able to generated a SQL query (which can be then retrieved in our postgresql logs with no error).

Concerns / critics: relying on the postgresql doc, it actually seems that the ILIKE operator is not in the SQL standard but a postgresql extension, and maplayer.c seems to be db-agnostic.

### DIFF
--- a/maplayer.c
+++ b/maplayer.c
@@ -623,7 +623,7 @@ const char *msExpressionTokenToString(int token) {
     case MS_TOKEN_COMPARISON_GE: return " >= ";
     case MS_TOKEN_COMPARISON_LT: return " < ";
     case MS_TOKEN_COMPARISON_LE: return " <= ";
-    case MS_TOKEN_COMPARISON_IEQ: return "";
+    case MS_TOKEN_COMPARISON_IEQ: return " ilike ";
     case MS_TOKEN_COMPARISON_RE: return " ~ ";
     case MS_TOKEN_COMPARISON_IRE: return " ~* ";
     case MS_TOKEN_COMPARISON_IN: return " in ";
--- a/mappostgis.c
+++ b/mappostgis.c
@@ -3990,7 +3990,6 @@ int msPostGISLayerTranslateFilter(layerObj *layer, expressionObj *filter, char *
           break;
 
 	/* unsupported tokens */ 
-	case MS_TOKEN_COMPARISON_IEQ:
         case MS_TOKEN_COMPARISON_BEYOND:
 	case MS_TOKEN_FUNCTION_TOSTRING:
 	case MS_TOKEN_FUNCTION_ROUND:


rouault:
ILIKE in PostgreSQL takes a pattern as its right operand, where _ and % have special meanings, so if the literal searched has those character, this will not behave properly.
Potential workaround: do proper escaping in mspostgis

But ILIKE is inefficient and will always result in sequential scan. An alternative mentionned in https://stackoverflow.com/questions/4482206/postgresql-case-insensitive-string-comparison?answertab=active#tab-top would be to use lower(column_name) = lower('literal') , since you can create an index on lower(column_name). 

Adding new test case in msautotest/wxs/wfs_filter_postgis.map would be appreciated

pmauduit:
Thanks @rouault ; I was aware of the possibility to user `lower(...) = lower(...)` as a clause, my understanding of the code is that it cannot be done in `msExpressionTokenToString()`, so I guess everything should be done into mappostgis.c. This PR will need a rework though.

rouault:
Yes that's a bit painful to do given the list structure instead of an abstract syntax tree which would be very helpful there. This is still doable by looking ahead for the next token. Typically you would do that in the MS_TOKEN_BINDING_STRING case.  (just for an exemple to give some idea, but you cannot immediately adapt this, you can look at https://github.com/mapserver/mapserver/blob/branch-7-0/mapogr.cpp#L1863 that consumes the next token at https://github.com/mapserver/mapserver/blob/branch-7-0/mapogr.cpp#L1883)

DennisRutjes:
+1

DennisRutjes:
bump

rahmabPublic:
what is the status of this issue?
I'm facing the same above described probblem.
I thought that the fix of @rouault  has been merged, but that is not the case.
Thanks any way

sdlime:
I don't believe the original pull request was reworked... @pmauduit or @rouault?

rouault:
> I don't believe the original pull request was reworked... @pmauduit or @rouault?

It wasn't. I might have the opportunity to work on that in the following weeks

rouault:
This has been fixed in 7.4.3 per https://github.com/mapserver/mapserver/pull/5950

PR #5513, opened 2017-11-09
WCS 2.0 RectifiedGrid: use 'lon lat' and 'x y' labels even with CRS with 'inverted axis' order

rouault:
My understanding of the thread at https://lists.osgeo.org/pipermail/gdal-dev/2017-November/047650.html is that the RectifiedGrid.axisLabels element should reflect the way the grid axis are ordered (and not the axis order of the CRS as mandated by EPSG). As our offsetVector s are ordered in the GIS friendly order, the labels should reflect that as well

Note that according to Peter Baumann per https://lists.osgeo.org/pipermail/gdal-dev/2017-November/047668.html the content could be pretty much anything.


rouault:
@ajolma @Schpidi opinions ?

ajolma:
I agree. The order is important but the names can be different from coverage axes' names.

Schpidi:
Reading over the mail thread your proposal here seems to be a good proper way forward. Many thanks again for taking care @rouault!

Issue #5508, opened 2017-11-01
Revive ANTIALIAS keyword in STYLE (for lines)


erik-h11:
The ANTIALIAS keyword seems to be silently ignored in the current version. We have a need for drawing aliased lines - with antialiased lines, the resulting colors (RGBA values) do not necessarily correspond to the values specified as COLOR, which prevents us from implementing an idea based on redrawing certain road segments with Javascript.

jratike80:
Please read this long mail thread from mapserver-users https://lists.osgeo.org/pipermail/mapserver-users/2017-October/080278.html . That is also the right place for starting a new discussion, not this issue tracker.

sdlime:
I don’t think we can do it at the STYLE level, more likely at the OUTPUTFORMAT level.
________________________________

erik-h11:
Actually I was the one starting that mapserver-users thread. I managed to achieve my goal my modifying (mostly) mapfile.c and mapagg.cpp; I created this ticket because I wanted to create a Pull Request (but perhaps issues and PRs are not necessarily related).

Having some problems with the PR because I branched off branch-7-0, and the instructions in the WorkingWithGit Wiki page mention 'master'; getting compile errors with 'master'.

 I think 'master' needs some updates; will try again later.

jratike80:
Sorry, I did not remember that. At least the mail thread is now recorded as background information. Perhaps the setting it could be implemented as a new FORMATOPTION. I was pondering about the existing GAMMA option, does it have anything common with this?

erik-h11:
My solution was to use the ANTIALIAS keyword: the parser in the current code still recognizes it, but the loadStyle() method simply ignores it. I'm now setting a flag in styleObj if the symbol is MS_FALSE, then copy that flag to strokeStyleObj, and in agg2RenderLine (mapagg.cpp) I pick renderer_scanline_bin_solid if the flag is MS_FALSE.

geographika:
Fixed with #6225

PR #5514, opened 2017-11-09
Issue #5508 - Revived ANTIALIAS keyword in STYLE section 

erik-h11:
Revived ANTIALIAS keyword in STYLE section (only for line rendering). 

The parser in the current code still recognizes this keyword, but the loadStyle() method simply ignores it. I'm now setting a flag in styleObj if the symbol is MS_FALSE, then copy that flag to strokeStyleObj, and in agg2RenderLine (mapagg.cpp) I pick renderer_scanline_bin_solid if the flag is MS_FALSE.

### DIFF
--- a/mapagg.cpp
+++ b/mapagg.cpp
@@ -67,11 +67,6 @@
 #include <pixman.h>
 #endif
 
-#ifdef AGG_ALIASED_ENABLED
-#include "renderers/agg/include/agg_renderer_primitives.h"
-#include "renderers/agg/include/agg_rasterizer_outline.h"
-#endif
-
 typedef mapserver::order_bgra band_order;
 
 #define AGG_LINESPACE 1.33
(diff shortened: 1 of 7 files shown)


mapagg.cpp
```diff
@@ -200,7 +202,8 @@ int agg2RenderLine(imageObj *img, shapeObj *p, strokeStyleObj *style)
 
   AGG2Renderer *r = AGG_RENDERER(img);
   line_adaptor lines = line_adaptor(p);
-
+  
+ 
 #ifdef AGG_ALIASED_ENABLED
```
tbonfort: please remove this unused ifdef'd block

mapagg.cpp
```diff
@@ -146,6 +147,7 @@ class AGG2Renderer
   renderer_base m_renderer_base;
   compop_renderer_base m_compop_renderer_base;
   renderer_scanline m_renderer_scanline;
+  renderer_scanline_aliased m_renderer_scanline_aliased;
 #ifdef AGG_ALIASED_ENABLED
```
tbonfort: please remove this unused block

erik-h11:
Yes, this change affects lines with width > 1.
I added two test files; not sure how to insert them into msautotest framework.

rouault:
Superesed per https://github.com/MapServer/MapServer/pull/6225

PR #5516, opened 2017-11-10
enable default value for variable substitution in map's WEB section

micking:
Added the option to set a default value for variable substitution in the map's WEB section. To do that there is a new method applyHashTableDefaultSubstitutions that is being called in msApplyDefaultSubstitutions with the map's WEB section as the hash table to work on

Prior to this a 'default_<variable name>' was ignored in the map's WEB section

### DIFF
--- a/mapfile.c
+++ b/mapfile.c
@@ -7140,6 +7140,25 @@ static void applyLayerDefaultSubstitutions(layerObj *layer, hashTableObj *table)
   return;
 }
 
+static void applyHashTableDefaultSubstitutions(hashTableObj *hashTab, hashTableObj *table)
+{
+	int i;
+	const char *default_key = msFirstKeyFromHashTable(table);
+	while (default_key) {
+		if (!strncmp(default_key, "default_", 8)) {
+			size_t buffer_size = (strlen(default_key) - 5);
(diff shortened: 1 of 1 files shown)


jmckenna:
@micking thanks for this.  Do you think a note about this change should be added to the docs?  If so, would you be willing to create a pull request for that?  (to https://github.com/mapserver/docs)

micking:
Hi Jeff,
Good suggestion, I'll create a note in the docs
Tx
Martin

Sent from my mobile device



Von: Jeff McKenna
Gesendet: Montag, 13. November, 22:07
Betreff: Re: [mapserver/mapserver] enable default value for variable substitution in map's WEB section (#5516)
An: mapserver/mapserver
Cc: Martin Icking, Mention


@micking<https://github.com/micking> thanks for this. Do you think a note about this change should be added to the docs? If so, would you be willing to create a pull request for that? (to https<https://github.com/mapserver/docs>://<https://github.com/mapserver/docs>github.com<https://github.com/mapserver/docs>/<https://github.com/mapserver/docs>mapserver<https://github.com/mapserver/docs>/<https://github.com/mapserver/docs>docs<https://github.com/mapserver/docs>)

PR #5520, opened 2017-11-14
Fix bug in msQueryByFilter

bjoernboldt:
Mapserver crashes when an error occurs in msQueryByFilter (e.g. value of filteritem is not a column name in a shape file).
We need to restore filter and filteritem at exit if we changed them and only then. In v6.4 mapserver crashes if the layer can not be opened (line 739).

### DIFF
--- a/mapquery.c
+++ b/mapquery.c
@@ -736,7 +736,7 @@ int msQueryByFilter(mapObj *map)
     paging = msLayerGetPaging(lp);
     msLayerClose(lp); /* reset */
     status = msLayerOpen(lp);
-    if(status != MS_SUCCESS) goto query_error;
+    if(status != MS_SUCCESS) return MS_FAILURE;
     msLayerEnablePaging(lp, paging);
 
     /* disable driver paging */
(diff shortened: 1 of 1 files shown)


rouault:
Wondering if https://github.com/bjoernboldt/mapserver/blob/58342ff2041f4d21f10fdb02736226ecd4ac9753/mapquery.c#L851 will not result in double-free

jmckenna:
Need more advice to proceed with this merge, as I am planning on tagging to 7.0.7 release momentarily...

rouault:
> Need more advice to proceed with this merge

I'd suggest not to merge it in the current state. Needs some refinement.

jmckenna:
ok thanks @bjoernboldt for this, and feedback from @rouault ..merging now..

jmckenna:
oh i misread!

jmckenna:
will hold off for now..

bjoernboldt:
@rouault is right, so what now? Who can fix this?

rouault:
> @rouault is right, so what now? Who can fix this?

You ;-) You can push an additional commit in your branch (or better git commit -a --amend your previous commit with that fix, and git push --force)

sdlime:
@bjoernboldt, @jmckenna, @rouault - is this ready? --Steve

jmckenna:
@bjoernboldt  can you re-do this pull request against master?

jmckenna:
@bjoernboldt can you rebase this against master (instead of mapserver:branch-7-0) ?

rouault:
Superseded per https://github.com/MapServer/MapServer/pull/6212

PR #5528, opened 2017-11-30
Fix \ handling in WMS propIsLike filter

gberaudo:
It looks like the code was missing escaping of `\`.
Here is an untested fix.

### DIFF
--- a/mapogcfilter.c
+++ b/mapogcfilter.c
@@ -3132,9 +3132,9 @@ char *FLTGetIsLikeComparisonSQLExpression(FilterEncodingNode *psFilterNode,
         pszEscapedStr[j++] = '\\';
       } else
         pszEscapedStr[j++] = c;
-    } else if  (c == pszSingle[0]) {
+    } else if (c == pszSingle[0]) {
       pszEscapedStr[j++] = '_';
-    } else if  (c == pszEscape[0]) {
+    } else if (c == pszEscape[0]) {
       pszEscapedStr[j++] = pszEscape[0];
       if (i+1<nLength) {
         char nextC = pszValue[i+1];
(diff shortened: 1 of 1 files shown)


mapogcfilter.c
```diff
@@ -3142,6 +3142,9 @@ char *FLTGetIsLikeComparisonSQLExpression(FilterEncodingNode *psFilterNode,
         if (nextC == '\'') {
           pszEscapedStr[j++] = '\'';
           pszEscapedStr[j++] = '\'';
+        } else if (c == '\\') {
```
rouault: Shouldn't that be nextC ?

rouault:
I would prefer that you test it...
And if possible add a test around https://github.com/mapserver/mapserver/blob/branch-7-0/msautotest/wxs/wfs_filter.map#L53

mapogcfilter.c
gberaudo: Yes, fixed.

gberaudo:
Thanks @rouault for the quick reply and the pointer to the test file.

I checked the `msautotest/wxs/data/popplace.dbf` db file for an entry containing the `\` character but could not find any. How am I supposed to add a test?

rouault:
> I checked the msautotest/wxs/data/popplace.dbf db file for an entry containing the \ character but could not find any. How am I supposed to add a test?

I somewhat suspected that. You could modify an existing value of popplace.dbf to add a \ character in it (and likely have to change a number of existing test results to accommodate for that change). Or you create a new mapfile with a new data file. CSV files can be convenient for that with a syntax like
```
somecolum,anothercolumn,WKT
somevalue,anothervalue,"POINT(2 49)"
```

gberaudo:
Thanks for the tips.
I actually was able to binarily edit the dbf file using `hexcurse` and it still looked valid. !!
However I have not yet built mapserver and it does not look straightforward: tried on a redhat machine, immediately got errors like
```
/usr/bin/cc  --coverage -DCHECK_FUNCTION_EXISTS=strlcpy    CMakeFiles/cmTryCompileExec3625615537.dir/CheckFunctionExists.c.o  -o cmTryCompileExec3625615537 -rdynamic -lm 
CMakeFiles/cmTryCompileExec3625615537.dir/CheckFunctionExists.c.o: dans la fonction « main »:
CheckFunctionExists.c:(.text+0x27): référence indéfinie vers « strlcpy »
collect2: erreur: ld a retourné 1 code d'état d'exécution
```

I doubt I will be able to invest more time on this at the moment.
Please consider this PR as a well documented issue.

PR #5530, opened 2017-12-04
WCS: support providing creation options that are layer specific

rouault:
If a Mapfile defines a GEOTIFF outputformat that is enabled in a WCS
layer, then it is possible to define a per-layer value for a creation option,
with a syntax like:

    "wcs_outputformat_GEOTIFF_creationoption_COMPRESS" "LZW"

to say that for 'GEOTIFF' output format, then COMPRESS=LZW should be passed to
the GDAL GTiff driver.

For GDAL 2.3.0 and the GRIB driver that supports band-specific creation options,
it is possible to define such band-specific creation options with for example
```
   "wcs_outputformat_GRIB_creationoption_BAND_1_IDS" "CENTER=54(Montreal) SUBCENTER=0 MASTER_TABLE=4 LOCAL_TABLE=0 SIGNF_REF_TIME=1(Start_of_Forecast) REF_TIME=2017-01-05T00:00:00Z PROD_STATUS=0(Operational) TYPE=2(Analysis_and_forecast)"
```
In case of band selection, the code will automatically renumber the band number.


rouault:
@tomkralidis  This will probably interest you. Relevant extract from tests: https://github.com/mapserver/mapserver/pull/5530/commits/6b5ed6e941078a85f743f975d0b90c794807aed0#diff-8039fc179b561eb19af1e40dfa5f1f87R103

jratike80:
Are those creation options documented somewhere? I know that for GeoTIFF the options which come with GeoTIFF Coverage Encoding Profile are documented in http://www.mapserver.org/ogc/wcs_server.html but I haven't seen "wcs_outputformat_GEOTIFF_creationoption_COMPRESS" "LZW" before.

rouault:
> Are those creation options documented somewhere?

Documentation is my next task. Those creation options are for the MapServer service administrator only (contrary to the GeoTIFF Coverage Encoding Profile). This is an extension of the OUTPUTFORMAT.FORMATOPTION, but at the layer level (so as to be able to provide appropriate values for the various metadata needed by the GRIB driver)

PR #5533, opened 2017-12-20
Enable WMS Layers group with GetMap request

yjacolin:


rouault:
Tests are failing: https://travis-ci.org/mapserver/mapserver/jobs/319179614 . Expected results should likely be updated, but I'm somewhat concerned about expected/ows_wms_rootlayer_name_empty_capabilities.xml and expected/ows_wms_rootlayer_name_capabilities.xml having now a <Name> element, which seems to go a bit against the purpose of  478e475af6d1921ae79335fb3a8fc621a88072ad (#5404). This doesn't really revert #5404 since the root layer name has no <name>, but this will make more layers queriable, what #5404 tried to limit. @tomkralidis opinion welcome

jratike80:
I am not sure if I read the output right, but isn't the layer group still unqueryable? Because layer-road2 is queryable then also layer-roadgroup should be queryable and look like
  <Layer queryable="1">
      <Name>roadgroup</Name>

By looking at the output from the failing test it seems that just <Name> has been added.

 wxs => ows_wms_rootlayer_name_capabilities.xml
--- ../expected/ows_wms_rootlayer_name_capabilities.xml	2017-12-20 13:50:19.049361998 +0000
+++ ows_wms_rootlayer_name_capabilities.xml	2017-12-20 13:54:35.021361998 +0000
@@ -74,6 +74,7 @@
     <BoundingBox CRS="EPSG:4326"
                 minx="-1" miny="-1" maxx="-1" maxy="-1" />
     <Layer>
+      <Name>roadgroup</Name>
       <Title>roadgroup</Title>
       <Layer queryable="1" opaque="0" cascaded="0">
         <Name>road2</Name>

yjacolin:
Does #5404 is about root element? if so this is still the case: https://github.com/yjacolin/mapserver/blob/508c4eb0289b27dbcb9fdcfbd3e681df22b93044/msautotest/wxs/expected/ows_wms_rootlayer_name_empty_capabilities.xml#L62-L77

jratike80:
Seems good to me, root layer has title but no name which means that it is just a category title. From WMS 1.1.1 standard:

> If, and only if, a layer has a <Name>, then it is a map layer that can be requested by using
> that Name in the LAYERS parameter of a GetMap request. If the layer has a Title but no
> Name, then that layer is only a category title for all the layers nested within. A Map
> Server that advertises a Layer containing a Name element shall be able to accept that
> Name as the value of LAYERS argument in a GetMap request and return the
> corresponding map. A Client shall not attempt to request a layer that has a Title but no
> Name.
> A server shall throw an exception (code="LayerNotDefined") if an invalid layer is
> requested.

yjacolin:
yes! Test passed!

jratike80:
Fine that the tests passed. Can you confirm that the group layer now has the queryable="1" attribute? Or are we perhaps speaking about different things, you about making the group requestable for GetMap and I about making it queryable for GetFeatureInfo?

yjacolin:
yes, this is two different things. We wanted a way to send a GetMap request for a group of layers with more than 1 level (more than 1 subdirectory).

I will see if we can do more but I don't think so.

jratike80:
The name of the ticket really made be think that "queryable" is dealing with GetFeatureInfo.

I have thought that there is an open ticket about group layers which appear always non-queryable in GetCapabilities but I could not find such. QGIS for example does not activate the query tool for any group layers now.

rouault:
> I have thought that there is an open ticket about group layers which appear always non-queryable in GetCapabilities but I could not find such

Must be https://github.com/mapserver/mapserver/pull/5220

PR #5534, opened 2017-12-20
Throw exception if wms layer is essential

umeier:
Normally, when a layer request to another WMS server fails, this layer is skipped and the map is rendered without this layer. This behaviour is absolutely OK in most cases, but sometimes it could be necessary that the complete map fails if the requested layer is essential for this map.
This patch looks whether "wms_essential" is set. If not, MapServer behaves like before, otherwise an exception is thrown.

### DIFF
--- a/mapwmslayer.c
+++ b/mapwmslayer.c
@@ -1255,7 +1255,7 @@ int msDrawWMSLayerLow(int nLayerId, httpRequestObj *pasReqInfo,
   if ( !MS_HTTP_SUCCESS( pasReqInfo[iReq].nStatus ) ) {
     /* ====================================================================
           Failed downloading layer... we log an error but we still return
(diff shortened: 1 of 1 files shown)


yjacolin:
did you plan to document this? :)

umeier:
Yes, I should document this. Therefore I need to clone the documentation project, right?

yjacolin:
yes, exaclty the same as mapserver code source :)

jmckenna:
I think this PR should be made against master (not branch-7-0)

jmckenna:
The default / working branch is now branch-7-2

jmckenna:
@umeier can you re-do this pull request against master?

jmckenna:
@umeier bump

jmckenna:
@umeier can you re-do this pull request against master?

jmckenna:
handled through #6061

PR #5535, opened 2017-12-20
log2vis check if multi-line

qwef:
It seems that log2vis is deprecated because it doesn't support multi-line paragraphs. See: https://lists.freedesktop.org/archives/fribidi/2008-January/000515.html
I added a if statement that would throw an error given a multi-line paragraph.


jmckenna:
@qwef can your change this PR to master instead?

jmckenna:
@qwef can you rebase this against master (instead of mapserver:branch-7-0) ?

rouault:
Superseded per https://github.com/MapServer/MapServer/pull/6246

PR #5544, opened 2018-02-08
Fix segfault with polyline labels for empty (multi)linestrings

olt:
longest_line_index can point outside pll->ll if shape only contains
lines with <2 points.

This can segfault here: https://github.com/mapserver/mapserver/blob/c6abf4224e6a1429869b9ee7c7a72dd791f74eb3/mapprimitive.c#L1821-L1822

I'm unable to create a minimal mapfile as WKT and POINT features do not support broken MultiLineStrings.
However, msPolylineComputeLineSegments does already expect lines with <2 points.
https://github.com/mapserver/mapserver/blob/c6abf4224e6a1429869b9ee7c7a72dd791f74eb3/mapprimitive.c#L1584

### DIFF
--- a/mapprimitive.c
+++ b/mapprimitive.c
@@ -1552,7 +1552,7 @@ void msPolylineComputeLineSegments(shapeObj *shape, struct polyline_lengths *pll
 
   pll->ll = msSmallMalloc(shape->numlines * sizeof(struct line_lengths));
   pll->total_length = 0;
-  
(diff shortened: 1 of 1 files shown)


rouault:
I guess this applies to 7.0 branch as well ?  If so, could you base your PR on that branch ? Thanks

olt:
Uh, that went wrong. I'm trying to push again.

PR #5547, opened 2018-02-14
Wms 1.3.0 client

ejn:
#3039 has been open for a good few years now, and there's starting to be some WMS services no longer supporting 1.1.*, so here's my stab at extending the WMS Client support to WMS 1.3.0 whilst making minimal changes.

Apart from some changed parameter names (SRS -> CRS), the main difference is the strict axis order in 1.3.0. This patch respects that and should automagically flip axes as appropriate for the EPSG code - if necessary the axis order can be fixed using metadata wms_strict_axis_order (1/true/0/false),

The autotests use the Mapserver demo WMS and (i) check that support for previous versions (1.1.1) is not broken and (ii) the support for 1.3.0 works as expected, including axis swapping.

Still TODO would be documentation changes. There are also possibly one or two edge cases where a slightly poor configuration would likely result in unexpected behaviour or bad requests, but I think that this patch should support almost all sensible use cases.

Although I have submitted this PR against master, I think this should be a candidate for 7.2 - if necessary I can rebase/repush.


sdlime:
@dmorissette, @jmckenna, @rouault - could use some review help with this pull request. Looks like a great addition and very complete. Thanks @ejn!

rouault:
+1 from my review. Well done @ejn

sdlime:
Maybe a few folks can review at the sprint and we can commit at the end of the event. --Steve

PR #5574, opened 2018-03-27
Restore CMakeList.txt to its state before #5376 merge, and remerge select protobuf bits

rouault:


sdlime:
Thanks Even!

itbeyond:
I understand the docs are still being updated for this feature however I am just wanting to check on the compilation requirements to build a version with mvt support. I notice in this commit that in the x64 cmake yml flags you have set -DWITH_PROTOBUFC=0 - Do I simply need to set this to 1 to test and use the mvt options or is there a different compilation flags that I need to set?, I am wanting to compile for Windows X64 and in the past have used the GISInternals development builds however if I am reading this right the mvt support will not be compiled by default?

PR #5587, opened 2018-04-16
WMS: tag a layer as queryable in GetCapabilities response if queryable or sublayers are queryable (fixes #5570)

### LINKED ISSUE #5570: WMS: a layer should be queryable if at least one sublayer is queryable
Currently MapServer only sets the queryable attribute on leaf layers. It would be desirable that it tags parents layers as queryable as soon as they have a child layer that is queryable

(for reference a related attempt was done in https://github.com/mapserver/mapserver/pull/5220, but not using the above proposed solution)

rouault:
Also fix GetFeatureInfo request on the root layer to work when some
sublayers are not queryable.


rouault:
@sgrellet If you want to try this
```
git clone https://github.com/rouault/mapserver
git checkout wms_fix_queryable_caps
```

PR #5597, opened 2018-05-26
Added error message when WMS filter parameter is empty

ravhed:
If the WMS filter parameter is empty (e.g. FILTER=) you will get a 500 error which doesn't make it clear what caused the request to fail.

This pull request will make sure a error message is returned. It also modifies two other error messages which seemed to be copied from WFS filtering.

### DIFF
--- a/mapwms.c
+++ b/mapwms.c
@@ -323,8 +323,13 @@ int msWMSApplyFilter(mapObj *map, int version, const char *filter,
   char **paszFilters = NULL;
   FilterEncodingNode *psNode = NULL;
 
-  if (!map || !filter || strlen(filter)==0)
+  if (!map)
     return MS_FAILURE;  
+
+  if (!filter || strlen(filter) == 0) {
+    msSetError(MS_WMSERR, "The provided filter can not be empty.", "msWMSApplyFilter" );
+    return msWMSException(map, version, "InvalidParameterValue", wms_exception_format);
+  }
 
   /* Count number of requested layers 
    * Only layers with STATUS ON were in the LAYERS request param.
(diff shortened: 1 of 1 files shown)


sdlime:
Out of curiosity, is an empty filter allowable as per the WMS specs? Also wondering if this should be a pull request against 7.0 instead... --Steve

ravhed:
The filter parameter is vendor specific for WMS and is not part of the OGC WMS standard (at least I have never found anything regarding filter for WMS).

I'm not sure what would be the ideal solution when you receive an empty filter. Either you could just continue with the request as normal (applying no filter) or do what is currently happening - return an error. I see pros and cons with both alternatives.

pelord:
+1

sdlime:
Should take this one to mapserver-dev. We have 2 options:

1. throw an exception
2. ignore the empty filter

I don't suppose the WMS specification provides any clarity does it?

--Steve

dmorissette:
I'd tend to go with silently ignoring an empty filter.

This would be consistent with the behavior when you have multiple layers and you are allowed to pass empty filters for some of the layers, e.g.
 ```
FILTER=()(<Filter><PropertyIsEqualTo><PropertyName>NAME</PropertyName><Literal>Digby</Literal></PropertyIsEqualTo></Filter>)
```

sdlime:
Is there an easy way to determine if a filter is indeed empty?

dmorissette:
Not sure if I understand the question. By "empty filter" I meant the "FILTER=" param was provided in the QUERY_STRING, but its value is empty.

This is already tested in the proposed patch using an ``if (!filter || strlen(filter) == 0) `` to produce an error ... that would have to be changed to silently ignore the FILTER param.

ravhed:
Closing this pull request to create a new one targeting branch-7-2-0

Issue #5602, opened 2018-06-02
mssql driver fails with large attribute filters


szekerest:
The query string can be max 10000 chars because of the fixed size array used.

sdlime:
@szekerest, can we close this?

szekerest:
Backported to branch-7-2 closing

PR #5603, opened 2018-06-02
Fix mssql filter string limitation (#5602)

szekerest:


sdlime:
Should this be back ported? Certainly could be part of 7.2 IMHO.

szekerest:
Yes,definitely. I’ll do that tomorrow

PR #5605, opened 2018-06-11
Apply map rotation to symbols, similarly to the rule of labels (fixes #5271)

### LINKED ISSUE #5271: Issue between map_angle query string and ANGLE in the symbole
Use mapserver version: 7.0.0

And we have the following results:
7c9644a0-6e59-11e4-9e83-bd1d5c108558

Class definition:

```
    CLASS
        NAME "Manchon"
        EXPRESSION "ma"
        STYLE
            SYMBOL "eau_manchon"
            SIZE   5
            ANGLE  [orientation]
        END
    END
```

See also: https://github.com/camptocamp/c2cgeoportal/issues/1136

Thanks in advance

rouault:
If the symbol angle is unspecified or 0, then the map rotation is ignored.
Otherwise, the map rotation is applied.


rouault:
Closing that one for https://github.com/mapserver/mapserver/pull/5606 that is against branch-7-2

PR #5606, opened 2018-06-11
Apply map rotation to symbols, similarly to the rule of labels (fixes #5271)

### LINKED ISSUE #5271: Issue between map_angle query string and ANGLE in the symbole

rouault:
If the symbol angle is unspecified or 0, then the map rotation is ignored.
Otherwise, the map rotation is applied.


rouault:
The appveyor build issue
```
  LINK : fatal error LNK1181: cannot open input file '..\sdk\release-1911-x64\lib\freetype2411.lib' [C:\projects\mapserver\build\mapserver.vcxproj]
```
 is unrelated and due to the upgrade of freetype to 2.9.1 in gisinternals ( https://github.com/gisinternals/buildsystem/commit/252a46ed44bad761e4a7db6a24bde0fa6d84c1da ) . Working on that separately (https://github.com/mapserver/mapserver/pull/5607)

PR #5608, opened 2018-06-14
Use strcpy rather than strdup

geographika:
I guess I must have missed this in the original pull request. This was a fix supplied by @tbonfort and relates to this - https://github.com/mapserver/mapserver/pull/5277

This pull request includes the code from https://github.com/tbonfort/mapserver/commit/bacf44f24a2dab8eff47d98788a302b7e0b0bb49

If strdup is used on Windows/MSVC then char variables don't get set correctly (at least when using the debugger). 

I'm not sure if this should committed to master or the soon to be released 7.2 branch (which would be ideal).

### DIFF
--- a/mapstring.c
+++ b/mapstring.c
@@ -2110,15 +2110,17 @@ char *msStrdup( const char * pszString )
   if( pszString == NULL )
     pszString = "";
 
-  pszReturn = strdup( pszString );
+  pszReturn = msSmallMalloc(strlen(pszString)+1);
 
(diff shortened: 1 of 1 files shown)


geographika:
Looking at the release steps at https://github.com/mapserver/mapserver/wiki/MapServer-7.2-Release-Plan then I guess this should be added to the 7.2 branch if it can be included in the next release. 

Replaced with #5610

PR #5610, opened 2018-06-14
Use strcpy rather than strdup

geographika:
I guess I must have missed this in the original pull request. This was a fix supplied by @tbonfort and relates to this - https://github.com/mapserver/mapserver/pull/5277

This pull request includes the code from https://github.com/tbonfort/mapserver/commit/bacf44f24a2dab8eff47d98788a302b7e0b0bb49

If strdup is used on Windows/MSVC then char variables don't get set correctly (at least when using the debugger).

### DIFF
--- a/mapstring.c
+++ b/mapstring.c
@@ -2103,22 +2103,24 @@ int msStringIsInteger(const char *string)
 
 /* Safe version of msStrdup(). This function is taken from gdal/cpl. */
 
-char *msStrdup( const char * pszString )
+char *msStrdup(const char * pszString)
 {
-  char        *pszReturn;
+    char        *pszReturn;
 
-  if( pszString == NULL )
-    pszString = "";
+    if (pszString == NULL)
+        pszString = "";
 
-  pszReturn = strdup( pszString );
+    pszReturn = msSmallMalloc(strlen(pszString) + 1);
 
-  if( pszReturn == NULL ) {
-    fprintf(stderr, "msSmallMsStrdup(): Out of memory allocating %ld bytes.\n",
-            (long) strlen(pszString) );
-    exit(1);
-  }
+    if (pszReturn == NULL) {
+        /*should never be here*/
+        fprintf(stderr, "msSmallMalloc(): Out of memory allocating %ld bytes.\n",
+            (long)strlen(pszString));
+        exit(1);
+    }
+    strcpy(pszReturn, pszString);
 
-  return( pszReturn );
+    return pszReturn;
 }
 
 
--- a/mapstring.c
+++ b/mapstring.c
@@ -2105,11 +2105,14 @@ int msStringIsInteger(const char *string)
 
 char *msStrdup(const char * pszString)
 {
-    size_t nStringLength = strlen(pszString) + 1; /* null terminated byte*/
-    char *pszReturn = malloc(nStringLength);
+    size_t nStringLength; 
+    char *pszReturn;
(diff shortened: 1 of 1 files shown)


mapstring.c
```diff
@@ -2103,22 +2103,24 @@ int msStringIsInteger(const char *string)
...
-  char        *pszReturn;
+    char        *pszReturn;
 
-  if( pszString == NULL )
-    pszString = "";
+    if (pszString == NULL)
+        pszString = "";
 
-  pszReturn = strdup( pszString );
+    pszReturn = msSmallMalloc(strlen(pszString) + 1);
```
rouault: a more efficient implementation would store strlen(pszString) in a nStringLength variable for reuse. And instead of strcpy(), you would do memcpy(pszReturn, pszString, nStringLength + 1 /* nul terminated byte*/)

geographika:
@rouault - thanks for the review. ``msStrdup`` is used throughout the codebase so I don't want to add any overhead. Could any string lengths be greater than an unsigned int used for ``nStringLength``?

mapstring.c
```diff
@@ -2103,22 +2103,23 @@ int msStringIsInteger(const char *string)
...
 {
-  char        *pszReturn;
 
-  if( pszString == NULL )
-    pszString = "";
+    if (pszString == NULL)
+        pszString = "";
 
-  pszReturn = strdup( pszString );
+    unsigned int nStringLength = strlen(pszString);
```
rouault: use size_t for type safety instead of unsigned int

mapstring.c
```diff
@@ -2103,22 +2103,23 @@ int msStringIsInteger(const char *string)
...
+    char *pszReturn = malloc(nStringLength + 1);
+    memcpy(pszReturn, pszString, nStringLength + 1); /* null terminated byte*/
```
rouault: memcpy() should be done after the below NULL check

mapstring.c
rouault: Being in a .c file, the variable declarations should be done at the beginning of the function to please MSVC
```
size_t nStringLength;
char* pszReturn;
```

mapstring.c
```diff
@@ -2103,22 +2103,23 @@ int msStringIsInteger(const char *string)
 
 /* Safe version of msStrdup(). This function is taken from gdal/cpl. */
 
-char *msStrdup( const char * pszString )
+char *msStrdup(const char * pszString)
 {
-  char        *pszReturn;
+    size_t nStringLength = strlen(pszString) + 1; /* null terminated byte*/
```
rouault: the declaration and its affectation of both variables should be separated. Namely after the if (pszString == NULL) pszString = ""; check
You're close ;-)

mapstring.c
```diff
@@ -2103,22 +2103,23 @@ int msStringIsInteger(const char *string)
...
 
-  if( pszString == NULL )
-    pszString = "";
+    if (pszString == NULL)
+        pszString = "";
 
-  pszReturn = strdup( pszString );
+    if (pszReturn == NULL) {
+        fprintf(stderr, "msSmallMalloc(): Out of memory allocating %ld bytes.\n",
+            (long)strlen(pszString));
```
rouault: strlen(pszString) could be replaced by nStringLength

geographika:
@rouault - thanks for your patience! Variables now set after the NULL check.

PR #5611, opened 2018-06-14
Branch 7.2 php mapscript ng

AlexanderGabriel:
Hi,

so this is my current state of work on PHP mapscript extension using SWIG.
I renamed the extension to mapscriptng because of windows. phpng_mapscript would cause the extension files to be named php_phpng_mapscript.
There is still work:
- There is a warning when generating the wrapper for a method called "empty". It is renamed to c_empty. You'll see this when compiling. How do "we" handle this?
- On Windows, there is a Problem with swig. It will not be possible to compile thread-safe extensions with swig 3.0.12 but if you build current swig from git build you can generate code that compiles. NTS-Extension work.
- Build Instruction need to be updated. Especially for Windows. I have some batch files here that will do everything. I can update the windows readmes.

Linux users should have no problems building it. I tested on Debian Testing a few weeks ago. Functionality should be there but i did really only do some basic tests...

Have Fun :)

Kind Regards,
Alex


AlexanderGabriel:
Hi,

please don't merge this. I will put everything into one commit and open another pull-request. Will look more professional ;)

AlexanderGabriel:
My new Request: https://github.com/mapserver/mapserver/pull/5616

PR #5615, opened 2018-06-18
.travis.yml and Vagrant: update to Trusty

rouault:
Due to UbuntuGIS PPA having removed support for Ubuntu Precise, we need to upgrade to Trusty. This causes a lot of expected results to be refreshed due to changes in library versions (GDAL in particular). There are a few strange changes though that could/should be later investigated, like msautotest/wxs/expected/wfs_ogr_native_sql_37.xml , msautotest/wxs/expected/wfs_ogr_native_sql_37.xml msautotest/wxs/expected/wfs_ogr_native_sql_40.xml , and msautotest/wxs/wfs_query_cache.map wfs_query_cache_2.xml that I had to disable due to inconsistant result between Vagrant and Travis


sdlime:
+1 - we can look into those tests later as you suggest.

PR #5616, opened 2018-06-18
Add Support for PHP-NG MapScript (via SWIG)

AlexanderGabriel:
Hi again,

this will add support for PHP-MapScript via SWIG.
There is still work:
- There is a warning when generating the wrapper for a method called "empty". It is renamed to c_empty. You'll see this when compiling. How do "we" handle this?
- On Windows, there is a Problem with swig. It will not be possible to compile thread-safe extensions with swig 3.0.12 but if you build current swig from git build you can generate code that compiles. NTS-Extension work.
- Build Instruction need to be updated. Especially for Windows. I have some batch files here that will do everything. I can update the windows readmes.

Linux users should have no problems building it. I tested on Debian Testing a few weeks ago. Functionality should be there but i did really only do some basic tests...

Kind Regards,
Alex


sdlime:
Any idea what's up with the ci fails? You may need to re-sync with the 7.2 branch (there were some changes to the travis setup).

AlexanderGabriel:
Not sure about the appveyor error. Building the extension requires php-sdk to be "installed" and "startet" because it proviedes an environment variable PHP_SDK_VC that helps to find the correct php sources.
Can you define env-vars for the builds? If not i can try to add some more ifs to determine used Visual Studio Verasion and find the path.
The travis-build i think failes because the versions of postgis changed.
In debian testing they are:
postgresql-10-postgis-2.4 - Geographic objects support for PostgreSQL 10
postgresql-10-postgis-2.4-scripts - Geographic objects support for PostgreSQL 10 -- SQL scripts

But before we do anything that is eventually already fixed: Is there a smart way to sync or do i have to make a new request?
Branch 7.2 was merged to 7.0 today. Should i use 7.0 or 7.2 as base-branch?

AlexanderGabriel:
Next try here: https://github.com/mapserver/mapserver/pull/5619

Issue #5582, opened 2018-04-14
Python SWIG bindings and problems updating hashTableObj


geographika:
As noted in the discussions at https://github.com/mapserver/mapserver/pull/5572 - in the Python SWIG bindings, the C hashTableObj is converted to a Python dictionary - (see [pymodule.i).](https://github.com/mapserver/mapserver/blob/master/mapscript/python/pymodule.i#L100) 
This allows for easy access to values using keys, for example at [test_mapio.py](https://github.com/mapserver/mapserver/blob/master/msautotest/mspython/test_mapio.py)

```python
headers = mapscript.msIO_getAndStripStdoutBufferMimeHeaders()
ct = headers['Content-Type']
cc = headers['Cache-Control']

```
It is now much easier to read any METADATA values, for example in a LAYER:

```python
l = mapscript.fromstring("LAYER TYPE POINT METADATA KEY1 VAL1 END END")
print(l.metadata) # {'KEY1': 'VAL1'}
```

However setting values is now harder. This dictionary is "disconnected" from its associated object - any edits to it won't be set on the original hashTableObj. 
The following attempts to set a METADATA value will fail:

```python
l = mapscript.fromstring("LAYER TYPE POINT METADATA KEY1 VAL1 END END")
l.metadata["KEY2"] = "VAL2" #  this can be set but will have no effect on the layerObj's metadata
l.metadata = {"k1": "v1", "k2": "v2"} # AttributeError: can't set attribute
```

The "new" get/set API as described at http://mapserver.org/mapscript/mapfile.html#metadata won't work, as these objects are now Python dicts:

```python
l.metadata.get('KEY1') # AttributeError: 'dict' object has no attribute 'set'
l.metadata.set("KEY1", "VAL2") # AttributeError: 'dict' object has no attribute 'set'
```
 
The old deprecated methods however still work (http://mapserver.org/mapscript/mapfile.html#backwards-compatibility-for-metadata)

```python
print(l.setMetaData ("KEY3", "VAL3"))
print(l.getMetaData ("KEY3")) # VAL3
```

There appear to be 3 solutions:

1. Update the docs to describe the above and leave the code as-is. 
2. Roll back the typemap conversion of hashTableObj to dict, and revert to the get/set API. 
[...]

PR #5620, opened 2018-06-20
Add Python methods to hashtable so it can be used like a dict

geographika:
Supersedes #5614

Add various Python methods so that a hashtable can be used like a dict. See #5582 for background. 
API is now as follows:

```python
import mapscript
ht = mapscript.hashTableObj()

ht["key1"] = "value1"
ht["key2"] = "value2"
ht["key3"] = "value3"
ht["key4"] = "value4"

del ht["key1"]
print(ht.keys)
print(len(ht))
print("k1" in ht)
print(len(ht))
```

Or for a more concrete example:

```python

print(map.web.metadata.keys()) # ['wms_srs', 'ows_enable_request', 'ows_http_max_age']
print(map.web.metadata["wms_srs"]) # EPSG:4326
map.web.metadata["wms_srs"] = "EPSG:2157"
print(map.web.metadata["wms_srs"])  # EPSG:2157
```

### DIFF
--- a/mapscript/python/pymodule.i
+++ b/mapscript/python/pymodule.i
@@ -105,11 +105,11 @@ CreateTupleFromDoubleArray( double *first, unsigned int size ) {
     def __setitem__(self, key, value):
         return self.set(key, value)
 
-    def __delitem__(self, k) :
+    def __delitem__(self, key) :
         return self.remove(key)
 
-    def __contains__(self, k):
-        return k.lower() in [key.lower() for k in self.keys()]
+    def __contains__(self, key):
+        return key.lower() in [k.lower() for k in self.keys()]
 
     def __len__(self):
         return self.numitems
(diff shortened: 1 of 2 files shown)


geographika:
Is this likely to make the 7.2 cut? If not I can recreate the pull request against master. 

The functionality is tested by both mapscript/python/tests/cases/hashtest.py and msautotest/mspython/test_mapio.py

Currently the hashtable implementation in Python Mapscript doesn't match the docs, or allow setting of values.

sdlime:
Yes, I think this can be merged. Can you do it?

geographika:
@sdlime - I don't have commit rights so can't help there sorry!

@rouault wrote the original SWIG typemap so could have reservations, although the new approach has the same API  as the test_mapio.py that was written to test they typemap so should be ok.

sdlime:
Probably should wait on a blessing from @rouault before pulling the trigger then.

rouault:
@geographika @sdlime Sorry, I missed your previous pings. Too many notifications from too many repositories.. Squashed and merged

geographika:
@rouault - thanks for merging!

PR #5630, opened 2018-07-21
Use RST for ReadMe

geographika:
In preparation for the 7.2 release, would it be worthwhile changing the README to an RST file so it is nicely formatted on GitHub? You can see a preview at https://github.com/geographika/mapserver/tree/7-2-readme
As the default branch in GitHub will be 7.2 until the next major release I created the pull request against the 7.2 branch. I can make it against master if preferable. 

I added in the status button of the master build from Travis. 
I can also add in Appveyor, but it seems like you need to be an admin to see the URL for the badge. @sdlime - if you get a chance maybe you could add me as an admin to the Appveyor MapServer project, similar to MapCache?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,7 +1,11 @@
+MapServer
+=========
+
+|Build Status|
+
 -------
 Summary
 -------
-
    
 MapServer is a system for developing web-based GIS applications. 
 The basic system consists of a CGI program that can be configured to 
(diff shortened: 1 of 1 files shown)


sdlime:
Crap, I just cut the release file before I saw this. We should be able to add it to whatever our default branch is though (currently 7.0). I'll merge once Jeff gets the website updated.

sdlime:
@geographika - you should have admin rights to both the MapServer and MapCache projects.

PR #5633, opened 2018-07-24
Remove -beta2 version suffix.

sebastic:
The `-beta2` suffix is still present in the MapServer 7.2.0 release.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,7 @@ include(CheckCSourceCompiles)
 set (MapServer_VERSION_MAJOR 7)
 set (MapServer_VERSION_MINOR 2)
 set (MapServer_VERSION_REVISION 0)
-set (MapServer_VERSION_SUFFIX "-beta2")
+set (MapServer_VERSION_SUFFIX "")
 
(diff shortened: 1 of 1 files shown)


sdlime:
Crap, I edited that file by hand, how didn’t it stick. Thanks for catching. Will have to recreate the tarball.

sebastic:
Ideally the 7.2.0 tarball is not modified, it has already been used by users who monitor the download directory closely.

The Debian packaging tools don't accept a tarball with the same name but different content for example.

For this reason an RC1 release would have been good which would already have had all changes for the final release, whose tarballs would only need to be renamed for the final release. See the GDAL release process for example.

If you do choose to recreate the 7.2.0 tarballs, and they will have more changes than this commit, I'll need to know about to be able to incorporate those changes in the Debian package, as we already have 7.2.0 release packaged without those changes.

sdlime:
The 7.2.0 tarball does have more changes than this commit. I *had* made the appropriate changes to HISTORY.TXT and the CMakeLists.txt but neglected to push them *argh*. So HISTORY.TXT wasn't reflecting the 7.2.0 release either so its worth fixing. The README.rst was also updated. Sorry for the hassle. Point well taken on an RC... --Steve

Issue #5613, opened 2018-06-15
Improve label WRAP


olt:
We want to improve label wrapping in MapServer so that road labels can be wrapped at space and dashes.

It goes into the same direction as #3266 and http://osgeo-org.1560.x6.nabble.com/Enhancement-to-RFC-40-conditional-wrapping-td4260294.html

MapServer only allows a single WRAP character. A possible solution would be to wrap at Unicode ZERO WIDTH SPACE (0x200b) and let the user insert this everywhere a wrap is possible. ZERO WIDTH SPACE is not rendered by Mapserver. This character can be inserted with PostgreSQL `replace(replace(name, ' ', E' \u200b'), '-', E'-\u200b')` for example.

However, WRAP is a char and does not support Unicode. Instead of changing this type (and opening a whole can of new issues), I suggest that we allow to insert an ASCII control character as a marker for ZERO WIDTH SPACE. RECORD SEPARATOR or UNIT SEPARATOR for example.

It's difficult to insert these chars for most endusers, maybe we can add a keyword for this and replace it internally with the ASCII control char? I don't know how the Mapfile parsing works and how this would be implemented.

Another idea would be to accept that wrapping at space is a common use case and change the default behavior to automatically wrap at SPACE and ZERO WIDTH SPACE. I would prefer this. ZERO WIDTH SPACE should be so uncommon that backwards compatibility should not be an issue. 

The msLayoutTextSymbol function needs to be changed to wrap at the provided WRAP char or at ZERO WIDTH SPACE if that char is RECORD SEPARATOR. This change is required anyhow (see #5612).

I have a working prototype and I can provide a PR.
Any opinions, suggestions, +1s?

sdlime:
I'm all for something that improves labeling and is easier for users. I'm having a hard time visualizing how these options would manifest themselves in a mapfile. What does your prototype do?

olt:
There is an implementation in #5634, together with a test case and two test images.

PR #5634, opened 2018-07-31
Fix label wrap (#5612) and wrap at ZERO WIDTH SPACE (#5613)

### LINKED ISSUE #5612: msLayoutTextSymbol replaces WRAP char unconditionaly with SPACE
The "new" msLayoutTextSymbol function replaces *all* occurences of the `WRAP` character with a SPACE character and then wraps labels at these characters if they are longer then `MAXLENGTH`. For example, if you use `WRAP '-'` the label `My-label-with-dash` is rendered as `My label with dash` even if the label is not wrapped.

This appears to be a regression with MapServer 7. (/cc @tbonfort)

Is there a reason for this behavior? My understanding is that we can remove:

https://github.com/mapserver/mapserver/blob/2b9d02203e38ecaad1214ef463ea501749824432/textlayout.c#L534-L540

and replace the checks for `' '` with `ts->label->wrap` at these positions:

https://github.com/mapserver/mapserver/blob/2b9d02203e38ecaad1214ef463ea501749824432/textlayout.c#L543
https://github.com/mapserver/mapserver/blob/2b9d02203e38ecaad1214ef463ea501749824432/textlayout.c#L549

I can provide a patch, but please consider my follow up issue.

olt:
This fixes #5612 and implements #5613.

Here are two images from the added test case.

This is the output with current master. Lines starting with W should wrap, lines with NW should not wrap. Note how the dash is removed in the first lines and how the second line is wrapped even if there is no wrap character. (#5612)

labels-wrap-master

This is the output with my fix. The first two lines are wrapped correctly. The lower lines show how it is possible to wrap at arbitrary positions by inserting ZERO WIDTH SPACE characters.
labels-wrap-fixed

Questions:
- I've added a test. How do I generate the expected image? Most "result images perceptually match, though files differ" on my system.
- The patch is against master but I can rebase it. Where? branch-7-2? branch-7-3?

I'll update to the documentation.


textlayout.c
```diff
@@ -531,27 +530,17 @@ int msLayoutTextSymbol(mapObj *map, textSymbolObj *ts, textPathObj *tgret) {
       }
     } else {
       assert(ts->label->maxlength > 0);
-      if(ts->label->wrap != ' '){
-        for(i=0;i<num_glyphs;i++) {
-          /* replace all occurences of the wrap character with a space */
-          if(glyphs.unicodes[i]== ts->label->wrap)
-            glyphs.unicodes[i]= ' ';
```
olt: This is where wrap characters were replaced unconditionally with SPACE.

textlayout.c
```diff
@@ -531,27 +530,17 @@ int msLayoutTextSymbol(mapObj *map, textSymbolObj *ts, textPathObj *tgret) {
...
-        }
-      }
       if(num_glyphs > ts->label->maxlength) {
+        int num_cur_glyph_on_line = 0; /*count for the number of glyphs on the current line*/
         for(i=0; i<num_glyphs; i++) {
-          if(glyphs.unicodes[i]== ' ')
-            break;
-        }
-        if(i != num_glyphs) { /* we have at least one space on which we can break */
```
olt: Loop to check if we have a wrap character is not necessary, as the actual wrapping loop does not do any computation (except counting num_cur_glyph_on_line).

sdlime:
Should we take this one to mapserver-dev?

olt:
I wrote on mapserver-dev already, but I didn't ask for a vote: https://lists.osgeo.org/pipermail/mapserver-dev/2018-August/015464.html

sdlime:
I saw that after my comment here. Let me reply to your note to try and solicit a little more review. If nothing then we’ll just move forward with the commit.

sdlime:
@olt - to confirm. No breaking changes on this one, correct? Seems appropriate for 7.0 or at worst 7.2 don't you think? --Steve

olt:
No, unless fixing a regression is a breaking change ;-)
I'm not that familiar with the release schedule and versioning, to know if 7.0 or 7.2 is appropriate. Please let me know on which branch I should rebase this.

sdlime:
Let’s go with 7.2.

olt:
Added the missing *expected* image and rebased on branch-7-2. Update to docs in https://github.com/mapserver/docs/pull/251

sdlime:
Thanks Oliver!

Issue #5636, opened 2018-08-01
Regular expressions does not work


bjoernboldt:
I use a regular expression in my mapfile to find all items with 'abc' in the middle
i.e.
CLASSITEM "FLD01"
CLASS
        EXPRESSION /.\*abc.\*/
        ..
END
It works fine in 7.0.7 and before, but got an error (loadClass(): Unknown identifier. Parsing error near (*): line xxx) in 7.2.0 (and 7.3.0).
When I am trying /.+abc.+/ it works in 7.2.0 too. Any ideas?

geographika:
As a guess it could be related to the multi-line comments implementation https://github.com/mapserver/mapserver/pull/5362

sdlime:
Good guess... I’ll look into it.
________________________________

sdlime:
In the meantime I believe a workaround would be something like:

  EXPRESSION ("[FLD01]" ~".*abc.*")

--Steve

bjoernboldt:
Thanks. This helps a lot.

sdlime:
While there is a workaround, if this is really related to the multi-line comment addition then there really is a bug since /.*abc.*/ is not a properly formed comment and shouldn't be causing a problem. I suspect there is an error with the regex associated with that functionality... --Steve

PR #5641, opened 2018-08-14
Made the lexer regex for regular expression EXPRESSIONs more specific…

sdlime:
… (see #5636).


sdlime:
The regex was to general and didn't allow for *'s anywhere in an expression. Now the * is not allowed as the first character after the leading / to avoid interpretation as a C-style comment. --Steve

PR #5583, opened 2018-04-14
Update C# bindings to work with SWIG 3.x

geographika:
See https://github.com/mapserver/mapserver/issues/5575

This initial pull request brings the number of errors when building with SWIG 3.0.12 from `1592 Error(s)` to`8 Error(s)`.

The remaining errors are:

```
[00:02:38]   mapscript.cs(25,58): error CS0234: The type or namespace name 'EntryPoint' does not exist in the namespace 'System.Runtime.InteropServices' (are you missing an assembly reference?) [C:\projects\mapserver\build\mapscript\csharp\mapscript.vcxproj]
[00:02:38]   mapscript.cs(25,4): error CS1729: 'DllImportAttribute' does not contain a constructor that takes 2 arguments [C:\projects\mapserver\build\mapscript\csharp\mapscript.vcxproj]
[00:02:38]   mapscriptPINVOKE.cs(191,10): error CS0111: Type 'mapscriptPINVOKE' already defines a member called '.cctor' with the same parameter types [C:\projects\mapserver\build\mapscript\csharp\mapscript.vcxproj]
[00:02:38]   mapscriptPINVOKE.cs(313,3): error CS0246: The type or namespace name 'ThreadStaticAttribute' could not be found (are you missing a using directive or an assembly reference?) [C:\projects\mapserver\build\mapscript\csharp\mapscript.vcxproj]
[00:02:38]   mapscriptPINVOKE.cs(313,3): error CS0246: The type or namespace name 'ThreadStatic' could not be found (are you missing a using directive or an assembly reference?) [C:\projects\mapserver\build\mapscript\csharp\mapscript.vcxproj]
[00:02:38]   mapscriptPINVOKE.cs(298,58): error CS0234: The type or namespace name 'EntryPoint' does not exist in the namespace 'System.Runtime.InteropServices' (are you missing an assembly reference?) [C:\projects\mapserver\build\mapscript\csharp\mapscript.vcxproj]
[00:02:38]   mapscriptPINVOKE.cs(298,4): error CS1729: 'DllImportAttribute' does not contain a constructor that takes 2 arguments [C:\projects\mapserver\build\mapscript\csharp\mapscript.vcxproj]
```

Issue #5557, opened 2018-03-04
PHP7-Support via SWIG


AlexanderGabriel:
Hi together,

i did some work to make PHP-MapScript compile with SWIG on Linux. I think, this is the way, we should implement this to have less work in future.
It is a really early state an very poor testet and theres still a little bit of work to do (for example check exact swig version on 3.0.11) I testet in this environment:
- Debian testing (from last week) x64

It completely builds with CMake. To enable PHP7-Support add -DWITH_PHP7=1.
I thought this is a good option to be safe in Future when new PHP-Versions arrive and SWIG does not support them. Maybe i can merge this together into one CMake-Module to autodetect PHP-Version and SWIG but this is Future.

It does not yet compile on Windows. Not sure if anyone ever testet to compile any SWIG-Extension for PHP7 on Windows and this is a bug in SWIG or if there is still anything missing but i thought, it would be good to make the current state available and invite everybody to test this on different environments and to a little QA :)

My branch is here: https://github.com/AlexanderGabriel/mapserver/tree/PHP7-MapScript
Please note, you have to include mapscript.php as already written here to make Classes and Constants available: http://mapserver.org/development/rfc/ms-rfc-117.html

If you have any Feedback or suggestions, write here so we can discuss.

Kind regards,
Alex

sdlime:
Hi Alex: I thought @jmckenna already did this work. How does your work support or build on Jeff's? --Steve

AlexanderGabriel:
Hi Steve,

i'm not sure but it is possible that i missed something.
I hat a look at his rep, linked in the RFC but he made changes to the existing Mapscript-Extension. 
If i understand the RFC correkt, the goal is not to update the existing code but use the swig-configuration for future and remove the old code someday.
I just read again. He suggested to use phpng as prefix i used php7. I can change this or merge everything together and we do some auto-detection.
I asked him per email some days ago but did not yet receive any answer (it was on wednesday so this is okay).

Do you have a link to his work? Google didn't help :(

Alex

AlexanderGabriel:
HI,

i did a little more work and i'm kind of done for now. At https://github.com/AlexanderGabriel/mapserver/tree/PHPNG-Mapscript there is a branch for adding PHP-Support via SWIG.
[...]

PR #5642, opened 2018-08-14
Add PHP Mapscript via SWIG

AlexanderGabriel:
Hi,

another try.
This will add support for PHP Mapscript via SWIG and it builds with travis and appveyor.

Kind Regards,
Alex


geographika:
I see you need to build SWIG from source to get this to work on Windows. Any idea when a SWIG release is planned that will produce the PHP MapScript?

It may also be worth moving the PHP build commands to a separate script file in appveyor.yaml in the future (I'll try to do the same with Python) to keep it easier to read (similar to the Travis script).

AlexanderGabriel:
It' even worse. I build a developement state of swig (didnt update for a while) for PHP but we still need swig from gisinternals sdk for csharp... it is not possible to use the built swig for csharp because of any other bug.
this is of course only a workaround. when a version of swig will be released that does not have any problems with php and csharp i don't know.

building php would be unnessesary aswell if it was in the sdk or the needed files were available as download somewhere.

I really tried a lot of things. look at the commit history of the other PR i already closed and this is only the stuff i pushed to github.
I didn't want to upload php or swig compiled to my homepage and create another unrelyable dependency.
This was the way/workaround I was able to get this working and i think, this is a good point to start and improve this.
If you seperate the python commands, I can look how that works and do this with php, too but I'm really a beginner with appveyor ;)

AlexanderGabriel:
I tried chocolatey and installed php but it does not contain dvelopement files

AlexanderGabriel:
Would it be better if i opened the pullrequest against master or can you switch this when merging?

geographika:
@AlexanderGabriel - would you be able to open against the master branch?
Also rather than building SWIG from source how about downloading form http://prdownloads.sourceforge.net/swig/swigwin-3.0.12.zip - this can then be cached in Appveyor which will speed up build times. 
Are all PHP versions required for Travis? The build time is up to 51 minutes (see https://travis-ci.org/mapserver/mapserver/builds/416091387). Perhaps one of the 5.x versions could be dropped. 

I'll then test locally and see if we can get this merged.

AlexanderGabriel:
Of course. Working on it right now. I dropped building with PHP 5.5. It isn's supported any more so we don't have to test with it.
I need SWIG from git because it fixes a problem when building thread-safe on Windows but PHP and SWIG could be cached. This would save time.

geographika:
@AlexanderGabriel - excellent!
I have finally got the .NET MapScript bindings working with SWIG 3.0.12 - see #5583 
This includes a few more Appveyor changes but it should be easy enough to merge later. 
Does SWIG 3.0.12 resolve the thread-safe issue? It means that PHP and .NET bindings could be combined in a single build step. I also plan on adding Python3 to Appveyor - this could be combined with one of the PHP versions to speed up build time.

AlexanderGabriel:
@geographika SWIG 3.0.12 does not solve this problem. it is from january and the fix is from december last year: https://github.com/swig/swig/commit/224bb9e0232a219b7cefd23472b592b39e277f46
In addition, there is no build system provided for windows. i created cmake files to be able to build it. not really a good solution but i didn't find any other way to get this running.
building php while building mapserver is not optimal at all but i couldnt find the needed files on php homepage to download... i could stage the stuff on my website. this would speed up everything but would only add more unnessesary dependencies to external ressources.
I'll prepare a pull request now against master branch and then we can try to improve this over time.
I'll be back ;)

geographika:
@AlexanderGabriel - it is possible in Appveyor to leave files/folders between builds which could be done for both SWIG and PHP. Personally I wouldn't be against hosting the compiled files/libs externally and downloading them similar to the SDK. Ideally the PHP libs could be added to the SDK at GISInternals. This would retain the focus of Appveyor on the MapServer build and CI. Maybe it is something to take to the mailing list?

AlexanderGabriel:
@geographika I already cache the sdk in my working branch (https://github.com/AlexanderGabriel/mapserver/tree/TryToFixAppVeyorAndSpeedup), now working on php. it costs me so much time to wait every time so i think, it's worth it.
the sdk would really be a good place to keep php and swig (in two versions until new swig version gets released). in the readme under mapscript/phpng is everything that is needed to build php and swig.

geographika:
I've had a quick attempt to run things locally. A couple of questions. 

SWIG compiled fine. I presume the patch required for PHP will be in the future SWIG 4.0 release?
I guess this could happen in the near future: https://github.com/swig/swig/issues/1299

With regards to the compiling PHP, is it possible to use the binaries at https://windows.php.net/download#php-7.2-ts-VC15-x64 ? Or does this output differ from the Appveyor steps?

AlexanderGabriel:
Hi Seth,

i thought about this today, too. I need current swig "only" for thread-safe php&mapscript on windows. Non-thread-safe mapscript for windows is really unuseful (can't load php it into apache as module for example) but we will for sure not be happy if mapserver relies on unstable software-dependencies.
In addition, appveyor changed cmake to 3.12 last week and now the building fails because swig cannot find swig.swg but only if i use my build so there is something broken aswell or this is a bug in cmake on windows. did you have any problems with SWIG in AppVeyor since 28th of Sept?
so i thought it would be best to go one step back, build php-mapscript non-thread-safe but based on stable releases like swig 3.0.12 (which works) and list thread-safety-support as known issue until swig-4.0.0 is released. this will be much less work for all of us.

for the cache: php is too big (>1GB) but i hat the idea to delete dependencies after compiling, zipping everything and download it from my server or cache it. depends on what is faster. even don't do a snapshot build saves time.

So what i will do now is:
- Build everything for Non-Thread-Safe PHP
- Use stable SWIG 3.0.12
- Make it work with CMake 3.12.x

Would that be ok?
And after SWIG-4.0 gets released, we solve the thread-safe-problem on windows. I think, theres more people waiting for php-mapscript-support on linux than on windows ;)

AlexanderGabriel:
> With regards to the compiling PHP, is it possible to use the binaries at https://windows.php.net/download#php-7.2-ts-VC15-x64 ? Or does this output differ from the Appveyor steps?

No. This does not contain development header... I tried this in the beginning. A few years ago, php offered headers to download...

AlexanderGabriel:
this can last a few days more. not sure, what the problem is. i opened an issue at cmake. can't build phpng-mapscript with cmake 3.12.3 but 3.11.4 works: https://gitlab.kitware.com/cmake/cmake/issues/18421

geographika:
@AlexanderGabriel - I've sent a mail to the dev list to see if hosting the PHP SDK and downloading to Appveyor is ok. See https://lists.osgeo.org/pipermail/mapserver-dev/2018-October/015511.html
Is it ok with the hosting costs on your side?
Agree - let's go with SWIG 3.0.12 - I have this setup and ready to add in #5583 - and will merge to master this weekend if there are no objections. 

Do we need to test all PHP v7 versions on Travis including nightly? Would sticking to just 7.2 be enough, and document this is the "supported/tested" version?

I've read through the details in #5557 and the RFC on the approach taken, and have a couple of questions.

Is there any easy test test that can be added to verify the build (even just outputting the MapServer version number through PHPNG MapScript)? The php7module.i doesn't contain any custom code - is this because it is unnecessary, or is this something to be added to in the future?

We'll get this merged soon!

AlexanderGabriel:
> Agree - let's go with SWIG 3.0.12 - I have this setup and ready to add in #5583 - and will merge to master this weekend if there are no objections.

Already saw this. Will wait for the merge.
I was able to strip down PHP-SDK to 168MB after building and mapserver still builds. Disabling different modules saves time, aswell. We'll see if it is necessary to buffer this somewhere or build it every time. Hosting it on my server is ok.

> Do we need to test all PHP v7 versions on Travis including nightly? Would sticking to just 7.2 be enough, and document this is the "supported/tested" version?

We can drop this. Travis is new to me so i was playing around with it. Thought it would be cool to build against nightly builds ;)

> I've read through the details in #5557 and the RFC on the approach taken, and have a couple of questions.

Not sure if i understand this. You have questions to me? Shall we meet in the irc?

> Is there any easy test test that can be added to verify the build (even just outputting the MapServer version number through PHPNG MapScript)?

Sure. phpinfo() or something and analyze the output. Didn't try something like this yet.

> The php7module.i doesn't contain any custom code - is this because it is unnecessary, or is this something to be added to in the future?

Not sure, why there should be custom code. The interface is defined in mapscript.i. If everything is fine, there must not be any code here? This file exists because @geomunion added some information to phpinfo().
There is one message when runnig swig:
CustomBuild:
  Swig source
C:\projects\mapserver\mapscript\v8\..\..\mapserver.h(932): warning 314: 'empty' is a PHP keyword, renaming to 'c_empty' [C:\projects\mapserver\build\mapscript\phpng\php_mapscriptng.vcxproj]
I think, this can be masked in php7module.i but there is nothing more i plan to put there but hey - we are at the beginning ;)

geomunion:
The `php7module.i` is used to generate php-module specific code to display MapScript version etc. via
```php
<?php
phpinfo();
```
see: 
* http://www.swig.org/Doc3.0/Php.html#Php_nn2_7
* http://php.net/manual/en/function.phpinfo.php

geographika:
@AlexanderGabriel - you answered my questions in your above comment - thanks!
Is it possible to download https://files.digital-infinity.de/php-sdk-$ENV{VSCMD_ARG_TGT_ARCH}-nts.tar.gz as part of the Appveyor script and simply unzip and point paths to the SDK?
It might be better in the long term to keep CMake focused on MapServer and let the CI take care of gathering dependencies - this would also mean users can build with the own dependencies if needed.

AlexanderGabriel:
> Is there any easy test test that can be added to verify the build (even just outputting the MapServer version number through PHPNG MapScript)?

AppVeyor has chocolatey installed. i can install php with cocholatey and it should be easy to install mapscript-module after building and produce some test-output.

AlexanderGabriel:
New PR here: https://github.com/mapserver/mapserver/pull/5675

PR #5645, opened 2018-08-18
Updated Python MapScript Build Process

geographika:
This pull request contains the updates to the Python CMakeLists including:

- handling Python2 and Python3
- allow newer CMake SWIG features when available
- use a template setup.py.in file
- builds a Python Wheels as output, rather than simply the mapscript.py and mapscript.pyd files
- adds a new CMake target - `cmake --build . --target pythonmapscript-wheel --config Release` - this is used to build the Wheels at https://pypi.org/project/mapscript/
- the previous process CMake process builds Python MapScript exactly as before, so should be fully backwards compatible

Along with the above the pull request:

- removes the old setup.py and setup.cfg which were no longer used (even prior to this pull request as CMake was building mapscript)
- updates a couple of the Python MapScript tests
- adds a new ReadMe.rst for Python MapScript, which is also added to the package and uploaded to PyPI

See the wiki page at https://github.com/mapserver/mapserver/wiki/Python-MapScript-Updates-2018 for ongoing updates to the Python MapScript bindings


rouault:
Would it be possible to add in Travis-CI testing of the wheel building target ?

geographika:
@rouault - yes this is on the todo list. I'll hopefully get a look at it this week and will add to this pull request. Currently none of the Python tests are run on Travis so it will be good to get these added back in.

rouault:
What does the wheel exactly contain ? The mapscript.py and  the corresponding .dll/.so right ? Not libmapserver.so/dll, and all its dependencies ? This could probably be clarified in the "Wheels and PyPI" section of mapscript/python/README.rst so people have a clear understanding that they also need to install the native mapserver binaries alongside the wheel (I'm saying that since for example some rasterio wheels provide the full GDAL stack with them)

geographika:
@rouault - I've made a few changes that hopefully address these. The wheel only contains the mapscript.py and .so, plus the examples folder (which is very outdated but I hope to sort out for a next release). They are all in a separate folder now which helps keep Python environments cleaner, and can be easily pip uninstalled. 

- the wheel is now built, installed, and the Python test suite run on Travis (and Appveyor). Unfortunately there is no stdout output written to the logs. This may be hidden by Travis? Also I was not sure where to best place the wheel command. I added a new target in the Makefile and call this separately in .travis.yml
- I have updated the README to make it clearer the binaries need installing.

geographika:
If there are no further comments I'll go ahead and merge this later today, to then focus on updating the Python MapScript examples distributed in the Wheel package.

PR #5650, opened 2018-08-20
mapogcsld: fix potential overflow when writing hexcolor 

gogglesguy:
Fix potential overflow when writing hexcolor in the unlikely case color components are bigger than 255. Just keeping the compiler happy.


sdlime:
Thanks for all of these!

Issue #5652, opened 2018-08-22
MapServer 7.2 produces empty MVT from point data


jratike80:
Discussed in mapserver-users thread https://lists.osgeo.org/pipermail/mapserver-users/2018-August/080679.html

Mapfile and data (OSM and Natural Earth) for reproducing the issue available for a limited period at http://193.167.189.71/temppi/.  Mapfile also as an attachment.
[osm_finland.txt](https://github.com/mapserver/mapserver/files/2310929/osm_finland.txt)

sdlime:
Found the error, patch forthcoming...

PR #5655, opened 2018-08-29
Fixed clipping error for point layers (#5652).

sdlime:
Stupid error meant no single point features escaped clipping. --Steve


jratike80:
I am very bad in reading the code but I do not understang how this new line makes things better if numlines=1 and layer type is point, because the bounding box is just a point and xmin=xmax and ymin=ymax.

`if(shape->numlines > 0 && (layer_type == MS_LAYER_POINT || (shape->bounds.minx != shape->bounds.maxx || shape->bounds.miny != shape->bounds.maxy))) `

PR #5656, opened 2018-09-01
Python CMakeLists updates

geographika:
Fix version check, add some status messages, and remove redundant endif syntax, attempt to hide linker warnings


geographika:
Superseded by #5657

PR #5657, opened 2018-09-04
Add test suite to Python Wheel Builds

geographika:
This pull request does the following:

+ Adds the Python MapScript test suite and associated test data to the Python Wheel. This make it easy to run the tests following the installation of MapScript and check all parts of MapScript and MapServer are working correctly
+ Simplified the test running process so it can be run with a single command using pytest (and renamed test files to match the standard naming convention)
+ Updated the CMake wheel target to create a virtual environment, install the newly created wheel, and run the test suite on both Travis and Appveyor
+ Updated README.rst that will be displayed in PyPI and on GitHub detailing the testing and modified build process

Note the existing Python MapScript build process is unaffected. The wheels are only built when running a specific CMake target:

` cmake --build . --target pythonmapscript-wheel`


mapscript/python/tests/cases/testing.py
```diff
@@ -41,22 +41,25 @@
 import sys
 import distutils.util
 import unittest
+import tempfile
 
 # define the path to mapserver test data
-test_path = os.path.abspath(os.path.join(__file__ ,"../../../../.."))
-TESTS_PATH = os.path.join(test_path, "tests")
+TESTS_PATH = os.path.abspath(os.path.join(os.path.dirname(__file__ ),"data"))
+# TESTS_PATH = os.path.join(test_path, "tests")
```
rouault: why not removing the commented code here and below ?

appveyor.yml
```diff
@@ -42,17 +42,9 @@ build_script:
...
-  - "%PYTHON_EXECUTABLE% -m pip install -r requirements-dev.txt"
-
 test_script:
   - set PATH=%BUILD_FOLDER%/build/Release;%BUILD_FOLDER%/sdk/%SDK%/bin;%PATH%
   - set PROJ_LIB=%BUILD_FOLDER%/sdk/%SDK%/bin/proj/SHARE
-  - set PYTHONPATH=%BUILD_FOLDER%/build/mapscript/python/Release;%BUILD_FOLDER%/build/mapscript/python
-  - cd %BUILD_FOLDER%
-  - set TESTS_FOLDER=%BUILD_FOLDER%/mapscript/python/tests/cases
-  - "%PYTHON_EXECUTABLE% -m pytest --ignore=%TESTS_FOLDER%/fonttest.py --ignore=%TESTS_FOLDER%/hashtest.py --ignore=%TESTS_FOLDER%/pgtest.py --ignore=%TESTS_FOLDER%/threadtest.py %TESTS_FOLDER%"
+  - cmake --build . --target pythonmapscript-wheel --config Release
```
rouault: the testing done previously with python -m pytest is no longer done ?

mapscript/python/tests/cases/testing.py
geographika: Yes these should be cleaned up. I was going to pylint the whole test suite as a separate pull request.

appveyor.yml
geographika: The pythonmapscript-wheel target now takes care of running the tests as part of the build process. It runs them on an installed wheel in a new virtual environment for both Travis and Appveyor

rouault:
OK, LGTM as far as I'm concerned

geographika:
Fixed conflicts and merged. Will address code clean up and pylinting in a separate request.

PR #5672, opened 2018-10-02
Make Travis stop building with php 5.5

AlexanderGabriel:
Make Travis stop building with PHP-5.5 (not supported any more. see http://php.net/supported-versions.php)


AlexanderGabriel:
Is included in https://github.com/mapserver/mapserver/pull/5675#issuecomment-427554698

PR #5673, opened 2018-10-02
Make AppVeyor cache the SDK

AlexanderGabriel:
Make AppVeyor cache the SDK

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,7 +17,7 @@ shallow_clone: true
 
 build_script:
   - set "BUILD_FOLDER=%APPVEYOR_BUILD_FOLDER:\=/%"
-  - set "LocalAppDataFolder=%LOCALAPPDATA:\=/%"
+  - set "SDK_ROOT=%LOCALAPPDATA:\=/%/sdk"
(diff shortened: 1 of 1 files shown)


geographika:
@AlexanderGabriel - a few notes on this!

Apparently Appveyor doesn't use the cache by default on pull requests, as the pull request could put malicious files in the cache - see https://www.appveyor.com/docs/build-cache/ (so I guess the cache would only be used when tagging new releases?). 

I enabled the option in the project settings to see what kind of speed-up we'd get, and it actually made things slower!

See: https://ci.appveyor.com/project/MapServer/mapserver/build/1.0.874 (no cache)
Compared to: https://ci.appveyor.com/project/MapServer/mapserver/build/1.0.875 (with cache)

Appveyor zips up the folder to be cached, then downloads and unzips it for the new build. This appears to be slower than simply downloading and unzipping directly from GISInternals - adding almost a minute to the build times. I guess the benefit would be to stop hitting the GISInternals servers. 

So probably leave this pull request for now? Although building SWIG and PHP libs once and caching the built output I believe be beneficial (although it seems even quicker to simply download a zip from elsewhere).

AlexanderGabriel:
@geographika 
Yes. If it doesn't help, we leave it.

PR #5675, opened 2018-10-05
Phpng mapscript master

AlexanderGabriel:
So now another Pull-Request for adding PHP-MapscriptNG-Support to the master-branch.
This PR does not build PHP-MapscriptNG in AppVeyor for this reason:
- CMake 3.12.4 which is used in AppVeyor right now is buggy and prevents building PHP-Extensions on Windows with SWIG
When Appveyor updates CMake the next time, it will be possible to build this extension there.
Users which want to build it can do it. CMakeLists.txt checks for right Version of CMake.

And because there where a log of PRs from me with a lot of information. Here a small summary:
- It is not possible to build MapscriptNG Thread-Safe (for example use it an apache with PHP as Module) ON WINDOWS because of a bug in SWIG < 4.0. Users which are interessted can build SWIG themselves and get this running. CMakeLists.txt checks for SWIG 4.0.0 if Thread-Safety is enables when building mapserver. With SWIG > 3.0.11 it will be possible to build this extension on Windows for non-thread-safe PHP
- When building this extension on Windows. Users will need to build php on their own because the windows-downloads do not contain the required developement headers and downloaded sources are not configured. Same with chocolatey-package. But this process is not that difficulty and frustrating as a few years ago. In README.WIN32 under mapscript/phpng is how to build PHP and SWIG and Mapserver for interested users but PHP-MapscriptNG on Windows should not be advertised as supported right now.
Maybe experimental?

Alex


geographika:
Looks good to me. When the new versions of SWIG and CMake are released we can add to Appveyor. 
I guess #5672 can be closed as included above?
Maybe post to the mapserver-dev list for any final thoughts/feedback and if there are no suggested changes I'll merge in a few days?

AlexanderGabriel:
Just for informatoin:
I found another thing, that can be improved in the future:
SWIG renames webObj->empty to webObj->c_empty which is wrong. empty is listed as keyword (http://php.net/manual/en/reserved.keywords.php) but not as "empty" but as "empty()".
This allows "empty" to be a variable or property of an object but no a function or method.
And if that was wrong, accessing "empty" property of webObj the last 15 years via PHP-Mapscript wouldnt have worked, too (https://mapserver.org/mapscript/php/phpmapscript.html#webobj).

I already tested to patch SWIG locally and it works. I'll try to make a PR to SWIG to get this into SWIG 4.0.0 so it's only a short time, linux-users need to use $webObj->c_empty.

geographika:
@AlexanderGabriel - I know we can't get the Appveyor builds for the new MapScript bindings, but is it possible to build on Travis? I've updated the SWIG version as a first step, but get failures when trying to build e.g. https://travis-ci.org/geographika/mapserver/builds/515453216

AlexanderGabriel:
i'll have a look.

AlexanderGabriel:
i'm sorry but i can't get it running. locally, everything is fine but it does not work with travis.

btw: there is a swig 4.0.0-beta1 downloadable and i'm sorry but i really have no time currently.

PR #5676, opened 2018-10-11
Silently ignore WMS filter if the provided value is empty

ravhed:
New pull request targeting branch-7-2 instead of master.

If the WMS filter parameter is empty (e.g. FILTER=) you will get a 500 error. This pull request changes this behavior so that it will instead ignore empty filters.

See discussion in #5597 for more details.

### DIFF
--- a/mapwms.c
+++ b/mapwms.c
@@ -323,7 +323,12 @@ int msWMSApplyFilter(mapObj *map, int version, const char *filter,
   char **paszFilters = NULL;
   FilterEncodingNode *psNode = NULL;
 
-  if (!map || !filter || strlen(filter)==0)
+  // Empty filter should be ignored
+  if (!filter || strlen(filter) == 0)
+    return MS_SUCCESS;
+  
+
+  if (!map)
     return MS_FAILURE;  
 
   /* Count number of requested layers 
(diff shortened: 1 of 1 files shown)


mapwms.c
```diff
@@ -454,7 +459,7 @@ int msWMSApplyFilter(mapObj *map, int version, const char *filter,
       errorObj* ms_error = msGetErrorObj();
 
       if(ms_error->code != MS_NOTFOUND) {
-	msSetError(MS_WFSERR, "FLTApplyFilterToLayer() failed", "msWFSGetFeature()");
+	msSetError(MS_WMSERR, "FLTApplyFilterToLayer() failed", "msWFSGetFeature()");
```
geographika: Should the function name for the error should also be corrected to `msWMSApplyFilter` ?

ravhed: You're right, thanks!

geographika:
Looks good to me. Anyone any objections with merging?

dmorissette:
I didn't test the patch, but it looks okay to me.  +1 to merge.

sdlime:
+1

geographika:
@ravhed - thanks!

PR #5679, opened 2018-10-12
Change URL to OpenLayers to be protocol relative

AlexanderGabriel:
There is a built-in OpenLayers viewer in Mapserver which uses http://www.mapserver.org/lib/OpenLayers-ms60.js as URL to OpenLayer.
I was able to use this over https only if i tell my browser to allow mixed-content.

This patch changes the url to be a protocol relative URL. This causes Firefox to use https even if i call the viewer via http.
I tested and it works a little bit better now, because there are still http://-urls to the images in OpenLayers-ms60.js:

grafik

grafik (in german, sorry)

If you merge this and anyone changes OpenLayers-ms60.js on your homepage to do the same, users can use the built-in OpenLayers viewer over https out-of-the box.

Thank you.

Alex

### DIFF
--- a/maptemplate.c
+++ b/maptemplate.c
@@ -43,7 +43,7 @@
 static inline void IGUR_sizet(size_t ignored) { (void)ignored; }  /* Ignore GCC Unused Result */
 static inline void IGUR_voidp(void* ignored) { (void)ignored; }  /* Ignore GCC Unused Result */
 
-static char *olUrl = "http://www.mapserver.org/lib/OpenLayers-ms60.js";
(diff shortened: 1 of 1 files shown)


AlexanderGabriel:
Forget my comment about to correct OpenLayer.js. This is already relative. So merging this will solve this problem. Nothing more to do.

Alex

geographika:
@AlexanderGabriel - thanks. 
Just to summarise this patch allows the inbuilt OpenLayers viewer to be requested over either http or https, depending on the parent page.

PR #5682, opened 2018-10-18
PHP Mapscript native for PHP7

AlexanderGabriel:
Merging this will bring back support for "native" PHP/Mapscript on both, Windows and Linux for thread-safe and non-thread-safe environments.
My patches base on the work of @bjoernboldt who enabled support for non-thread-safe environments.

Because there are no module global variables there is no need to mention about thread-safety. I removed code that checks for ZTS-switch.
There is still code that should be unneeded (TSRMLS_*) but removing this will break building for PHP5 so we need to keep this for backwards-compatibility.
The WITH_THREAD_SAFETY-switch is still needed to make PHP-headers work correct while compile-time.

In addition, i merged FindPHP5.cmake and FindPHP7.cmake into FindPHP.cmake to reduce some redundant code.


ejn:
Do f54ce3a and a5049b7 belong in this PR? Neither of them appear to have anything to do with PHP7 but rather appear to correct some problems in msQueryByFilter() and so should maybe form a separate PR?

Maybe @bjoernboldt can comment?

AlexanderGabriel:
seems they are not really realted to php mapscript. only https://github.com/mapserver/mapserver/pull/5682/commits/9dbfe1b23d9c9529c9c980bd3de79dbf2033f388 is related. sorry. i updatet his branch and then started to work. i will open a new PR so you can decide, which will be merged.

AlexanderGabriel:
New PR here: https://github.com/mapserver/mapserver/pull/5682

chseip:
May I direct your attention here: https://gis.stackexchange.com/questions/308999/how-to-compile-install-php-mapscript-with-php7
I'm lost :-(

PR #5687, opened 2018-10-22
PHP/Mapscript native for PHP7

AlexanderGabriel:
This PR only contains Commits related to PHP/Mapscript.
Now it will be possible to build on Windows and Linux in Thread-Safe and Non-Thread-Safe Environment.
I also merged FindPHP7.cmake and FindPHP5.cmake into one file for both.


geographika:
@AlexanderGabriel - looks good to me, and there has been no objections to these changes which have been up here for awhile. If no objections from anyone I think this can be merged tomorrow.

Issue #5490, opened 2017-10-13
IN operator is not working with PostGIS expression translation


sdlime:
MapServer expressions with the IN operator are not being translated to native SQL for PostGIS. Should verify that other drivers aren't suffering from the same oversight. 

This primarily affects attribute queries where expressions must be given as MapServer expressions. Filters are also affected but there are a couple of workarounds in that case:

1. use the NATIVE_FILTER processing key
2. put the SQL directly in the DATA statement

--Steve

ejn:
Just run into this problem too - `FILTER ('[attr]' IN 'my,list,of,items')` gets translated to `"attr"::text'my,list,of,items'`

@sdlime, have you made progress on this? Since it has now missed the 7.2 release, would it be worth putting a warning in the documentation until it gets fixed in a release?

sdlime:
Am working on a fix now - in the 7.0 branch... Trick is handling both string and numeric lists.

sdlime:
The fix has been added into the 7.2 branch, feel free to give it a try and let me know if you run into issues. I've added some regression tests and it seems to work fine. Closing... --Steve

PR #5696, opened 2018-11-06
Initial fix for issue #5490 (for PostGIS only).

sdlime:
This adds IN operator support to the PostGIS filter translation function. It also enables IN support in other drivers (e.g. Oracle, OGR, SQLServer) because they were expecting one token for that operator (MS_TOKEN_COMPARISON_IN) but the lexer was returning IN - the parser token for that operator. This patch fixes that too... --Steve


sdlime:
Added review requests from driver maintainers... Crap, I see tests are failing though.

geographika:
@sdlime - I think both the CI scripts on the 7-0 branch are broken - they've been updated in the 7-2 branch and master.

sdlime:
@geographika, looks that way... This particular issue was introduced at 7.0 so I started with that branch. I'm curious if folks think we should bother with a fix there or if I should redo the pull for 7.2 only?

Issue #5706, opened 2018-11-21
mapquery.c "No matching record(s) found." filling up logfiles


computerchemist:
My mapserver wms service log files fill up very quickly with these messages: "No matching record(s) found.", even with '`DEBUG 0`'

The frontend is an openlayers query with many layers; sometimes there is no match for the area/filter, this is not an error in itself.

I would suggest that all messages in mapquery.c are changed to only include this verbose output if verbosity is required in debugging - in other words, for all occurrences of 
`msSetError(MS_NOTFOUND, "No matching record(s) found.",  ......`

add this following line above the seterror, to only trigger the errormessage (but still exit with an error condition) if debug is required
`if( lp->debug >= MS_DEBUGLEVEL_V )`

I have been routinely making this change for several versions now but it would be nice if it was in the codebase - comments?

geographika:
That change sounds logical to me. Are you able to create a pull request with your change?

computerchemist:
Seth, feel free to add it in!
There are 5 changes in total to make in mapquery.c .

geographika:
@computerchemist - you can make these changes yourself directly in GitHub in the browser to create a pull request. See https://help.github.com/articles/editing-files-in-another-user-s-repository/ (I'm assuming you're not familiar with git - apologies if you are!). 
This will then trigger the test suite and can be reviewed by others.

pedros007:
Linking issue to #5707

PR #5707, opened 2018-11-21
Reduce log clutter from non-errors

computerchemist:
Prevent "No matching records" from filling up logfile in terse mode.
https://github.com/mapserver/mapserver/issues/5706

### DIFF
--- a/mapquery.c
+++ b/mapquery.c
@@ -986,7 +986,8 @@ int msQueryByFilter(mapObj *map)
       return MS_SUCCESS;
   }
 
-  msSetError(MS_NOTFOUND, "No matching record(s) found.", "msQueryByFilter()");
+  if( lp->debug >= MS_DEBUGLEVEL_V )
+    msSetError(MS_NOTFOUND, "No matching record(s) found.", "msQueryByFilter()");
   return MS_FAILURE;
 
 query_error:
@@ -1284,7 +1285,8 @@ int msQueryByRect(mapObj *map)
       return(MS_SUCCESS);
   }
 
-  msSetError(MS_NOTFOUND, "No matching record(s) found.", "msQueryByRect()");
+  if( lp->debug >= MS_DEBUGLEVEL_V )
+    msSetError(MS_NOTFOUND, "No matching record(s) found.", "msQueryByRect()");
   return(MS_FAILURE);
 }
 
@@ -1601,7 +1603,8 @@ int msQueryByFeatures(mapObj *map)
     if(GET_LAYER(map, l)->resultcache && GET_LAYER(map, l)->resultcache->numresults > 0) return(MS_SUCCESS);
   }
 
-  msSetError(MS_NOTFOUND, "No matching record(s) found.", "msQueryByFeatures()");
+  if( lp->debug >= MS_DEBUGLEVEL_V )
+    msSetError(MS_NOTFOUND, "No matching record(s) found.", "msQueryByFeatures()");
   return(MS_FAILURE);
 }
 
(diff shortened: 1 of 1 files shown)


mapquery.c
```diff
@@ -1284,7 +1285,8 @@ int msQueryByRect(mapObj *map)
       return(MS_SUCCESS);
   }
 
-  msSetError(MS_NOTFOUND, "No matching record(s) found.", "msQueryByRect()");
+  if( lp->debug >= MS_DEBUGLEVEL_V )
```
geographika: Probably best to put these into braces to avoid bugs in the future:

```
if( lp->debug >= MS_DEBUGLEVEL_V ) {
    msSetError(MS_NOTFOUND, "No matching record(s) found.", "msQueryByRect()");
}
```

geographika:
@computerchemist - Travis is failing due to a separate issue. I have changed the pull request to be on the master branch where Travis should be working. If you add in the braces Travis should restart automatically.

mapquery.c
computerchemist: Agreed! Thanks for the headsup.

geographika:
@computerchemist - thanks for the pull request!
I may have been mistaken about why Travis was failing. It looks like some tests may be relying on that text e.g. https://github.com/mapserver/mapserver/blob/master/msautotest/wxs/expected/wfs_200_exception_getfeaturebyid_nofeature.xml
Hopefully others will provide input on if this should be expected or not.

computerchemist:
@geographika I had a quick look, doesn't the context of the test just needs changing so the test is performed in verbose debug mode? The returned WMS makes it clear in any case there are no features to return - that should be really more the focus of the test -  the log item itself was IMO pretty useless as it was just telling me "something" had no features. But NOT seeing it repeated literally tens of thousands of times in a production log was the aim of the patch! 

I also double checked talking of context to ensure "lp" was there in each function; just something that alarmed me from Travis; `warning: ‘lp’ may be used uninitialized in this function`. It is - it's initialised via GET_LAYER from the mapObj "map".

I suspect I might be better off using "map->debug", becuase "map" as a mapobj seems to be passed to each function (and DEBUG I think is a global function anyway - is it even able to be turned on for selected layers?). Does that sound like a better idea in any case?

I simply used "lp->debug" as existing error messages were using it. It made ME do a review too and on that basis I think I only got away with it as it's in the layerobj structure, although it all seems to be passed from the mapobj using GET_LAYER, so maybe I'm just confusing travis more than myself.

I'm hoping others can weigh in too at this point who are more knowledgeable with the codebase - all I want is nice clean manageable error logs with only real errors in them!

Thanks for the help again Seth.

geographika:
@computerchemist - a good summary of the situation, it looked a nice easy update at the beginning! I agree with your logic that it should be possible to log only errors in a production system. 
I put in link in the mailing lists as there was an ongoing discussion about DEBUG levels which seemed relevant - https://lists.osgeo.org/pipermail/mapserver-users/2018-November/080870.html
The pull request should be left open to let others add opinions.

jmckenna:
I believe we should merge this pull request, as it is indeed very useful.   (just for reference: those log messages will then require at least debug level 3, "verbose").

jmckenna:
thank you @computerchemist !!  A nice improvement.

ravhed:
I think it was a mistake to merge this pull request into master in its current state as the tests are failing because of this pull request (they currently depend on this error message). It seems to me that there was no decision on how to handle this change so maybe we should revert it and decide on how to properly handle this? I think @geographika's email (https://lists.osgeo.org/pipermail/mapserver-users/2018-November/080870.html) summarizes the issue nicely.

sdlime:
@jmckenna, @geographika - what do you think? Could just fix the tests. I don't know which or how many tests are failing now but is it simply a matter of adding the appropriate debug level to those tests which depend on that messaging?

geographika:
The tests need to be updated - we don't want to get into a situation of allowing a few tests to fail. 

The ServiceExceptionReport could be improved by including a simple text string of:

> No matching records

jmckenna:
No problem, will revert now....

jbo-ads:
I am working on an update of failing msautotests with respect to this PR. It consists in adding a `DEBUG 3` line in all impacted layers so that output remains the same.

jmckenna:
thanks @jbo-ads i was going to tackle that this afternoon, you beat me to it :)

geographika:
I think with the update from @jbo-ads this pull request can be reopened and merged?

PR #5710, opened 2018-11-26
Styles' OFFSET values are saved as ints though they are doubles [issu…

### LINKED ISSUE #5310: Styles' OFFSET values are saved as ints though they are doubles.
In mapserver.h (row 1006) we can see the styleObj struct has offsetx and offsety. They are double.

mapserver.h

struct styleObj{
...
double offsetx, offsety; /\* for shadows, hollow symbols, etc... */
...
}

According to this we can see in mapfile.c (row 2726 and 2736) they are treated as doubles when they are loaded in.

mapfile.c

int loadStyle(styleObj *style)
{
...
case(OFFSET):
if((symbol = getSymbol(2, MS_NUMBER,MS_BINDING)) == -1) return(MS_FAILURE);
if(symbol == MS_NUMBER)
style->offsetx = (double) msyynumber;
else {
...
if(symbol == MS_NUMBER)
style->offsety = (double) msyynumber;
else {
...
}

But in mapfile.c (row 2986) the style is saved by a writeDimension (mapfile.c row 525) function which wait ints on its 4th and 5th operands
that is why the OFFSET value of styleObj losts the precision of double when the map file is saved.

mapfile.c

...

static void writeDimension(FILE *stream, int indent, const char *name, int x, int y, char *bind_x, char *bind_y)
{
writeIndent(stream, ++indent);
if(bind_x) msIO_fprintf(stream, "%s [%s] ", name, bind_x);
else msIO_fprintf(stream, "%s %d ", name, x);
if(bind_y) msIO_fprintf(stream, "[%s]\n", bind_y);
else msIO_fprintf(stream, "%d\n", y);
}

...

void writeStyle(FILE *stream, int indent, styleObj *style)
{
...
writeDimension(stream, indent, "OFFSET", style->offsetx, style->offsety, style->bindings[MS_STYLE_BINDING_OFFSET_X].item, style->bindings[MS_STYLE_BINDING_OFFSET_Y].item);
...
}

The solution for this issue could be anther writeDimension function with 2 double operands.

thisIsShubhamSharma:
…e #5310]

fixes #5310 and done some refactoring

### DIFF
--- a/mapfile.c
+++ b/mapfile.c
@@ -96,32 +96,32 @@ char *msPositionsText[MS_POSITIONS_LENGTH] = {"UL", "LR", "UR", "LL", "CR", "CL"
 */
 int msValidateParameter(const char *value, const char *pattern1, const char *pattern2, const char *pattern3, const char *pattern4)
 {
-  if(msEvalRegex(pattern1, value) == MS_TRUE) return MS_SUCCESS;
-  if(msEvalRegex(pattern2, value) == MS_TRUE) return MS_SUCCESS;
-  if(msEvalRegex(pattern3, value) == MS_TRUE) return MS_SUCCESS;
-  if(msEvalRegex(pattern4, value) == MS_TRUE) return MS_SUCCESS;
+    if(msEvalRegex(pattern1, value) == MS_TRUE) return MS_SUCCESS;
+    if(msEvalRegex(pattern2, value) == MS_TRUE) return MS_SUCCESS;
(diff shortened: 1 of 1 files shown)


geographika:
@thisIsShubhamSharma - you probably want to try the pull request again - there are too many changes to the file to review (https://github.com/mapserver/mapserver/pull/5710/files). If you auto-formatted the code you'll want to revert this and just add the minimal changes required for the fix with no refactoring.

thisIsShubhamSharma:
Sure

PR #5715, opened 2018-11-30
Add a Python test for checking the OFFSET is output correctly

geographika:
@thisIsShubhamSharma - I added a Python test to create a style with doubles as OFFSET values, convert it to a string, load it back and check the doubles are still there. The test however appears to be failing. Can you convert to string and see that double OFFSET values are preserved?

```
          output_string = new_style.convertToString()
          new_style2 = mapscript.fromstring(output_string)
      
          # ensure attributes are output as doubles
          assert new_style.size == 7.25
  >       assert new_style2.offsetx == 10.5
  E       AssertionError: assert 0.0 == 10.5
  E        +  where 0.0 = <mapscript.mapscript.styleObj; proxy of <Swig Object of type 'styleObj *' at 0x03ED0A10> >.offsetx
```


geographika:
For reference issue was #5310 and fix was by @sdlime with #5738

Issue #5658, opened 2018-09-05
SLD - Missing escaped characters when converting OGC propertyIsLike to Regex


pmauduit:
Using the following getmap request:

```
/usr/lib/cgi-bin/mapserv QUERY_STRING='MAP=/root/resources/version-shapefile/mapfile.map&
SERVICE=WMS&
VERSION=1.3.0&
REQUEST=GetMap&
FORMAT=image%2Fpng&
TRANSPARENT=true&
LAYERS=surval_30140_all_point_19_12_2017&
SLD=http://host-where-my-sld-is-located/sld-shapefile.xml&
CRS=EPSG%3A3857&
STYLES=&
WIDTH=2036&
HEIGHT=686&
BBOX=-79555169.84223041%2C-33479467.308605302%2C79555169.84223041%2C33479467.308605302'
```

The SLD used as parameter above, containing a filter which has the following form (truncated XML content):

```xml
          <ogc:Filter xmlns:xs="http://www.w3.org/2001/XMLSchema">
            <ogc:PropertyIsLike escapeChar="\" matchCase="false" singleChar="?" wildCard="*">
              <ogc:PropertyName>SUPPORTS_N</ogc:PropertyName>
              <ogc:Literal>*Support : Eau - Niveau : Surface (0-1m)*</ogc:Literal>
            </ogc:PropertyIsLike>
          </ogc:Filter>
```

Mapserver will parse the SLD and generate a temporary mapfile to produce the actual output, this mapfile will translate the previous `PopertyIsLike` property into the following expression:

```
    CLASS
      NAME "Unknown"
      EXPRESSION (("[SUPPORTS_N]" ~* ".*Support : Eau - Niveau : Surface (0-1m).*"))
      STYLE
        COLOR 0 92 230
        SIZE 6
        SYMBOL "sld_mark_symbol_circle_filled"
      END # STYLE
    END # CLASS
```
The wildcard character (`*`) is correctly escaped (-> `.* `) as a REGEXP, but the parenthesis are not (`(0-1m)` instead of `\(0-1m\)`). The expected expression should have every REGEXP control characters escaped when generating the temporary mapfile.

Note: The issue was encountered on a oracle spatial datasource but adapted to a shapefile source to simplify its reproduction. In case of an Oracle datasource, another part of the temporary mapfile might be affected by a similar issue, about the `FILTER` directive as below:

```
 LAYER
    CONNECTION "aaa/bbb@oracle"
[...]

PR #5740, opened 2019-01-30
WMS: apply FILTER (RFC 118) vendor parameter after the CRS parameter is taken into account to avoid extent inconsistencies (found during refs #5658 investigation)

rouault:
@dmorissette If you want to have a look at https://github.com/mapserver/mapserver/commit/f8ed7b23b64ebb436f81b31ee500737dd04cb28d that modifies a bit RFC118 related functionality. I observed that when the WMS request involves reprojection, msWMSApplyFilter() resulted in incorrect bounding box in the call chain because the map->projection wasn't yet set, but the overall request still worked, because the querymap rendering re-evaluated the request due to the resultcache being empty

### DIFF
--- a/mapwms.c
+++ b/mapwms.c
@@ -1367,22 +1367,6 @@ int msWMSLoadGetMapParams(mapObj *map, int nVersion,
   }
 
   /*
-  ** Apply vendor-specific filter if specified
-  */
-  if (filter) {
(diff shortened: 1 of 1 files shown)


sdlime:
@rouault and @dmorissette, should this be merged? --Steve

rouault:
+1 from me

Issue #5743, opened 2019-02-18
MSSQL: Add support for curve geometries


szekerest:

szekerest:
Curve geometry support has been introduced in MSSQL2012

sdlime:
@szekerest, can this be closed?

szekerest:
yes

PR #5745, opened 2019-02-19
MSSQL: Add support for curve geometries (#5743)

szekerest:


jmckenna:
@szekerest I merged now, to make it into Steve's 7.2.2 release.  thanks for this!

Issue #5726, opened 2019-01-10
MVT: additional/missing data types


BergWerkGIS:
Similar to #5594 but I'm particularly interested in `Int64/BigInt/Long` support.

Looking at the code
https://github.com/mapserver/mapserver/blob/ffb119137bf339630ff9fd9b441a7745a9f84b6b/mapmvt.c#L270-L284
I see that only `Interger|Real|Boolean` get dedicated conversions whereas everything else gets converted to string.

However the docs for [`gml_[item name]_type`](https://mapserver.org/ogc/wfs_server.html#layer-object) state:
> Specifies the type of the attribute. Valid values are the OGR data types: Integer|**Long**|Real|Character|Date|Boolean. MapServer translates these to valid GML data types. 
Note: **Long is to be used for 64-bit integers, _starting with MapServer 7.0.1._**

Any change to get `Int64/BigInt/Long` support for MVT output format?

The [vector-tile-spec](https://github.com/mapbox/vector-tile-spec/blob/26b7823b16df1e15f55adb44243a3896acb7b804/2.1/vector_tile.proto#L17-L28) would support it 😄:

```proto
        message Value {
                // Exactly one of these values must be present in a valid message
                optional string string_value = 1;
                optional float float_value = 2;
                optional double double_value = 3;
                optional int64 int_value = 4;
                optional uint64 uint_value = 5;
                optional sint64 sint_value = 6;
                optional bool bool_value = 7;

                extensions 8 to max;
        }
```

----

Additional information:
* I use `CONNECTIONTYPE OGR` to connect to SQL Server
* `ogrinfo` identifies the data type correctly:
![image](https://user-images.githubusercontent.com/4549888/50988427-d8a51880-150c-11e9-9024-40b8ca860cce.png)
* By the way: `CONNECTIONTYPE PLUGIN / PLUGIN "msplugin_mssql2008.dll"` gets totally screwed by `BigInt`s.
Either empty string (even for `ogr_fid`) or garbage:
![image](https://user-images.githubusercontent.com/4549888/50988813-ffb01a00-150d-11e9-8585-27cfb8b4deff.png)

sdlime:
I'm wondering if it's enough to have both GML types Int and Long use atol since internally both are handled as 64-bit ints in the MVT schema...

struct  _VectorTile__Tile__Value
{
  ProtobufCMessage base;
  char *string_value;
  protobuf_c_boolean has_float_value;
  float float_value;
[...]

PR #5747, opened 2019-02-21
Added explicit conversion for long attributes for MVT generation.

sdlime:


sdlime:
Addresses #5726... --Steve

PR #5749, opened 2019-02-22
write ScaleDenom to the mapfile

3x0dv5:
Hi, do you guys know if there is a reason why the SCALEDENOM is not written?
Also are those two "deprecated" correct?

They are shown in the documentation as being still used.

### DIFF
--- a/mapfile.c
+++ b/mapfile.c
@@ -6309,7 +6309,7 @@ static void writeMap(FILE *stream, int indent, mapObj *map)
   writeBlockBegin(stream, indent, "MAP");
   writeNumber(stream, indent, "ANGLE", 0, map->gt.rotation_angle);
   writeHashTableInline(stream, indent, "CONFIG", &(map->configoptions));
-  writeString(stream, indent, "DATAPATTERN", NULL, map->datapattern); /* depricated */
+  writeString(stream, indent, "DATAPATTERN", NULL, map->datapattern);
   writeNumber(stream, indent, "DEBUG", 0, map->debug);
   writeNumber(stream, indent, "DEFRESOLUTION", 72.0, map->defresolution);
   writeExtent(stream, indent, "EXTENT", map->extent);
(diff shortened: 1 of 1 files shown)


geographika:
`DATAPATTERN `and `TEMPLATEPATTERN` are both now possible within a `VALIDATION `block. Hopefully there will be a cleanup of the Mapfile syntax for a version 8 release. 
In the absence of a `SCALEDENOM` a default of -1 will be applied?

aimdev:
Hi

I note with interest that there may be a cleanup of the Mapfile Syntax in Version8, which prompts a few questions.

1. Currently I am in the process of developing a Map editor, which is based on the xml definitions (sadly a bit out of date), will this file be brought up to date as part of the Version 8 release?
2. Will Version 8 still support xml base map files?
3. Will the syntax for Version 8 alter radically from the 
THING
stuff
END
structure?
4. When do you expect Version 8 to be release.

Thanks

geographika:
@aimdev - the cleanup will involve a RFC which will need to be voted on. However it won't be anything radical - just suggestions to remove some of the keywords which have been deprecated for many releases (some as far back as v4). 

If you are looking for a recent description of the Mapfile syntax please take a look at https://github.com/geographika/mappyfile/tree/master/mappyfile/schemas (and a slightly older merged version of these - https://github.com/geographika/mappyfile/blob/master/docs/schemas/mapfile.json ). 

The JSON schema can be used for validation or to get keywords, so could form the basis of an editor. There is a possibility to export to XSL so it might be possible to update the XSL schema from this.

aimdev:
Hi

Thanks for the pointers.
Will the json files referred to be the maintained versions in the future?, as I have been working to the xml in the mapserver repository.
I will need to convert them any way to xml for to create the java classes required for the generator. (or find a way to directly convert, either way not a big problem)

Thanks

3x0dv5:
hey. I am going to pull the PR out, looked at it a second time now after a while and actually doesn't make much sense especially the ScaleDenom to -1 won't work well. 

looking at https://github.com/mapserver/mapserver/blob/d4875f9be5e18c3c25055f9898d0caef590bf112/mapfile.c#L6568 
I realize that setting the scaledenom to -1 is just plain and simple bad. 
if(getDouble(&(map->scaledenom)) == -1) return MS_FAILURE;

PR #5754, opened 2019-02-27
Change DWITH_POINT_Z_M to be ON by default

geographika:
See #5456


szekerest:
@geographika I'd prefer to get rid of WITH_POINT_ZM from the code entirely, especially from mapprimitive.h, because if we leave this optional, then we should make sure the SWIG interface files were also generated with this setting enabled, otherwise the interface will not be aligned with the core mapserver. I think neither of the languages take this into account (at least for the cmake build) which I think may lead to memory corruption issues when WITH_POINT_ZM is enabled.

geographika:
@szekerest - sounds good. It came up as an issue for me when testing the Python MapScript bindings against the builds at http://gisinternals.com/ - by default WITH_POINT_ZM is OFF, but tests could pass and fail depending on what this setting was. 
Looking at the discussions at #5456 it seems it shuold be ok to switch to using POINT_Z_M throughout the codebase. There are ~28 files affected though - https://github.com/mapserver/mapserver/search?q=USE_POINT_Z_M so would be a signficiant change. One for a vote on the mailing list?

szekerest:
yes it would be reasonable to make a motion on the list

geographika:
Link to mailing list post for reference: https://lists.osgeo.org/pipermail/mapserver-dev/2019-February/015610.html

jmckenna:
Merging this, as there were no objections when this was discussed back in 2018, and also in 2019.

jmckenna:
thanks for pushing this @geographika ! 

 (I enabled this for MS4W users by default back in 2018)

PR #5757, opened 2019-03-04
Fix PHP7 build problems with 7.3

ejn:
Fixes some problems encountered building (old-school) PHP-Mapscript with PHP 7.3


mapscript/php/php_mapscript.c
```diff
@@ -898,9 +898,7 @@ PHP_FUNCTION(ms_tokenizeMap)
                                         filename);
     return;
   } else {
-    if (array_init(return_value) == FAILURE) {
-      RETURN_FALSE;
-    }
+    array_init(return_value);
```
geographika: Can this still return a FAILURE? What is the main reason for changing these?

ejn: From PHP 7.3 onwards `array_init()` is no longer a function and no longer returns a value (see [UPGRADING.INTERNALS](https://github.com/php/php-src/blob/PHP-7.3/UPGRADING.INTERNALS#L44)) and so from PHP 7.3 onwards cannot be used in the if-condition.

In any case the return value was not checked in every case in the code where `array_init()` is used, and as far as I can see then at least as far back as PHP 5.6 it was impossible for  `array_init()` to return anything other than `SUCCESS`.

What happens in case of failure is a good question - but I assume that this would be pretty fatal in any case (out of memory? I can't think of any other likely error causes) and so (i) there was never much point checking the return value and (ii) that would be the reason why there is no longer a return value.

I guess it would be possible to shim all the uses of array_init() such that up until 7.2 the return value is checked, but as far as I can see that would be pretty pointless, so it's simpler, cleaner and has no disadvantage to just use it in the "new" style without the return check.

geographika: @ejn - thanks for the thorough explanation.

geographika:
@ejn - above request looks good to me as far as I can tell. I'll merge in a couple of days in case anyone wants to add anything. 
Do you have experience with both the "old" and "new" (SWIG) PHP MapScripts? Are there any features missing in one but not the other? Any details you may have would be good to decide on a future roadmap.

ejn:
@geographika sorry, I've not looked at the PHP-NG MapScript yet at all. Medium-term we'll be looking at moving across to it on the assumption that it is probably (i) more complete and consistent and (ii) likely to be better-maintained in future, but short-term we're moving to PHP7 with the old MapScript rather than try taking two steps at once.

PR #5764, opened 2019-03-09
Fix crash on corrupt qix files (#2362)

szekerest:


rouault:
Looks good to me

jmckenna:
looks good here also

sdlime:
+1

PR #5765, opened 2019-03-13
Add GeoInterface to Python MapScript Classes

geographika:
This pull request adds a `__geo_interface__ `property for PointObj, LineObj, and ShapeObj. 
See https://gist.github.com/sgillies/2217756 for details about the __geo_interface__. 

It allows a GeoJSON-like interface for MapScript objects to help with interoperability with other geospatial Python libraries. 

```python
        res = layer.getResult(0)
        s = layer.getShape(res)
        s.itemdefinitions = l.getItemDefinitions()
        print(s.__geo_interface__)

```

Results in:

```
          {
            'geometry': {
                'type': 'Polygon',
                'coordinates': [[(-0.25, 51.227222), (-0.25, 51.727222), (0.25, 51.727222),
                                 (0.25, 51.227222), (-0.25, 51.227222)]]
            },
            'type': 'Feature',
            'bbox': (-0.25, 51.227222, 0.25, 51.727222),
            'properties': {'FNAME': 'A Polygon', 'FID': 1}
         }
```

A new MapScript method **getItemDefinitions** has been added to the layerObj. 

```
Returns the requested item's field type.
A layer must be open to retrieve the item definition. 

Pass in the attribute index to retrieve the type. The 
layer's numitems property contains the number of items 
available, and the first item is index zero.
```
This allows the type of an item (field) to be returned as a string e.g. Character, Integer, Real (these are all GML data types as listed at https://mapserver.org/ogc/wfs_server.html). 

Types are set on fields in a layer through `gml_[item name]_type` METADATA settings, or using `"gml_types"   "auto"`

These types can then be set as a property on a shapeObj as shown in the example above so that item values are returned in the GeoJSON output as ints, floats, or strings. 

A review of the C code in `mapscript/swiginc/layer.i` would be very much appreciated.

### DIFF
--- a/mapows.h
+++ b/mapows.h
@@ -53,290 +53,310 @@ extern "C" {
  *   mapows.c
  *====================================================================*/
 
-typedef struct {
-  char *pszVersion;
-  char *pszUpdateSequence;
-  char *pszRequest;
-  char *pszService;
-  char *pszTypeName;
-  char *pszFilter;
-  char *pszFilterLanguage;
-  char *pszGeometryName;
-  int nMaxFeatures;
-  char *pszBbox; /* only used with a Get Request */
-  char *pszOutputFormat;
-  char *pszFeatureId;
-  char *pszSrs;
-  char *pszResultType;
-  char *pszPropertyName;
-  int nStartIndex;
-  char *pszAcceptVersions;
-  char *pszSections;
-  char *pszSortBy; /* Not implemented yet */
-  char *pszLanguage; /* Inspire extension */
-  char *pszValueReference; /* For GetValueReference */
-  char *pszStoredQueryId; /* For DescribeStoredQueries */
-  int   countGetFeatureById; /* Number of urn:ogc:def:query:OGC-WFS::GetFeatureById GetFeature requests */
-  int   bHasPostStoredQuery; /* TRUE if a XML GetFeature StoredQuery is present */
-} wfsParamsObj;
-
+#ifndef SWIG
+    typedef struct {
+      char *pszVersion;
+      char *pszUpdateSequence;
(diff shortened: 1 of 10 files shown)


sdlime:
I'll review today...

geographika:
@sdlime - it is mainly the C function at https://github.com/mapserver/mapserver/pull/5765/files#diff-deee1b0c6f077b32fa05657f1799802c that it would be good to review. The Python extension code and tests should be fine. I changed the approach from returning a gmlItemDefinition which involved adding SWIG guards everywhere in the mapows.h and simply return the data type name as a copied string.

sdlime:
> @sdlime - it is mainly the C function at https://github.com/mapserver/mapserver/pull/5765/files#diff-deee1b0c6f077b32fa05657f1799802c that it would be good to review. The Python extension code and tests should be fine. I changed the approach from returning a gmlItemDefinition which involved adding SWIG guards everywhere in the mapows.h and simply return the data type name as a copied string.

Sorry, sorry... Should check item_list is not NULL before accessing it since msGMLItems() can return if it has issues allocating memory.  I think it's fine otherwise.

geographika:
@sdlime - thanks for the review. I've added in the check. I presume if `item_list `is NULL then there is no need to call `msGMLFreeItems(item_list);` and that line can stay in the If block?
If you're happy I can merge so it is included in the 7.4 release.

sdlime:
> @sdlime - thanks for the review. I've added in the check. I presume if `item_list `is NULL then there is no need to call `msGMLFreeItems(item_list);` and that line can stay in the If block?
> If you're happy I can merge so it is included in the 7.4 release.

Looks good to me - merge on!

PR #5778, opened 2019-04-04
add perl to travis

AndyColson:
A start just to see if it'll build and test.  Only one test and one example so far.  If this goes ok, then I'll add more.


geographika:
@AndyColson - looks great so far. You can see build and test output at https://travis-ci.org/mapserver/mapserver/jobs/515505933

sdlime:
Great addition... Should this be completed and merged as part of beta2?

AndyColson:
I have two small swig changes I'd like to make:
1) enable msIO_getAndStripStdoutBufferMimeHeaders for everything. (why is it python only?)
2) clean up compiler warning in image.i (and also return null on error, rather than broken object)

Would that be ok to check in?  If I check it in, and you don't like it, I dunno how to un-check-out.

AndyColson:
Ahh, bummer, my merge seems to have picked up the version change in CMakeLists.txt.  How do I undo that?

geographika:
@AndyColson - is this pull request ready to merge to a new 7.4-beta release?

AndyColson:
Yes.

geographika:
@AndyColson - thanks very much for setting this up

Issue #5781, opened 2019-04-07
Improve the layer extent calculation of the MSSQL driver


szekerest:
We should rely on the server side aggregates instead of specifying a fixed entent

PR #5782, opened 2019-04-07
Improve the layer extent calculation of the MSSQL driver (#5781)

szekerest:


sdlime:
Do you want this in 7.4 instead?

szekerest:
@sdlime Yes, but I'm going to apply a small fix in that. 
Not sure if the other drivers work that way, but if the layer extent is set (ie from mapscript or in the mapfile) the driver should not turn to the database to calculate the layer extent.

PR #5785, opened 2019-04-10
Add sampling along diagonal for rectangle as polygon reprojection

ravhed:
This pull request will solve reprojection issues in certain edge cases. 

### Issue
Not all geometries of the shapefile I was rendering appeared in the final image even though the extent cover all geometries. When I found the issue I had the following setup:
- Shapefile with data in EPSG:32633 projection
- WMS request with bounding box specified in EPSG:4326 projection
  - The following query string was used to debug the issue:
  ```QUERY_STRING=map=fylker.map&FORMAT=image%2Fpng&LAYERS=Fylker_Norge_flate&TRANSPARENT=TRUE&VERSION=1.3.0&EXCEPTIONS=INIMAGE&SERVICE=WMS&REQUEST=GetMap&STYLES=&CRS=EPSG%3A4326&BBOX=-90,-180,112.92325084041,22.92325084041&WIDTH=512&HEIGHT=512```

When the bounding box is reprojected (-90,-180,112.92325084041,22.92325084041) to EPSG:32633 the reprojected bounding box will not be correct, even with the current implementation which takes samples along the edges of the bounding box. To give an example, in the case described above the bottom left corner will have the following coordinates:

`500000.00       -9997964.94`

Using a sample in e.g. center of the bottom line of the rectangle will still give the same X value (500000.00) due to the Y value. 

### Proposed fix
The solution I propose is to sample the diagonal line in addition to the rectangle edges. In my case e.g. center point of the original bounding box would give a lower minimum X values than any of the samples along the edges.

This is how the resulting image looks before the fix:
pre_fix
This is after the fix:
post_fix
In my case I had an issue with the min x value being too high of the reprojected bounding box even though it clearly shouldn't as I specify -180 deg in the request bounding box (which is in EPSG:4326).

### DIFF
--- a/mapproject.c
+++ b/mapproject.c
@@ -806,7 +806,7 @@ msProjectRectAsPolygon(projectionObj *in, projectionObj *out,
   lineObj  ring;
   /*  pointObj ringPoints[NUMBER_OF_SAMPLE_POINTS*4+4]; */
   pointObj *ringPoints;
-  int     ix, iy;
+  int     ix, iy, ixy, sampleDiagonal, numPolyPoints;
 
   double dx, dy;
 
@@ -841,7 +841,8 @@ msProjectRectAsPolygon(projectionObj *in, projectionObj *out,
   }
 
   /* -------------------------------------------------------------------- */
-  /*      Build polygon as steps around the source rectangle.             */
+  /*      Build polygon as steps around the source rectangle              */
+  /*      and possibly its diagonal.                                      */
   /* -------------------------------------------------------------------- */
   dx = (rect->maxx - rect->minx)/NUMBER_OF_SAMPLE_POINTS;
   dy = (rect->maxy - rect->miny)/NUMBER_OF_SAMPLE_POINTS;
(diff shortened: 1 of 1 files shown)


sdlime:
This looks like a nice (and simple) addition. I'd like someone with more projection experience to quickly review... --Steve

mapproject.c
```diff
@@ -857,7 +858,10 @@ msProjectRectAsPolygon(projectionObj *in, projectionObj *out,
     return MS_SUCCESS;
   }
   
-  ringPoints = (pointObj*) calloc(sizeof(pointObj),NUMBER_OF_SAMPLE_POINTS*4+4);
+  /* If there is more than two sample points we will also get samples from the diagonal line */
+  sampleDiagonal = NUMBER_OF_SAMPLE_POINTS > 2 ? MS_TRUE : MS_FALSE;
```
rouault: NUMBER_OF_SAMPLE_POINTS is a constant, so this variable isn't needed, and the below assignment can be simplified too

ravhed: That is true, I've pushed a new commit which removed redundant code.

PR #5790, opened 2019-04-12
allow NDEBUG to work for both set/unset cases

akrherz:
When `-DNDEBUG` is set, the opposite happens with the code checks of `ifndef`.  This change should allow `NDEBUG` to properly function when it is both set and unset.

### DIFF
--- a/mapdrawgdal.c
+++ b/mapdrawgdal.c
@@ -86,7 +86,7 @@ int msDrawRasterLayerGDAL(mapObj *map, layerObj *layer, imageObj *image,
 {
   int i,j, k; /* loop counters */
   int cmap[MAXCOLORS];
-#ifndef NDEBUG
+#if !defined(NDEBUG)
   int cmap_set = FALSE;
 #endif
   unsigned char rb_cmap[4][MAXCOLORS];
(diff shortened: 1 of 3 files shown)


sdlime:
Looks ok to me...

rouault:
This commit doesn't make sense to me. `#ifndef FOO` and `#if !defined(FOO)` are strictly equivalent, or I'm missing something...

akrherz:
Hi @rouault , I can't imagine the day that I am able to explain code to you (!!), but lets try this table

compile flag | desired impact | `ifndef NDEBUG` evals to | `!defined(NDEBUG)` evals to 
-- | -- | -- | --
-DNDEBUG | I don't want DEBUG statements | true | false
(nothing set) | I want DEBUG, maybe | true | true

rouault:
Your table is wrong, at least with gcc and hopefully with all sane compilers. -DNDEBUG *defines* NDEBUG, so #ifndef NDEBUG evaluates to false.
Below an example demonstrating my above statement ( and stack overflow discussion : https://stackoverflow.com/questions/8617712/difference-between-ifndef-and-if-defined-in-c)

```
$ cat ndef.c
#if !defined(FOO)
#error "case 1"
#endif
#ifndef FOO
#error "case2"
#endif
```

```
$ gcc -c ndef.c 
ndef.c:2:2: error: #error "case 1"
 #error "case 1"
  ^
ndef.c:5:2: error: #error "case2"
 #error "case2"
  ^
````
```
$ gcc -c ndef.c -DFOO
[no output]
```

akrherz:
sorry for wasting everyone's time again. Ufff

Issue #5766, opened 2019-03-13
Support for PROJ 6.0.0


sebastic:
Mapserver still uses `proj_api.h` which is deprecated in PROJ 6.0.0 and will be removed in PROJ 7.0.0 (scheduled for March 2020).

When using `proj_api.h` with PROJ 6.0.0 `-DACCEPT_USE_OF_DEPRECATED_PROJ_API_H=1` needs to be added to `CFLAGS` to prevent a compiler error.

Mapserver should be updated to (also) support the `proj.h` API.

Note that PHP MapScript still includes `projects.h` which is no longer available in PROJ 6.0.0.

bmulcahy:
In the mean time a quick work around for anyone needing to use PROJ 6 or anyone who has recently built GDAL with PROJ 6 without realizing mapserver wasn't compatible, not saying that is what I did but just in case someone does.

Add

`#define ACCEPT_USE_OF_DEPRECATED_PROJ_API_H 1`
before line 41 in mapserver/mapproject.h

THIS IS NOT A SOLUTION just a work around.

sdlime:
This should really be targeted for 7.4. @rouault, any feel for how big a deal this is likely to be?

rouault:
> any feel for how big a deal this is likely to be?

I haven't looked in details, but not trivial.
And I'm not sure we can keep PROJ <6 and PROJ >= 6 support at the same time.

sdlime:
Thanks and ugh... I had this as a code sprint activity for 8.0, perhaps it's best left there. Will start reading up.

fxcoudert:
Currently blocking integration of latest proj in Homebrew: https://github.com/Homebrew/homebrew-core/pull/38215

sdlime:
@rouault is the preferred method to set a cflag -DACCEPT_USE_OF_DEPRECATED_PROJ_API_H=1 or the define in mapproject.h as suggested by @bmulcahy? They're equivalent and the latter is trivial (no messing with cmake) so I was going to add that quickly before creating the 7.4 branch. Just making sure...

rouault:
> or the define in mapproject.h

this is fine to me, but I also see that mapwcs20.c directly includes <proj_api.h>. I guess that include should be replaced by mapproject.h (or it is possible that mapproject.h is indirectly included)

sdlime:
[...]

PR #5792, opened 2019-04-15
Remove php_proj.c

rouault:
See https://github.com/mapserver/mapserver/issues/5766#issuecomment-483350415


sdlime:
+1 from me. There's been a rash of lack of attention on my part recently. Thanks for the triage...

PR #5793, opened 2019-04-18
GML:Name is now encoded in WMS GetFeatureInfo (fixes #5361)

### LINKED ISSUE #5361: WMS:GetFeatureInfo, encoding of gml:name
gml output from WMS:GetFeatureInfo is invalid if the wms_title of a layer contains "&". 

_Mapfile:_
```
[snip]
LAYER
  NAME "layer1"
  METADATA
    "wms_title" "this & that"
  END
[snip]
```

_Request:_
`?REQUEST=GetFeatureInfo&INFO_FORMAT=application/vnd.ogc.gml`

_Output:_
```
<?xml version="1.0" encoding="UTF-8"?>

<msGMLOutput 
         xmlns:gml="http://www.opengis.net/gml"
         xmlns:xlink="http://www.w3.org/1999/xlink"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
        <layer1_layer>
        <gml:name>this & that</gml:name>
[snip]

```
The `<gml:name>` element in the result is in plain text: `<gml:name>this & that</gml:name>` but it should be xml encoded `<gml:name>this &amp; that</gml:name>`

Since the output is not valid gml, clients like QGIS will fail to consume the result. In the case of QGIS the end user will see "No features found".

I think the problem is in mapgml.c:1458 where msOWSPrintMetadata() should be changed to msOWSPrintEncodeMetadata()

ravhed:
This will fix #5361.

### DIFF
--- a/mapgml.c
+++ b/mapgml.c
@@ -1455,7 +1455,7 @@ int msGMLWriteQuery(mapObj *map, char *filename, const char *namespaces)
 
       value = (char *) msOWSLookupMetadata(&(lp->metadata), "OM", "title");
       if (value) {
-        msOWSPrintMetadata(stream, &(lp->metadata), namespaces, "title", OWS_NOERR, "\t<gml:name>%s</gml:name>\n", value);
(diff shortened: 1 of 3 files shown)


sdlime:
Thanks @ravhed!

PR #5799, opened 2019-05-01
Fix reference to TEMPPATH in SLD error message

dbaston:
This PR fixes an error message that requests `TMPPATH` be set in the mapfile, when the correct parameter name is `TEMPPATH`.


jmckenna:
thanks @dbaston !

PR #5800, opened 2019-05-06
Fix build issue introduced by #5799

ravhed:
This pull request will fix testing errors introduced by #5799. Since the error message was changed the tests which produces this error needs to be updated


ravhed:
I believe the tests are running as they should now. The Appveyor issue is not related to what I changed from what I can see and seems to be a temporary error:

https://ci.appveyor.com/project/MapServer/mapserver/builds/24325849/job/rgoqyot8l97p653k
`Error uploading artifact the storage: The remote name could not be resolved: 'appveyorcidatav2.blob.core.windows.net'`

geographika:
@ravhed - I restarted the CI and it completed the checks successfully this time. 
Thanks for the update.

PR #5804, opened 2019-05-16
Fix vagrant/Travis builds.

sdlime:
Dealing with changes to UbuntuGIS packaging. If ok, then we should probably should backport to 7.4, 7.2, 7.0 and 6.4 branches... --Steve


rouault:
@sdlime  To fix Travis-CI builds, a similar change must be done in the .travis.yml file

Issue #5815, opened 2019-06-03
WMS layer cause crash if intersecting rectangle is 1px


wicksell:
MapServer will crash if a WMS layer cover only part of the requested extent and the resulting WMS request will be 1 pixel wide or high. 

I have tracked this down to a problem with MS_CELLSIZE which always subtract one from the pixel distance resulting in a division by zero and an invalid cell size. This will later cause MapServer trying to create an image of size -2147483648.

I do not know what the correct fix for this is. Maybe MS_CELLSIZE should be redefined as
`#define MS_CELLSIZE(min,max,d) (((max) - (min))/(MS_MAX((d)-1, 1)))`
or MS_OWS_CELLSIZE should be used in msDrawWMSLayerLow instead of MS_CELLSIZE

ravhed:
I encountered this issue today. I believe the second option (using MS_OWS_CELLSIZE instead of MS_CELLSIZE) is correct since it is OK for MS_CELLSIZE to return a decimal value between 0 to 1 if I'm not mistaken. Can you make a pull request with the fix?

wicksell:
I do not have the environment setup for making pull request since we build using GIS internals packages. We have plans to fix this but it will be in a couple of months.

PR #5821, opened 2019-06-13
WMS Client - Crash when WMS client request is one pixel in width or height

ravhed:
See #5815 for details.


ravhed:
It seems like this change didn't pass one of the WMS client tests. This pull request should not be merged until I've pushed a fix for it.

sdlime:
I suppose we need to fix the regression created by #5707 first...

ravhed:
I've updated the remaining tests now. I believe the appveyor build failure is not related to my changes.

jmckenna:
thanks @ravhed !

PR #5824, opened 2019-06-17
Re-apply "Reduce log clutter from non-errors""

jmckenna:
Reverts mapserver/mapserver#5822


sdlime:
I want to check out just fixing the tests first... hang on.

sdlime:
Sorry Jeff, was out Friday and now see the tests were updated. So you must have reverted previously and this re-applies, correct? If so then carry on...

jmckenna:
Correct.  thanks, re-applying..

PR #5831, opened 2019-07-01
SLD: Add support to arithmetic expressions

jbo-ads:
This pull request aims at adding support to arithmetic expressions in SLD documents read by MapServer. It implements the first section of [MS RFC 124: Improving SLD Support in MapServer](https://mapserver.org/development/rfc/ms-rfc-124.html#literal-constants-variables-and-arithmetic-expressions).


sdlime:
At first glance this looks good. I kinda wish we could just extend the binding object to be more general to handle both a straight item and an expression but this work doesn't preclude that from happening at a later date. --Steve

sdlime:
Do you have a feeling about how easy it would be to add this functionality in a more general sense? I think that should be done even if not part of your work. There are also at least one label property where this would be a nice addition (SIZE).  --Steve

jbo-ads:
Do you mean extending the binding object instead of having both `bindings` and `exprBindings`? I think the changes would affect mainly the MapFile parsing step. As a side effect, expressions would be allowed in MapFile in places where currently only attributes and literals are.

sdlime:
That's what I was thinking but it's probably not worth the effort and you're right that expressions for things like colors don't make sense. Although enabling an expression binding would still need to be done in the mapfile parsing code (my second comment). I can take that piece on...

mapfile.c
```diff
@@ -1675,9 +1675,11 @@ void initLabel(labelObj *label)
   label->styles = NULL;
 
   label->numbindings = 0;
+  label->nexprbindings = 0;
   for(i=0; i<MS_LABEL_BINDING_LENGTH; i++) {
     label->bindings[i].item = NULL;
     label->bindings[i].index = -1;
+    msInitExpression(&(label->exprBindings[i]));
```
rouault: where is this freed ?

mapfile.c
```diff
@@ -2599,9 +2601,11 @@ int initStyle(styleObj *style)
   style->linejoinmaxsize = MS_CJC_DEFAULT_JOIN_MAXSIZE;
 
   style->numbindings = 0;
+  style->nexprbindings = 0;
   for(i=0; i<MS_STYLE_BINDING_LENGTH; i++) {
     style->bindings[i].item = NULL;
     style->bindings[i].index = -1;
+    msInitExpression(&(style->exprBindings[i]));
```
rouault: where is this freed ?

mapfile.c
jbo-ads: Thanks for pointing this out. I added msFreeExpression() in freeLabel() function.

mapfile.c
jbo-ads: Thanks again. I added msFreeExpression() in freeStyle() function.

jbo-ads:
@sdlime As an exercise I am currently investigating how easy it would be to add expressions on label size. It has been a long time since I last played with lex and yacc :)

jbo-ads:
@sdlime I just added expression handling on Label Size. It only involves changes in [mapfile.c](https://github.com/jbo-ads/mapserver/commit/fc07dd9168d06a2fbca3e833d0ec56959c52449b#diff-298174ee3e9d9b9aea01335331b60d0f). It should not be difficult to propagate these changes on other properties. Tell me if you want that I include this commit in this pull request.

sdlime:
> @sdlime I just added expression handling on Label Size. It only involves changes in [mapfile.c](https://github.com/jbo-ads/mapserver/commit/fc07dd9168d06a2fbca3e833d0ec56959c52449b#diff-298174ee3e9d9b9aea01335331b60d0f). It should not be difficult to propagate these changes on other properties. Tell me if you want that I include this commit in this pull request.

I think we should add this capability across the board - I can help too. I think folks would be bummed if this nice addition was SLD-only. That should be a broader goal. I would include that commit in the pull request.

Which got me wondering - how do things behave if an object already has an expression binding and then SLD defines another one? I would guess it would just free the old one and install the new one (essentially overwrite). Or this use case, there's an existing attribute binding and then SLD defines an expression binding? This might be a reason to have a single binding object?



--Steve

jbo-ads:
Commit is included in this pull request. 

I didn't check these specific use cases. They are worth a test in msautotest. An expression binding from SLD overwrites the one in MapFile and takes precedence over a simple binding from Mapfile.

sdlime:
Thinking about it more doesn't SLD just create a bunch of new classes in a layer? (I'm not an SLD user) I mean, if classes are already defined then I'd guess they are removed and new ones created.

jbo-ads:
I just [checked the code](https://github.com/mapserver/mapserver/blob/0b3687b7147330af1d337e992692aeae3d7269ad/mapogcsld.c#L247-L255): if a layer is defined both in MapFile and SLD, all classes from MapFile inside that layer are just removed and replaced by classes from SLD.

sdlime:
I think we should go ahead and merge this code. Then we can work on a separate pull request that exposes the expression binding to more general mapfile processing (using your work in labels as a pattern).  --Steve

geographika:
> 
> 
> @sdlime I just added expression handling on Label Size. It only involves changes in [mapfile.c](https://github.com/jbo-ads/mapserver/commit/fc07dd9168d06a2fbca3e833d0ec56959c52449b#diff-298174ee3e9d9b9aea01335331b60d0f). It should not be difficult to propagate these changes on other properties. Tell me if you want that I include this commit in this pull request.

@jbo-ads - some excellent new features within this pull request. 
One property that would really benefit from being set by expressions is [STYLE>SIZE](https://mapserver.org/mapfile/style.html). There are many cases where symbol size depends on an attribute, but the attribute itself has no relation to a MapServer style size.

jbo-ads:
@geographika - If you don't mind waiting until september, I can add expression handling on STYLE->SIZE just like on LABEL->SIZE.

sdlime:
Could merge now and add style->size as another pull request...

geographika:
> 
> 
> @geographika - If you don't mind waiting until september, I can add expression handling on STYLE->SIZE just like on LABEL->SIZE.

@jbo-ads - whenever you get a chance works for me! As @sdlime says don't let it hold up merging this work.

rouault:
Unless someone objects, I'm considering pressing the 'Merge' button by tomorrow

sdlime:
+1 on merging...

rouault:
I guess we should squash the commits into a single one as the commit history in the PR is a bit messy ?

jbo-ads:
It makes sense. Commit history was useful only for me while implementing the feature.

rouault:
Squashed and merged !

PR #5832, opened 2019-07-01
SLD: Improve WMS GetStyles request

jbo-ads:
This pull request aims at improving WMS GetStyles request, especially when a SLD or SLD_BODY parameter is present. It implements the second section of [MS RFC 124: Improving SLD Support in MapServer](https://mapserver.org/development/rfc/ms-rfc-124.html#wms-getstyles-request).


sdlime:
I don't have any objections here. This is really limited to WMS/SLD and doesn't result in anything that we might want to expose more broadly (unlike the other 2 pull requests). Could anything sensitive be divulged through the XML generation? I can't think of anything beyond attribute bindings or expressions already defined in the mapfile. I mean, what if style SIZE is already referencing an expression (based on the other pull request).

mapogcsld.c
```diff
@@ -3856,14 +3942,38 @@ char *msSLDGenerateTextSLD(classObj *psClass, layerObj *psLayer, int nVersion)
...
+
+  for (lid=0 ; lid < psClass->numlabels ; lid++)
+  {
+    char * psLabelText;
+    psLabelObj = psClass->labels[lid];
+
+    psLabelText = NULL;
+    if (psLabelObj->text.string)
+    {
+      psLabelText = msStrdup(psLabelObj->text.string);
```
rouault: what if psLabelObj->text.string doesn't start with [ and ends with ] ? Should the type of the expression be checked ?

mapogcsld.c
```diff
@@ -3856,14 +3942,38 @@ char *msSLDGenerateTextSLD(classObj *psClass, layerObj *psLayer, int nVersion)
...
+      psLabelText = msReplaceSubstring(psLabelText, "[", "<ogc:PropertyName>");
+      psLabelText = msReplaceSubstring(psLabelText, "]", "</ogc:PropertyName>");
+    }
+    else if (psClass->text.string)
+    {
+      psLabelText = msStrdup(psClass->text.string);
```
rouault: same as above

mapogcsld.c
```diff
@@ -3856,14 +3942,38 @@ char *msSLDGenerateTextSLD(classObj *psClass, layerObj *psLayer, int nVersion)
...
+      psLabelText = msReplaceSubstring(psLabelText, "[", "<ogc:PropertyName>");
+      psLabelText = msReplaceSubstring(psLabelText, "]", "</ogc:PropertyName>");
+    }
+    else if (psLayer->labelitem)
+    {
+      psLabelText = msStrdup(psLayer->labelitem);
+    }
+    if (!psLabelText) continue; // Can't find text content for this <Label>
+
+    psLabelText = msReplaceSubstring(psLabelText, "\"", "");
```
rouault: what is this replacement for ?

mapogcsld.c
```diff
@@ -3856,14 +3942,38 @@ char *msSLDGenerateTextSLD(classObj *psClass, layerObj *psLayer, int nVersion)
...
+    }
+    else if (psClass->text.string)
+    {
+      psLabelText = msStrdup(psClass->text.string);
+      psLabelText = msReplaceSubstring(psLabelText, "[", "<ogc:PropertyName>");
+      psLabelText = msReplaceSubstring(psLabelText, "]", "</ogc:PropertyName>");
+    }
+    else if (psLayer->labelitem)
+    {
+      psLabelText = msStrdup(psLayer->labelitem);
```
rouault: shouldn't that be XML escaped ?

mapogcsld.c
jbo-ads: You are correct. In my attempt to upgrade this part of code to MapServer version 6.2, I missed the whole picture and handled only few simple use cases. I'll rework on this as soon as I can. Up to you to decide whether the merge needs these upcoming changes or if it can be performed as is and changes be applied in another pull request.

mapogcsld.c
jbo-ads: same as above

mapogcsld.c
jbo-ads: same as above

mapogcsld.c
jbo-ads: same as above

jmckenna:
@jbo-ads what is the status of this change?  It seems you mentioned that you will 'rework' this.

jbo-ads:
Actually I am currently on it. Comments from @rouault are trickier than they look like... I need to update mapparser.y in order to properly convert from Mapfile to SLD expressions.

jmckenna:
@jbo-ads ok thanks for the update.  Do you want me to wait for your changes, before beginning the 7.6.0 release process?

jmckenna:
meaning: ok I will wait for your changes, before proceeding (unless you advise otherwise here).

jbo-ads:
Thank you, but I can wait for mid-year's 8.0 release for my changes to be integrated. I think that many people are waiting for a 7.6.

jbo-ads:
@rouault : I updated the code to handle labels of both types MS_STRING and MS_EXPRESSION. I didn't include handling of Mapfile functions though. I don't know what to do with them:
- Should a SLD document returned by a `GetStyles` request be as close as possible to original Mapfile despite not being fully compliant to standard?
- or for instance, should `tostring()` function be converted to `numberFormat()` with tricky conversions from `"%f"`-like formats to `"##.##"`-like formats?

rouault:
@jbo-ads I let you issue a backport PR against branch 7.6 if you wish

Issue #5830, opened 2019-06-26
MapServer PROCESSING “SCALE=a,b” does not stretch raster data


asenradov:
I have a map file that loads GeoTIFF data, which is of type float and in the [0.0-100.0] range. It gets scaled to be between 0-255 with this command: 
PROCESSING "SCALE=0,100"
I want to stretch a smaller subset of the data like [10.0-20.0], to be in the 0-255 range, so I do the following. 
    PROCESSING "SCALE=10,20"

The expected behavior is for data between 10.0-20.0 to be stretched to the 0-255 range (mapping 0 to 10.0 and 255 to 20.0) and than classification to be applied on it.

The observed behavior is data of the range 10.0-20.0 is returned not stretched but as is, with classification applied to the pixel range 10-20. In other words, data outside of the 10-20 range is not displayed and the colors of the data returned is the same as when the PROCESSING "SCALE=0,100" was applied.

For more details visit:
https://gis.stackexchange.com/questions/316651/mapserver-processing-scale-a-b-does-not-stretch-data

jmckenna:
Can you create a small test example, so developers can test your issue: create a zipfile containing: small subset of your image, mapfile with single layer pointing to that subset file, and a 'commands.txt' containing the shp2img commands to replicate (or other commands or urls to replicate).  Please attach the zipfile here to this ticket.  thanks.

jmckenna:
you might also include the shp2img output image files, or screenshots of your subset image in QGIS showing the pixel values, or, correct pixel values.   Anyway, the more you can include in your sample.zip the better, and the easier it is for devs to replicate your issue.  thanks again.

asenradov:
[chlor_test.tif.zip](https://github.com/mapserver/mapserver/files/3330505/chlor_test.tif.zip)
Here are couple links provided by Steve Lime:

Hmm… Certainly seems like the values being returned via [pixel] are the raw values and not the stretched values, and that everything outside the scale range gets the offsite.
 
No processing directive, just your classes:
https://maps1.dnr.state.mn.us/cgi-bin/mapserv72?map=/usr/local/mapserver/apps/test/radov/test.map&mode=map&layer=test
 
Processing “SCALE=10,20” and your classes:
https://maps1.dnr.state.mn.us/cgi-bin/mapserv64?map=/usr/local/mapserver/apps/test/radov/test.map&mode=map&layer=test-10-20
 
We must be missing something obvious. Data is like so:
 
[...]

PR #5834, opened 2019-07-02
Adding processing option to classify 16bit rasters using scaled values instead of raw values.

sdlime:
Initial attempt at addressing issue #5830. Use would be something like:

```
    PROCESSING "SCALE=10,50"
    PROCESSING "SCALE_BUCKETS=256"
    PROCESSING "CLASSIFY_SCALED=TRUE"
```

Only supported with 16bit rasters.

--Steve


mapdrawgdal.c
```diff
@@ -2141,6 +2156,17 @@ msDrawRasterLayerGDAL_16BitClassification(
             (starttime.tv_sec+starttime.tv_usec/1.0e6) );
   }
 
+  // testy...
+  if(1==1) {
```
rouault: to remove

mapdrawgdal.c
```diff
@@ -1998,6 +2001,13 @@ msDrawRasterLayerGDAL_16BitClassification(
     }
   }
 
+  /* -------------------------------------------------------------------- */
+  /*      Fetch the scale classification option.                          */
+  /* -------------------------------------------------------------------- */
+  pszClassifyScaled = CSLFetchNameValue( layer->processing, "CLASSIFY_SCALED" );
+  if( pszClassifyScaled != NULL && EQUAL(pszClassifyScaled,"TRUE") )
```
rouault: Perhaps use
```
if( pszClassifyScaled != NULL )
   bClassifyScale = CSLTestBoolean(pszClassifyScaled);
```
so as to accept TRUE, YES, ON, 1

jmckenna:
This looks ready to merge, any objections?  (to include in 7.6.0)

sdlime:
Looks like I need to resolve @rouault comments. Will do so this evening.

mapdrawgdal.c
sdlime: Whoops, removed...

mapdrawgdal.c
sdlime: Ok, cool - a good pattern to know about.

sdlime:
If things pass definitely will want to squash then merge...

sdlime:
@jmckenna - this was merged.

jmckenna:
thank you @sdlime !

PR #5839, opened 2019-07-24
SLD: Implement SLD "Painters Model" Style Layering

jbo-ads:
This pull request proposes an implementation of SLD "Painters Model" style layering, where all applicable classes are used for rendering a feature instead of only the first one in original MapServer policy. This is described in third section of [MS RFC 124: Improving SLD Support in MapServer](https://mapserver.org/development/rfc/ms-rfc-124.html#style-layering).


sdlime:
The alternative here to achieve similar visual effects would be use multiple layers, correct? I could see this being more broadly popular so we really should expose this functionality via the mapfile/mapscript as well. I think that means exposing the layer RENDERMODE property and then the associated values. I guess that would mean on supporting something like:

  RENDERMODE FIRSTCLASS|ANYCLASS

@geographika might have some opinion in terms of syntax. ANYCLASS is probably more descriptive than painter mode.

geographika:
Would a possible option be to have it as a PROCESSING directive? E.g. 

`PROCESSING "RENDERMODE=FIRSTCLASS|ANYCLASS"`

I think this would then already be supported by MapScript looking at [this test](https://github.com/mapserver/mapserver/blob/master/mapscript/python/tests/cases/layer_test.py#L158), without need for modifications. 

A couple of other options for the option naming (agree `MS_PAINTERS_MODEL` may be too technical:

`FIRSTCLASSMATCH|ALLCLASSMATCH`

Or slightly more literal but longer:

`FIRSTMACTHINGCLASS|ALLMATCHINGCLASSES`

geographika:
With regards to using multiple classes from the same layer for a single feature, it is possible using `GROUP` in the class and then (at least with WMS) passing in `style=group1,group2`.

sdlime:
We don't have a good definition on when to add a full-on attribute vs using a processing directive. I agree that this feels more like a the latter...

jbo-ads:
You read my mind! Exposing layer RENDERMORE property was something I thought of when developing the feature for SLD. I will be more available on september for implementing this. I let you choose the syntax details that best fit existing usage.

jratike80:
I wonder what will happen if CLASSes have STYLEs with OPACITY. What is the result when RGBA 200 0 0 128 is rendered over RGBA 0 200 0 128?

jbo-ads:
@geographika, @sdlime: As suggested, I have added a processing directive:

    PROCESSING "RENDERMODE=FIRST_MATCHING_CLASS|ALL_MATCHING_CLASSES"

jmckenna:
@jbo-ads cool feature, thanks!

Issue #5825, opened 2019-06-21
Missing '$' when parsing ogc:PropertyIsLike ?


pmauduit:
When a regex is generated from a `ogc:propertyIsLike` filter, a '^' is prepended to the search, but we cannot find the line end ('$').

For example, with the following snippet in our SLD:

```xml
<ogc:PropertyIsLike escapeChar="\" matchCase="false" singleChar="?" wildCard="*">
  <ogc:PropertyName>PROGRAMMES</ogc:PropertyName>
  <ogc:Literal>REPHYTOX</ogc:Literal>
</ogc:PropertyIsLike>
```

The corresponding SQL statement is generated:

```sql
REGEXP_LIKE( PROGRAMMES, '^REPHYTOX','i' ) 
```

If a '^' is appended, why isn't there also a '$' added at the end of the pattern ?

One would have expected the following:

```sql
REGEXP_LIKE( PROGRAMMES, '^REPHYTOX$','i' ) 
```

The '^' seems to be added in this code:
https://github.com/mapserver/mapserver/blob/master/mapogcfiltercommon.c#L100-L103

This might be related to this previous work: https://github.com/mapserver/mapserver/pull/5739

geographika:
@rouault - the commit at https://github.com/mapserver/mapserver/commit/eea166a5157d4742b617cd7561c62cd2df5c8f5c which adds the $ character causes an issue in the mssql driver (and possibly others database drivers?). 

The SQL clause generated by the filter is now:

`
WHERE ((LOWER([OwnerName]) LIKE LOWER('%Unknown%$')))
`

Which returns no results due to the $

PR #5841, opened 2019-07-26
MSSQL2008: deal with termination '$' at end of regular expression. Follow-up of eea166a5157d4742b617cd7561c62cd2df5c8f5c (refs #5825)

rouault:
@geographika I believe this should do it, but couldn't test. I'd appreciate if you could report if it solves your issue

We should probably add mssql tests in msautotest. Relevant snippets from GDAL autotest suite: https://github.com/OSGeo/gdal/blob/master/gdal/ci/travis/ubuntu_1604/before_install.sh#L10 and https://github.com/OSGeo/gdal/blob/master/gdal/ci/travis/ubuntu_1604/before_install.sh#L46

### DIFF
--- a/mapmssql2008.c
+++ b/mapmssql2008.c
@@ -2988,6 +2988,8 @@ int process_node(layerObj* layer, expressionObj *filter)
           i++;
           continue;
         }
+        if( c == '$' && c_next == 0 && strtmpl[0] == '^' )
+          break;
 
         if (c == '\\') {
           i++;


geographika:
@rouault - many thanks, that removed the $ and the LIKE queries are working again. I didn't realise it was an issue with the MSSQL driver only. 

Yes the MSSQL driver could do with some tests / testdb. I forgot we could also test on Linux now as well. I'll try and get something running in the next couple of months.

PR #5854, opened 2019-09-05
WIP: Medial Axis Approximation GEOMTRANSFORM

sdlime:
This pull request adds an initial implementation of a medial axis approximation that can be used to generate path-following labels within polygons - so the goal is nice looking labels not a perfect approximation. It uses a Voronoi diagram generated by GEOS and then additional simplification to further isolate a multi-linestring suitable for labeling. The transformation can be referenced in a GEOMTRANSFORM like so:

```
LAYER
    TYPE LINE # source shapes must be polygons
    ...
    GEOMTRANSFORM (medialaxis([shape], -15))
    CLASS
      LABEL
        ...
        ANGLE FOLLOW
      END
    END
END
```
The first argument is a shapeObj and the second (optional, default=0) argument is a tolerance value. A tolerance greater than zero will densify the input shape in order to get a better approximation and would be useful for geometries with sparse vertices. A tolerance less than zero will simplify the input shape. In LAYER context the tolerance is given in layer units while in STYLE context tolerance is given in pixels as the shape has already been transformed. I think in most instances this transformation makes most sense at the LAYER level. A couple of example output images are included below. The first takes a complex lake polygon with a simplification argument. The second is an example of polygons with sparse vertices that need to be densified.

Again, this is an initial implementation and hasn't been tested extensively with a wide variety of polygon inputs. Initial tests are encouraging though. I'll add an animation that shows the transformation steps when I get a little more time.

lake-step5

test

--Steve


sdlime:
Nuts, looks like GEOSVoronoiDiagram_r was introduced in GEOS 3.5 and AppVeyor must be referencing something older. Regardless I'll need to update the pull request with a version check...

jmckenna:
This would be a good one to tackle/finish for the 8.0 release.

rouault:
@sdlime Is there some pending work ? Otherwise I could do final review, polishing, rebase to latest master and add tests

sdlime:
Hi @rouault and @jmckenna... I'm not quite sure what to do with this one - it's still sort of a WIP (I edited the pull request name to reflect that). I think it could be a cool new feature but I'm concerned about performance and that my method of computing the medial axis approximation from the Voronoi diagram is sub-optimal (mostly based on trial and error, literature is sketchy). The goal was skeletonization to facilitate labeling. I suppose if we got it in shape versus main we could solicit feedback in the near term and then make a call relative to 8.0. Any help in making that happen would be most welcome...

--Steve

https://stackoverflow.com/questions/29921826/how-do-i-calculate-the-medial-axis-for-a-2d-vector-shape

jratike80:
You might be interested in having a look at the OpenJUMP skeletonizer plugin as a reference. The zip file contains also Java source code and documentation. The theory and some considerations about exceptional cases are in the Technical Summary part.

 https://sourceforge.net/projects/jump-pilot/files/OpenJUMP_plugins/More%20Plugins/JUMP%20Skeletonizer%20Plugin/

sdlime:
> You might be interested in having a look at the OpenJUMP skeletonizer plugin as a reference. The zip file contains also Java source code and documentation. The theory and some considerations about exceptional cases are in the Technical Summary part.
> 
> https://sourceforge.net/projects/jump-pilot/files/OpenJUMP_plugins/More%20Plugins/JUMP%20Skeletonizer%20Plugin/

Will have a look. I'm curious how the existing code works for folks though. I need to rebase against main and update the pull request. Anyone interesting in trying it out?

--Steve

jmckenna:
@sdlime I'm very interested, I had a response written here last night but didn't press the green button ha, asking for your working test .map file + data ;)

jmckenna:
but will test here now, and make a test mapfile + data....

jmckenna:
@sdlime I have built your branch, but I think it would be good to update to latest 'main' (as I notice that your branch requires 'mapprojhack.c' and is missing 'mapiconv.c').  I was able to build it on Windows, but mapserv seems to fail silently and a map image is not generated when I enable: `GEOMTRANSFORM (medialaxis([shape], -15))`

sdlime:
> @sdlime I have built your branch, but I think it would be good to update to latest 'main' (as I notice that your branch requires 'mapprojhack.c' and is missing 'mapiconv.c'). I was able to build it on Windows, but mapserv seems to fail silently and a map image is not generated when I enable: `GEOMTRANSFORM (medialaxis([shape], -15))`

I think I rebased correctly (I hope). Can you give it a try?

jmckenna:
@sdlime I tried with your latest branch, same result (mapserv fails silently and no image is created).  But maybe it's my test case.  Here is what I'm using: https://gatewaygeomatics.com/dl/ticket-5854-medial-axis.zip

sdlime:
> @sdlime I tried with your latest branch, same result (mapserv fails silently and no image is created). But maybe it's my test case. Here is what I'm using: https://gatewaygeomatics.com/dl/ticket-5854-medial-axis.zip

I'll have a look. What version of GEOS do you have (requires 3.5+)?

jmckenna:
GEOS 3.9.0

jmckenna:
(related: GEOS 3.9.1 has just been released today)

sdlime:
> (related: GEOS 3.9.1 has just been released today)

It's generating a segfault so I can replicate...

jmckenna:
ah, glad, thanks.

sdlime:
> ah, glad, thanks.

Fixed the issue. The functions that operate on the resulting network can go too far and result in a NULL geometry... Your test case runs now. --Steve

jmckenna:
Thanks Steve.  I want to give you the results of my local testing (sorry if some of it is known to you, but I have to explain for the other curious quiet readers here), as I try to get to your similar nice label of "Lower Prior" location in the middle of my main polygon. as this trial-error workflow would be the exact steps a regular user would do, as they try to get a magical angled label in the middle of their big polygon...

### Raw Polygon
ttt-raw

### GEOMTRANSFORM (medialaxis([shape], 1))
mapserv hangs forever, have to cancel shp2img request, no image generated  (actually any tolerance value above zero hangs mapserv)


### GEOMTRANSFORM (medialaxis([shape], 0))
ttt-medialaxis0

### GEOMTRANSFORM (medialaxis([shape], -15))
ttt-medialaxis-15

### GEOMTRANSFORM (medialaxis([shape], -30))
ttt-medialaxis-30

### It's about this time that I realize my Nova Scotia main polygon isn't the best test for this algorithm, so I must settle with..
- GEOMTRANSFORM (medialaxis([shape], 0))
- a small size 5 label (any larger value will not display the label)
- MINFEATURESIZE 50
- disabled the line style
- so that I get at least one angled label inside the main polygon, however not where I want it to be (which is really the "Lower Prior"  location in Steve's examples).
ttt-medialaxis0-minfeaturesize50-size5-nostyle

Let me know if I'm understanding this properly.  I think my examples though will be the exact trial/error of a normal user trying to get an angled label inside their main polygon.

jmckenna:
More feedback: regarding speed, there is a hit but not too bad (3X), such as:
- 0.038s (polygon label)
ttt-raw2
- 0.131s (medialaxis)
ttt-medialaxis0-minfeaturesize50-size5-nostyle

sdlime:
If you go look at the wiki animation I think you're seeing the limitations of the algorithm described in step 5 and the notes section. That will be key. With your polygon, one more trip through the dangling line removal algorithm would be just the ticket. That's the bit that needs attention... You can edit mapgeos.c and edit the segmentRatioThreshold value. It's 0.10 now and you could try dropping to 0.05 or whatever, recompile and try your demo.

Regarding your specific instance observations:


  *   GEOMTRANSFORM (medialaxis([shape], 1)) - Well that's densifying the geometry with a vertex every meter so that's going to take forever, plus that results in a massive Thiessen Polygon computation and subsequent processing. You kinda need this option with very sparse polygons (see my second example in the ticket) but you can hose yourself - so don't do that. We could take the option away and require the user to call generalize or densify functions independently - e.g. GEOMTRANSFORM (medialaxis( generalize([shape], 100) )
  *   I don't exactly understand the differences between 0, -15 and -30. What I think is happening is that your data is way less dense and those values might not be having the affect we hope for. I'll need to do more investigating. You could add another layer that uses the simplify (or generalize?) function independently to see what's going on. I'll do the same.
  *
--Steve

sdlime:
Jeff: So, if I change my algorithm to remove spurs until I can't anymore, you end up with something like the images below. Not a medial access transformation, more of a polygon label line generator. This uses MAXFEATURESIZE AUTO to avoid labeling the small stuff. Certainly not perfect but much better. 

There are possible issues with features with islands where the rings around the islands persist (think of a balloon with a string) and if the ring is long enough relative to the string, the label will get placed around the island. Perhaps an option to ignore inner rings would be useful since often you don't care if a label crosses those - in a lake for example. I think that's the easiest solution. 

Also the resulting line geometry can be pretty complex for labeling, I think the main Nova Scotia label is an example - it would be better if it was a simpler shape. You can run a simplification function (10000m) on the resulting geometry (second image). I could see adding a smoothing function as a another option.

So I'm now thinking:

1. Renaming this function since it's not a medial axis transformation.
2. Removing the **tolerance** argument and instead opting to let a user transform the input to the function as they see fit.
3. Adding an **ignore inner rings** argument.
4. Adding/exposing a **smooth** function to the expression parser, not sure if GEOS has something or not.
5. Adding/exposing a **densify** function to the expression parser, not sure if GEOS has something or not. What's in the code now, either I wrote or it was pilfered from the projection code in MapServer.

ttt
ttt

--Steve

jmckenna:
Nice Steve, I modified mapgeos.c as you said and changed to `segmentRatioThreshold = 0.01` and it gives your first result
ttt-segmentratiothreshold-01

Then as you, I want to simplify or smooth the line (it took me a few tries to figure out the correct syntax in the LAYER to chain a `simplify` with a `medialaxis` call:
  `GEOMTRANSFORM (simplify((medialaxis([shape], 0)), 10000))`
which gives your latest result:
ttt-segmentratiothreshold-01-simplify-10000
Now I'm also wondering if GEOS has a smooth function.

Wow this is so very useful to MapServer users.  You're right the proper use-case is lake polygons, and less verticies.  

I also wonder if there is a way, for your new renamed function, to not require users to first chain it with simplify or smooth, and trigger that in the parameters of this new function instead (as I bet soon users will be coming to this ticket and reading how to combine these here, even though it is actually deep in the docs).

jmckenna:
Also, how will the user set this segmentRatioThreshold ?

jmckenna:
Maybe my comment about combining these chained requests into the new function is odd, and could be ignored.  Anyway that did pop into my head, but now that I think on it more, users should become used to running the resulting shape through multiple of these functions.

sdlime:
> Also, how will the user set this segmentRatioThreshold ?

This goes away. We de-dangle until:

- the resulting geometry is NULL and we revert to the previous version, basically there's just a single resulting line OR
- the process doesn't further reduced the number of lines in the resulting shape (no improvement) - this prevents infinite looping in a case the second image from the top (a rectangle w/two islands)

--Steve

jmckenna:
Steve I tried to pull from upstream the latest changes from main into your branch.  Please check that I didn't mangle your changes in mapgeos.c as there were some other recent changes there.

jmckenna:
> 
> This goes away. We de-dangle until:
> 
>     * the resulting geometry is NULL and we revert to the previous version, basically there's just a single resulting line OR
> 
>     * the process doesn't further reduced the number of lines in the resulting shape (no improvement) - this prevents infinite looping in a case the second image from the top (a rectangle w/two islands)

Thanks, makes sense.  Very interesting (your wiki images helped me understand the de-dangle process).

sdlime:
> Thanks, makes sense. Very interesting (your wiki images helped me understand the de-dangle process).

Pruning is a better term. Really looking for something closer to a Scale Axis Transformation (which is derived from the MAT). Here's a good diagram... https://ars.els-cdn.com/content/image/1-s2.0-S0097849312000684-fx1.jpg

sdlime:
Ok, I refactored things a bit:

- Since this isn't really a medial axis transformation I generalized the function name to "skeletonize".
- The skeletonize function takes 1 optional argument - pruning depth. The default is -1 (for AUTO), 0 does no pruning and positive integers set the maximum number of pruning iterations being done.
- I exposed "densify" as a new expression function that takes a tolerance argument (similar to simplify). This function can be useful for simple geometries.
- I added to new functions "inner" and "outer" that return the corresponding rings from an input polygon as a new shapeObj. The outer function is particularly useful in this context where, for example, you might want ignore islands when labeling a lake.

So, a common use case now looks like:

```
LAYER
    NAME "lakes"
    TYPE LINE # source shapes must be polygons
    ...
    GEOMTRANSFORM (skeletonize(outer([shape])))
    CLASS
      LABEL
        ...
        ANGLE FOLLOW
      END
    END
END
```

--Steve

jmckenna:
Thanks Steve.  For the record, with skeletonize I have to set a depth of 3 to get my previous results, such as: `GEOMTRANSFORM (simplify((skeletonize([shape], 3)), 10000))`

I also played with this locally with a simple great lakes polygon file (I am hesitant to mention this here, as we could be debugging forever each test dataset ha).  But I was able to get from the raw labels:
ttt-lakes

to better labels using skeletonize with depth 1 (to be honest it seems I need a value between 0 and 1, which I realize isn't possible) and some simplifying (and FORCE TRUE labels) through: `GEOMTRANSFORM (simplify((skeletonize([shape], 1)), 0.7))`
ttt-lakes-skeletonize1

jmckenna:
I'm +1 for this now.  (and then can start the related doc updates, once this is merged)

sdlime:
Can you share that file?

jmckenna:
https://gatewaygeomatics.com/dl/ticket-5854-skeletonize.zip

sdlime:
> updated test files: https://gatewaygeomatics.com/dl/ticket-5854-skeletonize.zip

It's actually a good test case - one that I would expect to handle better. You actually do get a better initial skeleton if you densify first but it's clear the pruning algorithm needs work. It's not good enough "as is" IMHO. Rats...

LarsSchy:
I have started to run experiments with this. Will report when I have a couple of examples ready.  I also found a recent research paper about this:  https://arxiv.org/abs/2001.02938 
It is interesting since they focus on speed and they seem to have an idea about how to do the pruning fast.   I have to read it a couple of more times to try and understand all the details.  I dont like the idea to maximize the label size.  Thomas, one of the authors have almost the same content in one chapter in his dissertation from last year. https://www.semanticscholar.org/paper/Improved-algorithms-for-map-rendering-and-route-Mendel/a1f34a58239ff22609897c383640ce765b03a562

sdlime:
Thanks Lars. Definitely needs more work - thanks for taking the time. That's not an article I've run into. The scholarly papers on MAT and SAT I've read are indeed a handful (e.g. https://web.archive.org/web/20110916083816/http://www.balintmiklos.com/scale-axis/scale_axis_transform_socg_2009.pdf<https://web.archive.org/web/20110916083816/http:/www.balintmiklos.com/scale-axis/scale_axis_transform_socg_2009.pdf>). There's a nice discussion on Stack Overflow here: https://stackoverflow.com/questions/29921826/how-do-i-calculate-the-medial-axis-for-a-2d-vector-shape, which references an online SVG demo (http://mat-demo.appspot.com/#!/test-mat). I've been trying to work backwards from the type script to understand how they go from MAT to SAT... --Steve

LarsSchy:
I have created some test data and run some experiments.  I find it very useful for one use case that I have with Norwegian data.
The Swedish data is a little more challenging but the results are quite good.
All my experiments can be found in my GIST:  https://gist.github.com/LarsSchy/0bec238d5cb6dabd75e6ebc792c1de91
I have used the example code from Jeff and expanded it.  Thanks Jeff for the template. You have to run the tests yourself to see all the results.

dmorissette:
Looking at: 
https://gist.githubusercontent.com/LarsSchy/0bec238d5cb6dabd75e6ebc792c1de91/raw/102c9f9e6902d4d00545d6f974601048ce22227d/test_250K_s3.png 

a possible enhancement to get better results would be to filter out (or ignore) the inside rings (lakes) that are smaller than the target label font size before running the algorithm... this should result in longer and smoother label lines in this specific example. Just an idea.

sdlime:
> Looking at:
> https://gist.githubusercontent.com/LarsSchy/0bec238d5cb6dabd75e6ebc792c1de91/raw/102c9f9e6902d4d00545d6f974601048ce22227d/test_250K_s3.png
> 
> a possible enhancement to get better results would be to filter out (or ignore) the inside rings (lakes) that are smaller than the target label font size before running the algorithm... this should result in longer and smoother label lines in this specific example. Just an idea.

Actually I added "inner" and "outer" functions that can be used in a geotransform in concert with skeletonize so you can do _skeletonize(outer([shape]))_.

LarsSchy:
Yes, that would be good with arguments on inner.  I just did tests with outer.  My first thought was to make a smart DATA sql remove all inside rings smaller then a certain size.  But it is even better if it is possible in the c-code within the function.

LarsSchy:
It seems like the zip file was truncated in the GIST.  It should be OK now. I did another pull and push.

sdlime:
> Yes, that would be good with arguments on inner. I just did tests with outer. My first thought was to make a smart DATA sql remove all inside rings smaller then a certain size. But it is even better if it is possible in the c-code within the function.

Would be an argument to outer actually, so you get the outer rings + inner rings that meet a size threshold. The more I think about it we might be consider a single function to do all this ring manipulation. Something like _rings_ that would take 2 min size values, one for outer and one for inner rings. Use could look like:
- All outer rings: rings([shape], -1, 0.0)
- All inner rings: rings([shape], 0.0, -1)
- All outer rings with small inner rings removed: rings([shape], -1, 500)
There may be some cool rendering things that could be done with this apart from skeletonization.
--Steve

dmorissette:
In which units would the argument to `rings([shape], -1, 500)` be?

It probably needs to be in pixels in this case since you want to filter out rings that are small relative to the labels at the current map scale (the size of the ring vs the label depends on map scale so it cannot be in ground units).

However it may be useful to have the same function in ground units in some cases for other purposes, so we probably need to support both.

sdlime:
Depends on the context of the GEOMTRANSFORM, so at the layer level it would be layer units and in the context of a style it would be pixels.

jmckenna:
Today I saw a QGIS-related thread on Voronoi diagrams for labelled polygons, and it reminded me of this MapServer pull request here, which will be a useful improvement.  (the thread pointed to https://bl.ocks.org/veltman/13a7234c4ea073bd7caaa11abb1f7b5d)   My comment here needs no response, I just like the link to our work here.

sdlime:
> Today I saw a QGIS-related thread on Voronoi diagrams for labelled polygons, and it reminded me of this MapServer pull request here, which will be a useful improvement. (the thread pointed to https://bl.ocks.org/veltman/13a7234c4ea073bd7caaa11abb1f7b5d) My comment here needs no response, I just like the link to our work here.

Thanks for this. There may be some hints at a finishing up the implementation here. I've not been able to arrive at a suitable approach/implementation to post-process the clipped Voronoi diagram. The findCenterline() bit is probably the key.

rouault:
From https://blog.ianturton.com/osgeo/2021/09/26/Opensource.html by Ian Turton: "The latest project is looking at adding a label line along the centre of a polygon. In asking for more help on this problem I’ve found that MapServer is trying to add this functionality too (so now it’s a race). In the time this blog post has been languishing part written I’ve managed to finish this up (the answer is to use Dijkstra to find the longest shortest path between nodes of index 1)."

sdlime:
> From https://blog.ianturton.com/osgeo/2021/09/26/Opensource.html by Ian Turton: "The latest project is looking at adding a label line along the centre of a polygon. In asking for more help on this problem I’ve found that MapServer is trying to add this functionality too (so now it’s a race). In the time this blog post has been languishing part written I’ve managed to finish this up (the answer is to use Dijkstra to find the longest shortest path between nodes of index 1)."

Wonder if there is code to leverage... ;-)

ianturton:
> > From https://blog.ianturton.com/osgeo/2021/09/26/Opensource.html by Ian Turton: "The latest project is looking at adding a label line along the centre of a polygon. In asking for more help on this problem I’ve found that MapServer is trying to add this functionality too (so now it’s a race). In the time this blog post has been languishing part written I’ve managed to finish this up (the answer is to use Dijkstra to find the longest shortest path between nodes of index 1)."
> 
> Wonder if there is code to leverage... ;-)

Have a look at https://blog.ianturton.com/geotools/geoserver/2021/09/26/Centrelines-part2.html and for the full PR see https://github.com/geotools/geotools/pull/3644 

Not sure as it will help you much as I off loaded all the hard work to our graph module.

sdlime:
I know what I'll be doing this weekend... Thanks for the tips and the blogpost Ian.

sdlime:
Note that I have picked this up again and have added some simple graph support that works. Just going through the shortest path part... --Steve

sdlime:
Another sample image using the state of california boundary. It's a dense polygon so the layer-level GEOMTRANSFORM looks like so:

```
GEOMTRANSFORM (simplify(skeletonize(simplify([shape],500)),50000))
```

So, simplify the geometry before running the skeletonization and then simplify the resulting line. As an aside, I cannot get the SmoothSIA transform to work at all. No error, the geometry just doesn't render. Anyone had luck with that? It would probably be the preferred method to post-process the skeleton. 

The code that generated these images has been is part of the pull request so now would be a great time to test and provide some feedback.

--Steve

ca

sdlime:
One last note, probably should change _skeletonize_ to _centerline_ - it's a better description of the functionality. Since this work is quite dated I think I'll create a new branch and re-integrate... Still, please test it out! --Steve

jmckenna:
Very nice results now Steve!  Here is my results locally, with:
```
  GEOMTRANSFORM (skeletonize(densify([shape], 0.1)))
```
ttt-lakes-today2


(also, after having to explain 'skeletonize' at the MapServer workshop this week, and this morning at MapServer-Status report at FOSS4G-Asia, I'm much happier with 'centerline', ha, a strong +1)

LarsSchy:
Now I have a result with a lakes labeled:   
(simplify((skeletonize([shape])),50))

Very happy with the result!

test_250K_TT9

Now I have started to play a little with REPEATDISTANCE for the labels.  I wish I could skip the first and last label, because I don't want labels in the bays.  But maybe this is wishing for 2.0 already. :-)

sdlime:
Can someone test with smoothsia? I've had no luck with that but want to make sure it's not just me. --Steve

LarsSchy:
I have tried these cases now:
(smoothsia(skeletonize([shape])))
(smoothsia(skeletonize([shape]), 3, 1, 'angle'))
(smoothsia(skeletonize([shape]), 3, 1, 'all'))

The parsing goes OK and uses some time, but no line or text is drawn.
The same happens with both the 250K data set and the 1M data set.
/Lars

sdlime:
> I have tried these cases now: (smoothsia(skeletonize([shape]))) (smoothsia(skeletonize([shape]), 3, 1, 'angle')) (smoothsia(skeletonize([shape]), 3, 1, 'all'))
> 
> The parsing goes OK and uses some time, but no line or text is drawn. The same happens with both the 250K data set and the 1M data set. /Lars

Same experience. Didn't work for me outside the context of skeletonize either...

jmckenna:
So far I can't get a line drawn with smoothsia with main or branch-7-6

jmckenna:
ah, yikes, I finally got smoothsia to work with branch-7-6, by setting "all" instead of "angle" :
````
GEOMTRANSFORM (smoothsia([shape], 3, 1, 'all'))
```
results look good

sdlime:
medial-axis branch is dated and needs to be re-based. I'll do that tonight.

sdlime:
Closing - see https://github.com/MapServer/MapServer/pull/6417 instead.

Issue #5558, opened 2018-03-07
Missing LABEL TEXT attribute bind causes crash in shp2img


geographika:
On MapServer 7.0.4, Windows x64 I am getting the following error / crash when using more than one iteration of shp2img, and referring to a field name that does not exist for LABEL text:

```
Unhandled exception thrown: write access violation.
msyystring_buffer was nullptr.
```

This occurs in  mapfile.c - line 6751:

```
  msyystate = MS_TOKENIZE_FILE;
  msyylex(); /* sets things up, but doesn't process any tokens */

  msyyrestart(msyyin);  // crash here
```

Running one iteration of shp2img works:

`shp2img -m test.map -c 1 -o test.png`

Running two iterations causes a crash:

`shp2img -m test.map -c 2 -o test.png`

Test Mapfile below.

```
MAP    
    EXTENT 0 0 100 100
    SIZE 100 100
    LAYER
        TYPE POINT
        STATUS ON
        FEATURE
            POINTS
                50 50
            END
        END
        CLASS
            LABEL
                TEXT "[Missing]"
            END
        END
    END
END
```

geographika:
Thanks @rouault

PR #5855, opened 2019-09-06
Fix memory leaks and run a Travis-CI build under address sanitizer

rouault:
This PR fixes all memory leaks found when running msautotest (at least the parts that involve spawning the mapserv and shp2img binaries. mapscript testing isn't done currently) under address sanitizer (-fsanitize=address), which is somewhat equivalent to Valgrind (detects invalid uses of dynamic memory and memory leaks) but runs much faster. 
This also modifies the php 7.2 target to run under address sanitizer, and fail if a memory error is found, so that new memory errors aren't added.


geographika:
Possibly related/fixed by this? #5558

rouault:
Unless someone needs more time to review this, I'll merge this in the following days

sdlime:
Wow, that was a lot of work! Looks good... (first change in mapoutput.c is just adding spaces but isn’t worth removing)  +1

rouault:
> first change in mapoutput.c is just adding spaces but isn’t worth removing

Good catch. I could say this was voluntarily done to test reviewer eyes, but actually was accidental. Just removed it :-)

sdlime:
It’s a good strategy – I only brought it up as evidence I actually looked at the changes… ;-)

PR #5856, opened 2019-09-08
Remove MakeValid() from MSSQL Driver

geographika:
These calls as disable spatial indexes causing a performance hit. 
See post in [mailing list](https://lists.osgeo.org/pipermail/mapserver-users/2019-September/081380.html).


szekerest:
We may also backport this to the 7-4 branch.

PR #5863, opened 2019-09-09
Fix memory leak in msINLINELayerNextShape() and null pointer dereference in maplexer when rerunning it after msCleanup() (fixes #5558)

### LINKED ISSUE #5558: Missing LABEL TEXT attribute bind causes crash in shp2img

rouault:


sdlime:
Looks good at first glance... Back port to 7.4 as well I assume? --Steve

PR #5868, opened 2019-09-12
Backport of memory leak fixes

rouault:
@sdlime This PR is a backport of my recent memory leak fixes from master to 7.4. I cannot completely exclude the risk of new double-free in code paths not tested, but as far as msautotest is concerned, it is leak-free and "double-free"-free... Do you think we should merge this in 7.4 ?


sdlime:
I guess I'd say yes. I already cut the 7.4.2 release, I wonder how easy it is to redo that to get these changes in - do we just remove the release tag? Is the AppVeyor issue unrelated?

sdlime:
I figure we're not going to be "more sure" at any point so why not now. I re-cut the release (thanks for the commands) so this is included. Thanks! --Steve

PR #5870, opened 2019-09-13
[WIP] Generated Mapscript API Documentation

geographika:
I've made a start on generating new MapScript API docs using SWIG and Sphinx. I've completed documenting `CLASS`. I have some sample output at: http://www.geographika.net/mapserver-mapscript/mapscript/classes/class.html

All SWIG generated languages should be able to follow the new API docs, with individual language extensions to be documented separately. 

The mapscript module can be installed as part of the docs `requirements.txt` file. It does not need MapServer installed to generate the docs - mocks are used to avoid import errors, and Sphinx just needs to access the docstrings in the mapscript Python module. 

## New Features

* All mapscript functions will be automatically listed even if a docstring has not been added (generating the docs shows what is not documented as they get a giant mock object docstring)
* Links to relevant Mapfile documentation for properties e.g. classObj.title links to the TITLE section on the CLASS Mapfile doc page
* Function parameters automatically documented
* Can create links to API keywords using tags such as :attr:`classObj.metadata` (see [Domains](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html) page
* Can link to other classes using :py:class:`labelLeaderObj`
* Linking to the relevant Mapfile keyword avoids duplication of documentation (although the links have to use a slight hack as index entries have auto-generated URLs - see [this question](https://stackoverflow.com/questions/2479503/sphinx-cross-reference-indexed-item))

To get everything working involved a few hacks to get round missing functionality in SWIG and Sphinx and its extensions. I've tried to keep the approach as clean as possible, and posted a few questions to the SWIG user groups to check I'm not missing something. 

Some notes:

* I made `numlabels` and `leader` immutable in mapserver.h - confirmation this is ok would be appreciated
* Class attributes need to be manually listed to get output in Sphinx. Module constants will also likely need to be manually added. 
* Docstrings have to be added to mapserv.h near the struct declarations (rather than in .i SWIG files) to make sure attribute doc strings are inserted first in the output SWIG class, otherwise it is not possible to document them with autodoc. To keep clutter out of mapserver.h I've defined macros. 
* A few other workarounds are needed to avoid long parameter names such as `mapscript.mapscript.layerObj` appearing everywhere
* Whilst going through the `classObj` it was apparent several attributes were missing from the `CLASS` Mapfile page. I've added these to the main docs. 
* As expected the `classObj` SWIG docs had a number of missing methods and attributes which have been added (along with an updated Python test). 

If the above approach is ok and not too disruptive I will continue with documenting more MapScript classes.


sdlime:
This looks *really* nice. Regarding the properties you made immutable:

1. Numlabels  - or any similar counters - should be immutable. So following that logic numstyles should also be immutable.
2. Leader  - no so sure about that one. Looks like styles and labels aren't even exposed to MapScript and users *must* work through the helper methods. Makes sense since they are arrays. There's only one leader though so how would a user modify it?

--Steve

geographika:
@sdlime - thanks for the feedback. 
Yes `numlabels` was an easy fix - it had a `TODO make immutable` comment next to it in the header file. 
Currently there is no leaderObj in MapScript. I think it is best to add an issue for this and come back to it later or the docs of the current state of MapScript will never be completed. 

I've finished the colorObj now and will add to the pull request. 
Do you think it is best to squash and merge these now, or have a single (giant) merge with all autodoc changes? I'm in favour of the former as it should be easier to review / comment on changes.

geographika:
Closing in favour of smaller more specific commits

PR #5873, opened 2019-09-13
Fix MSVC compiler warnings

rouault:


rouault:
Doesn't compile due to missing cairo-win32.h header. See https://github.com/gisinternals/buildsystem/issues/144

szekerest:
@rouault The build kit is now fixed

PR #5887, opened 2019-10-01
Wrong OFFSET in query for CONNECTIONTYPE OGR

gerdos82:
There is an issue in the OGR driver regarding the OFFSET used in the query. It is based on the startindex parameter in the WFS request. In [this code](https://github.com/mapserver/mapserver/blob/e39cf5dd20c599fe8f87c24e2abe921902fa4a4d/mapwfs.c#L2889) the startindex is increased with one. In the postgis driver this is being compensated by doing a -1 on the startindex before using it as the OFFSET in the query. For the OGR driver this is omitted.


rouault:
It would be great if we could have a msautotest test case showing the effect of this fix

rouault:
Superseded by https://github.com/mapserver/mapserver/pull/5913 that adds a test case

PR #5888, opened 2019-10-01
Port to PROJ 6 API

rouault:


geographika:
Hi @rouault - I see there are some FastCGI related changes as part of this. 

I made a new build from master and I'm getting HTTP 500 errors on IIS. The only information I can get from the failed request tracing is `The system cannot open the file.  (0x4)`. 
The error happens even with an empty request (no Mapfile or data involved) - it should return _No query information to decode. QUERY_STRING is set, but empty._
I've checked a few things with process explorer but there are no obvious errors. 
If I test with a Mapfile it successfully queries a database but won't return any data via IIS. 

From the command line, and also when using the CgiModule rather than FastCgiModule in IIS everything works fine. 

I also downloaded a master build from https://www.gisinternals.com/development.php and it has the same issue when running under FastCGI. 

Any ideas on what may be the cause? I can take to the mailing list if more appropriate.

rouault:
> I see there are some FastCGI related changes as part of this.

In the code you mean ? In theory no. mapserv.c which is where FCGI specific code is hasn't been touched. Commit https://github.com/mapserver/mapserver/pull/5888/commits/f595e91f1b418db72b806162dba4470109ac8dc1 is indeed supposed to help for FCGI code or Python WSGI + MapServer. That said, I'm not sure I tested FCGI after my changes

rouault:
OK, I've just test mapserver master against a PROJ 5 and PROJ 6 builds, with Apache and FastCGI under Linux, and it works fine...

geographika:
@rouault - thanks for the update. I've looked through some other recent commits and nothing stands out. Simply switching from FastCgi to Cgi and everything works, so it doesn't appear to be a dependency issue. I'll try compiling with USE_PROJ set to false to see if that works.

rouault:
you could also try `git bisect` to find the offending commit

Issue #5889, opened 2019-10-04
Allow string padding in templates


geographika:
Items in a template allow for a number of formatting options as per: https://mapserver.org/mapfile/template.html#query

For example: `[item name="GeomType" format="$value" uc="true"]`

A useful addition for creating table-like text output would be a padding option to ensure a fixed width e.g.

`[item name="GeomType" format="$value" uc="true" padding="10"]`

Item values such as "LINE" would then become a padded "LINE.........." (using dots to represent spaces as they are removed by GitHub).

geographika:
Closed with #5890

PR #5890, opened 2019-10-04
Add a new padding template format option

geographika:
See #5889 - adds a new "padding" format option for fixed length strings in a template. 

Online testing in REPL at https://repl.it/repls/LeftSlimQuotient
Reviews of C code welcome, especially if variable declarations should be moved, and everything is created and freed correctly.

### DIFF
--- a/maptemplate.c
+++ b/maptemplate.c
@@ -1320,8 +1320,8 @@ static int processItemTag(layerObj *layer, char **line, shapeObj *shape)
           char *paddedValue = NULL;
           paddedValue = (char *) msSmallMalloc(paddedSize);
           snprintf(paddedValue, paddedSize, "%-*s", padding, tagValue);
-          tagValue = msStrdup(paddedValue);
-          msFree(paddedValue);
+          msFree(tagValue);
+          tagValue = paddedValue;
       }
 
       if(!tagValue) {
--- a/maptemplate.c
+++ b/maptemplate.c
@@ -1207,8 +1207,9 @@ static int processItemTag(layerObj *layer, char **line, shapeObj *shape)
   const char *argValue=NULL;
 
   const char *name=NULL, *pattern=NULL;
-  const char *format=NULL, *nullFormat=NULL;
+  const char *format=NULL, *nullFormat=NULL, *paddingFormatString = NULL;
   int precision;
+  int padding;
   int uc, lc, commify;
   int escape;
 
@@ -1223,8 +1224,10 @@ static int processItemTag(layerObj *layer, char **line, shapeObj *shape)
 
   while (tagStart) {
     format = "$value"; /* initialize the tag arguments */
+    paddingFormatString = "%-*s";
     nullFormat = "";
-    precision=-1;
+    precision = -1;
+    padding = -1;
     name = pattern = NULL;
     uc = lc = commify = MS_FALSE;
     escape=ESCAPE_HTML;
@@ -1240,6 +1243,9 @@ static int processItemTag(layerObj *layer, char **line, shapeObj *shape)
 
       argValue = msLookupHashTable(tagArgs, "precision");
       if(argValue) precision = atoi(argValue);
+
+      argValue = msLookupHashTable(tagArgs, "padding");
+      if (argValue) padding = atoi(argValue);
 
       argValue = msLookupHashTable(tagArgs, "format");
       if(argValue) format = argValue;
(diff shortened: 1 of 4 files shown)


maptemplate.c
```diff
@@ -1310,6 +1316,15 @@ static int processItemTag(layerObj *layer, char **line, shapeObj *shape)
       tagValue = msReplaceSubstring(tagValue, "$value", itemValue);
       msFree(itemValue);
 
+      if (padding != -1) {
+          int paddedSize = sizeof(tagValue) + (padding * sizeof(char*));
+          char *paddedValue = NULL;
+          paddedValue = (char *) msSmallMalloc(paddedSize + 1);
+          snprintf(paddedValue, paddedSize, paddingFormatString, padding, tagValue);
```
rouault: for clarity, remove paddingFormatString and replace it by its inline value paddingFormatString

maptemplate.c
```diff
@@ -1310,6 +1316,15 @@ static int processItemTag(layerObj *layer, char **line, shapeObj *shape)
       tagValue = msReplaceSubstring(tagValue, "$value", itemValue);
       msFree(itemValue);
 
+      if (padding != -1) {
+          int paddedSize = sizeof(tagValue) + (padding * sizeof(char*));
```
rouault: several errors:
- sizeof(tagValue) is sizeof(char*) = 8 on a 64 bit host. You probably meant strlen(tagValue)
- sizeof(char*) should likely be sizeof(char) (or can be just omitted since it is guaranteed that sizeof(char)==1)
- once the above is fixed, you probably lack a +1 for the final zero terminator (for snprintf() purposes), and thus remove the +1 in the msSmallMalloc()

maptemplate.c
```diff
@@ -1310,6 +1316,15 @@ static int processItemTag(layerObj *layer, char **line, shapeObj *shape)
...
+          char *paddedValue = NULL;
+          paddedValue = (char *) msSmallMalloc(paddedSize + 1);
+          snprintf(paddedValue, paddedSize, paddingFormatString, padding, tagValue);
+          tagValue = msStrdup(paddedValue);
```
rouault: you lack a msFree(tagValue) before.
And you could then just do tagValue = addedValue; to avoid the extra dance of msStrdup() / msFree()

geographika: The following:
```
          tagValue = msStrdup(paddedValue);
          msFree(paddedValue);
```
should become?
```
          msFree(tagValue);
          tagValue = paddedValue;
```

geographika:
@rouault - many thanks for the review and explanations. I'll try and find an autotest for templates to add this to (also I'll update docs if this pull request is accepted).

maptemplate.c
rouault: yes

maptemplate.c
```diff
@@ -1241,6 +1243,9 @@ static int processItemTag(layerObj *layer, char **line, shapeObj *shape)
       argValue = msLookupHashTable(tagArgs, "precision");
       if(argValue) precision = atoi(argValue);
 
+      argValue = msLookupHashTable(tagArgs, "padding");
+      if (argValue) padding = atoi(argValue);
```
rouault: this is a bit nitpicking, but there's a potential security implication if the padding value is negative or too large, so some clamping wouldn't hurt here

maptemplate.c
geographika: Updated

maptemplate.c
geographika: I've added the following guard. 

`padding > 0 && padding < 1000`

Not sure what to set as an arbitrary max value. I can add to the docs. Padding should always be greater than 0 though. Should an error be thrown or just ignore the value?

As far as I understand users can't add their own templates so any security implications would imply someone had access to the server already?

sdlime:
I think I’d just ignore the value. Is this for only padding on the right? What about supporting padding on the left or both? A negative number would be convenient for left padding, no idea on both. You’d almost need a second attribute, padding direction… Then again I don’t understand the use case very well.

maptemplate.c
rouault: > As far as I understand users can't add their own templates so any security implications would imply someone had access to the server already?

yes, security is a big word. Was mostly to avoid a crash if someone would put a crazy value. Any behavior other than crashing is OK to me.

geographika:
@sdlime - the main use case is to allow for creation of RST tables which rely on fixed width strings. I've no current use case for left padding though. 
I've added docs and the RST table example to https://github.com/mapserver/docs/pull/286
@rouault - thanks for taking the time to explain the errors. I think this is my first non-Python MapServer "feature" :)

rouault:
> @rouault - thanks for taking the time to explain the errors. I think this is my first non-Python MapServer "feature" :)

Champaign !

Issue #5895, opened 2019-10-13
VALIDATION default_ is case sensitive


geographika:
Currently the default values used in VALIDATION blocks require "default_" to be in lowercase. E.g. the following block will work as expected:

```
VALIDATION
    "default_VAL1" "A"
    "VAL1" "^[a-zA-Z\-]+$"
END
```

The following will not:

```
VALIDATION
    "DEFAULT_VAL1" "A"
    "VAL1" "^[a-zA-Z\-]+$"
END
```
The validation keys themselves are case-insensitive (at lease URL params - see #4516) and there is no mention of case in the [docs](https://mapserver.org/mapfile/validation.html#default-values-if-not-provided-in-the-url). 
It is a good way to lose half an hour of debugging..

I propose that the "default_" check should be case-insensitive.

geographika:
Fixed with #5896

PR #5896, opened 2019-10-13
Make default_ in VALIDATION block case insensitive

geographika:
See #5895

* Pull request changes `strncmp` to `strncasecmp` in the relevant checks of "default_"
* Adds a new test and results using DEFAULT_
* Fixes a couple of typos in a string function comment

Can be squashed before merging.

### DIFF
--- a/mapstring.c
+++ b/mapstring.c
@@ -1503,7 +1503,7 @@ char *msCommifyString(char *str)
 /* ------------------------------------------------------------------------------- */
 /*       Replace all occurrences of old with new in str.                           */
 /*       It is assumed that str was dynamically created using malloc.              */
-/*       Same function as msReplaceSubstring but this is case ins                  */
+/*       Same function as msReplaceSubstring but this is case insensitive          */
 /* ------------------------------------------------------------------------------- */
 char *msCaseReplaceSubstring(char *str, const char *old, const char *new)
 {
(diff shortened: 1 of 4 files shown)


mapstring.c
```diff
@@ -1501,9 +1501,9 @@ char *msCommifyString(char *str)
 
 
 /* ------------------------------------------------------------------------------- */
-/*       Replace all occurances of old with new in str.                            */
+/*       Replace all occurrences of old with new in str.                           */
 /*       It is assumed that str was dynamically created using malloc.              */
-/*       Same function as msReplaceSubstring but this is case incensitive                        */
+/*       Same function as msReplaceSubstring but this is case ins                  */
```
rouault: 'ins' is truncated ?

geographika: Thanks - now fixed

rouault:
@sdlime I've no special opinion on this. Does that look good to you ?

sdlime:
I think this looks good. Potential issue is one where folks are taking advantage of case sensitivity although I would think that would be kinda risky. Probably something to note in upgrade notes.

sdlime:
@rouault or @geographika, is there an easy way to re-run the Appveyor check since this has been sitting here a month (I don't have the appveyor creditials handy)? Just in case...

rouault:
Closing & reopening like I just did is one way :-)

PR #5897, opened 2019-10-14
Add additional EXPRESSION msautotests

geographika:
This pull request adds some more complex EXPRESSION tests to the msautotest suite. Focus is on combined expressions (AND, OR etc.) using the various supported syntaxes, several of which are not covered by tests. 

Test results are CSV exports so easy to review and add new tests on different OSs.


jbo-ads:
It looks like `msautotest/query/expected/expressions_test010.txt` is missing from git repo.

geographika:
@jbo-ads - thanks for spotting this. Fixed with #6000

PR #5907, opened 2019-10-24
Add support of TrueType symbols in UTFGrid

charles-plante:


charles-plante:
Amended the commit to skip treatment instead of fail if symbol not found

dmorissette:
As discussed offline, we need tests in msautotest to test utfgrid behavior vs TTF symbols *and* labels (via labelcache vs labels with force=true) and to confirm the fix. Closing for now.

PR #5911, opened 2019-10-30
Added check for Contains to force a Join on the rtree table for large…

DennisRutjes:
… Spatiallite/Geopackage DB's and improves the lookup performance significantly.


jratike80:
Could you please give a use case when having a check for both Intersects and Contains gives significant speedup.

PR #5918, opened 2019-11-11
5901 utf grid support for true type symbols

charles-plante:
Add support of glyph symbol in UTFGrid with a test


dmorissette:
Hi @rouault. When we look on Travis the tests seem to have passed but this PR does not seem to have received the update. Any idea?  Could this be related to the demo server problem from yesterday? Is there a way to force a new CI test?

rouault:
> Any idea?

I restarted the Travis-CI jobs after @jmckenna fixed the demo server issue. I assume this is just a bug in the connection between GitHub and Travis-CI. You could potentially "git commit --amend && git push --force" to force a new build

msautotest/renderers/utfgrid_truetype_symbol.map
```diff
@@ -0,0 +1,39 @@
...
+    END
+  END
+  FEATURE
+    POINTS
+        50 50
+    END
+    text 'label_sample' END
+  END
+END
+
```
dmorissette: Since there are two possible code paths for the label rendering (via labelcache and directly in marker) and the goal of this PR is to handle both, I'd recommend one test case for each code path even if they both return the same result.
```suggestion
# RUN_PARMS: utfgrid_truetype_symbol_labelcache.json [SHP2IMG] -m [MAPFILE] -l symbol-labelcache -i json -o [RESULT]
# RUN_PARMS: utfgrid_truetype_symbol_nolabelcache.json [SHP2IMG] -m [MAPFILE] -l symbol-nolabelcache -i json -o [RESULT]
#
#
MAP

STATUS ON
EXTENT 0 0 400 300
SIZE 200 150
FONTSET "../misc/fonts.lst"
IMAGETYPE png24
symbolset "symbolset"

# This layer renders the symbol to the utfgrid output but not the label
# (in this case the label is handled via the labelcache code path)
LAYER
  NAME "symbol-labelcache"
  TYPE point
  STATUS OFF
  LABELCACHE ON
  CLASS
    NAME "test1"
    STYLE
        SYMBOL "ttf"
        COLOR 0 0 0
        SIZE 15
    END
    LABEL
	 COLOR  12 0 150
	 OUTLINECOLOR 0 0 255
	 TYPE truetype
	 SIZE 6
	 POSITION AUTO
    END
  END
  FEATURE
    POINTS
        50 50
    END
    TEXT 'label_sample'
  END
END

# This layer renders the symbol to the utfgrid output but not the label
# (in this case the label is handled via the msDrawLabel() code path)
LAYER
  NAME "symbol-nolabelcache"
  TYPE point
  STATUS OFF
  LABELCACHE OFF
  CLASS
    NAME "test1"
    STYLE
        SYMBOL "ttf"
        COLOR 0 0 0
        SIZE 15
    END
    LABEL
	 COLOR  12 0 150
	 OUTLINECOLOR 0 0 255
	 TYPE truetype
	 SIZE 6
	 POSITION AUTO
    END
  END
  FEATURE
    POINTS
        50 50
    END
    TEXT 'label_sample'
  END
END

END #MAP

```

Issue #5744, opened 2019-02-18
RFC 118 & Geotransform


pelord:
Here my class for a linestring stored in postgis as a multilinestring.

```
CLASSITEM "state"
CLASS
	NAME "bad"
	EXPRESSION 'bad'
	STYLE
		LINECAP butt
		WIDTH 8
		COLOR 255 165 0
	END
	STYLE
		GEOMTRANSFORM "end"
		SYMBOL "vertline"
		COLOR 0 0 0
		SIZE 13
		WIDTH 1
		ANGLE AUTO
	END
	STYLE
		GEOMTRANSFORM "start"
		SYMBOL "vertline"
		COLOR 0 0 0
		SIZE 13
		WIDTH 1
		ANGLE AUTO
	END
END
CLASS
...
END 
```

If I don't apply a wms filter, everything is rendered (getmap) as in the class's definition. 
![image](https://user-images.githubusercontent.com/7397743/52964145-a1c5ec00-336f-11e9-9756-cb843e9aec85.png)

If a apply a filter (rfc 118):

`<Filter xmlns="http://www.opengis.net/ogc"><PropertyIsEqualTo matchCase="true"><PropertyName>state</PropertyName><Literal>bad</Literal></PropertyIsEqualTo>
`
Here the result.

![image](https://user-images.githubusercontent.com/7397743/52964183-b7d3ac80-336f-11e9-974f-109c41a5f1cd.png)

A query on wms with geotransform end/start seem to render a densified version of the linestring. 

Here  the vertices of this line:
![image](https://user-images.githubusercontent.com/7397743/52964104-83f88700-336f-11e9-8c08-d12d6d8cedae.png)

I've tried to convert the multilinestring to a linestring  as done on [gis.stackexchange](https://gis.stackexchange.com/questions/19864/converting-multilinestring-to-linestring-with-postgis) with the same result.

Is it an issue or a bad mapfile definition?

 MapServer version 7.2.1

pelord:
The same result with GEOMTRANSFORM "vertices".

sdlime:
I think it's a bug, likely in the query map code. What happens if you add:

    QUERYMAP
      STYLE NORMAL
    END

to your mapfile?

pelord:
Same result.

sdlime:
[...]

PR #5920, opened 2019-11-12
Bring msDrawQueryLayer() in line with msDrawVectorLayer() - addresses…

sdlime:
… #5744.


sdlime:
@geographika, can you try testing again? I pushed the WMS filter execution below style handling so  that the right class is referenced within the query code. This should address your specific issue.

geographika:
@sdlime - excellent! The style is now applied along with the filtering.

mapdraw.c
```diff
@@ -1428,19 +1432,28 @@ int msDrawQueryLayer(mapObj *map, layerObj *layer, imageObj *image)
...
 
     if(cache) {
-      drawmode |= MS_DRAWMODE_SINGLESTYLE;
-      status = msDrawShape(map, layer, &shape, image, 0, drawmode); /* draw only the first style */
-    }
-    else
+      styleObj *pStyle = layer->class[shape.classindex]->styles[0];
+      if (pStyle->outlinewidth > 0) msOutlineRenderingPrepareStyle(pStyle, map, layer, image);
+      status = msDrawShape(map, layer, &shape, image, 0, drawmode|MS_DRAWMODE_SINGLESTYLE); /* draw only the first style */
+      if (pStyle->outlinewidth > 0) msOutlineRenderingRestoreStyle(pStyle, map, layer, image);
```
geographika: Is something meant to happen here?

mapwms.c
```diff
@@ -1910,6 +1887,22 @@ this request. Check wms/ows_enable_request settings.",
...
+  if (filter) {
+    if (sld_url || sld_body) {
+      msSetError(MS_WMSERR,
+                 "Vendor-specific FILTER parameter cannot be used with SLD or SLD_BODY.",
+                 "msWMSLoadGetMapParams()");
+      return msWMSException(map, nVersion, NULL, wms_exception_format);
+    }
+
+    if (msWMSApplyFilter(map, nVersion, filter, need_axis_swap, wms_exception_format) == MS_FAILURE) {
+      return MS_FAILURE;/* msWMSException(map, nVersion, "InvalidFilterRequest"); */
```
geographika: Is there a reason not to throw a msWMSException (I see this is from the original function).

mapdraw.c
sdlime: I can't tell specifically what here means? The code follows the same pattern as msDrawVectorLayer() in terms of drawing when caching is required. We draw the first/bottom style (manipulating the style in cases where outline width is set. It looks right to me.

mapwms.c
sdlime: This is probably a better question for Dan. My solution was simply to move the query/filter code below style-handling block.

sdlime: That's probably a question for Dan.

mapdraw.c
geographika: Sorry - I was looking at it in the browser and it looked like an empty if statement without the brackets.

mapwms.c
geographika: If it is a simple MS_FAILURE then I thins servers just return a HTTP 500 error which are hard to debug compared to an error message? Maybe one for @dmorissette but not a blocker.

sdlime:
@rouault, this is the first time I've tried using the backport function. I thought all that was needed was to set the appropriate label and then merge. Is that not true?

PR #5924, opened 2019-11-19
WMS Client - Fixes GetLegendGraphic in WMS 1.3.0

ravhed:
SLD_VERSION parameter is mandatory in GetLegendGraphic as specified by SLD version 1.1.0. This commit adds this parameter when requesting WMS services with version 1.3.0 or higher.

In addition to this it also fixes style parameter: in GetLegendGraphic it should be named STYLE and not STYLES.

Two new testcases has been added for the WMS client which will test GetLegendGraphic functionality for both WMS 1.1.1 and WMS 1.3.0.

Note: SLD_VERSION is hard coded to be set to version 1.1.0. According to these lines
https://github.com/mapserver/mapserver/blob/ee52c482531b0f8fa2e2fe4abe0fb2d979f93f40/mapwms.c#L4467-L4469
WMS version 1.3.0 only supports SLD_VERSION 1.1.0. I'm not entirely sure if this is correct as I've not found anything in the standards stating this.


sdlime:
There are some changes to the autotest python in the mix in addition to the WMS changes.

ravhed:
> There are some changes to the autotest python script in the mix in addition to the WMS changes.

Yes that is correct and I should probably have mentioned that. The demime function as it was previously didn't work for the PNG generated by the WMS client GetLegendGraphic test I added (it kept the content-type header even though I set it to be demimed). With the changes I did it works for that case as well. I'm actually not sure why this PNG was different from the other ones but I assume it has something to do with the WMS client.

geographika:
Looks good to me, and just reran tests against master. Thanks @ravhed !

jmckenna:
thanks @ravhed !!

PR #5926, opened 2019-11-21
Legendicon 500 error

sdlime:
This fixes an error (segfault) with legendicon requests against certain layer/class configurations. In my case it was with lines using TrueType symbols placed along the line (e.g. railroad ties).

--Steve


sdlime:
Should squish/squash to one commit message. @rouault, can this be done from the UI or does one have to go to the command line?

rouault:
>  can this be done from the UI

Sure. Just at the right of the "Merge pull request" button there is a small bottom-oriented arrow ti display a combobox. Click on it and select "Squash and merge". Then click on the updated button, and adjust the squashed commit title and message. Note: the setting will apparently remain for next visits. So be careful to check it if you want to do a regular merge rather than a squash

sdlime:
Woot! Thanks, never noticed that before. --Steve

PR #5933, opened 2019-11-26
Revert #5929 due to line endings

geographika:
Revert "Make numlabels and labelLeaderObj immutable (#5929)" due to line ending changes (thanks Visual Studio). 

This reverts commit 5a06fc84789bc2daf9f936dcaeacad382673f6ce.

### DIFF
--- a/mapserver.h
+++ b/mapserver.h
@@ -579,7 +579,7 @@ extern "C" {
     char *filter;
     struct _CompositingFilter *next;
   } CompositingFilter;
-
+  
   typedef struct _LayerCompositer{
     CompositingOperation comp_op;
     int opacity;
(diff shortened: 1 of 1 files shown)


geographika:
@rouault @sdlime - apologies I didn't notice my editor had changed the line endings from Unix to Windows on this 2 line pull request to mapserver.h
I've reverted the commit in a branch and made this new pull request - is this the best way to revert the change to restore the history on mapserver.h? Is it ok to merge?

rouault:
This will not "restore" history: w would have to rewrite it with git rebase etc, but this is a bad practice to do in the master branch, so yes, the git revert here is the appropriate action.

We could probably also have some CI testing of line endings to reject a PR that would change them

PR #5934, opened 2019-11-27
Force LF line endings

geographika:
As per https://stackoverflow.com/a/42136008/179520 this should force LF endings. 
Hopefully this will avoid errors such as the one I made with the #5929 and its reversion in #5933.


dmorissette:
With this change, when we clone a new copy of the master from github, all binary files in the msautotest/expected folders end up appearing as modified due to a linefeed difference in "git status". Maybe something missing or your setting is a bit too agressive?

```
$ git clone https://github.com/mapserver/mapserver.git ms-test
Cloning into 'ms-test'...
[...]
$ cd ms-test/
$ git status
On branch master
Your branch is up-to-date with 'origin/master'.
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   msautotest/renderers/data/VL-Gothic-Regular.ttf
	modified:   msautotest/renderers/data/cities.shp
	modified:   msautotest/renderers/data/cities.shx
...
```

rouault:
I had an odd feeling this might have unwanted side effects. Reverting

rouault:
Reverted per https://github.com/mapserver/mapserver/pull/5936

alexbrault:
Looking into this further (as I don't encounter the same problem as @dmorissette on my local machine), it seems to be a bug in git that was fixed with git 2.10.0 (released in September 2016)

geographika:
Ok looks like that made things worse!
Looks like GDAL had a similar issue - https://github.com/OSGeo/gdal/blob/master/.gitattributes

Note [git issue fix noted here](https://github.com/git/git/blob/master/Documentation/RelNotes/2.10.0.txt#L248). 

The suggestiong by @rouault for some CI checks for line endings on pull requests seems a better way to go (compared to JS/Python C seems to lack a nice linting tool from my initial searches).

rouault:
> C seems to lack a nice linting tool from my initial searches

There are tools like clang-format (used by PROJ) or astyle (used by QGIS, openjpeg) that can be used to check that code is formatted according to a standard defined by the project. Not sure they check for line endings by the way and requires to do a global reformat of the code to comply with the formatting choices.

GDAL has a pretty ad-hoc approach with a few scripts like https://github.com/OSGeo/gdal/blob/master/gdal/scripts/detect_tabulations.sh , but we do not actually check line endings

PR #5935, opened 2019-11-29
Assert our addiction to GDAL and PROJ, and make them required dependencies

rouault:
Because:
- this was already the case. branch-7-4 does not even compile without them.
- the sheer number of places where they were needed is an evidence that MapServer
  is not MapServer without them
- it makes the code easier to comprehend.


jmckenna:
related ticket in docs: mapserver/docs#236

PR #5937, opened 2019-11-30
Add setSymbolByName test

geographika:
See mailing list post at [1]. Setting a style symbol name has no effect. This pull request makes it immutable and adds a test for the newer `setSymbolByName` as added in #1835

[1] https://lists.osgeo.org/pipermail/mapserver-users/2019-November/081492.html


sdlime:
Looks good to me.

PR #5939, opened 2019-12-02
Require C++11 and C99; bump CMake min version to 3.0

rouault:


jmckenna:
these changes seem to have affected PHPNG (issue #6015)

Issue #5798, opened 2019-04-30
PHP MapScript WMS failing on GetMap requests


jwgda:
PHP Mapscript WMS GetMap requests are not returning any result.

WMS requests going through CGI requests with a static mapfile are successful, e.g.

`http://localhost:8080/cgi-bin/mapserv.exe?map=mymap.map&SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&FORMAT=image%2Fpng&TRANSPARENT=TRUE&LAYERS=L_01&SRS=EPSG%3A3857&CRS=EPSG%3A3857&STYLES=&WIDTH=1920&HEIGHT=1068&BBOX=-19200000%2C-10680000%2C19200000%2C10680000`

However, loading that same mapfile in PHP Mapscript with exact same parameters returns no result.

No errors are written in the MS_ERRORFILE, and owsDispatch command returns 0 (MS_SUCCESS).

Example of code that returns no result.

`$oMap = new mapObj("mymap.map");`
`$request = new OWSRequest();`
`$request->setParameter('service','WMS');`
`$request->setParameter('SLD_VERSION','1.1.0');`
`msIO_installStdoutToBuffer();`
`$oMap->owsDispatch($request);`
`$contenttype = msIO_stripStdoutBufferContentType();`
`header('Content-type: image/png');`
`msIO_getStdoutBufferBytes();`
`msIO_resetHandlers();`

sdlime:
Hi @jwgda... This should be reported via the mailing lists where many more eyes can take a look at it. Lot's of PHP/MapScript users out there but I'm not one of them. --Steve

jwgda:
Seems like I'm not the only one

https://github.com/mapserver/mapserver/issues/5130

sdlime:
@jwgda, what version of everything are you running? I'll reopen this and close the old ticket... I would still take the question to the user list to see if others are experiencing the same issue. I thought this was a pretty common use case. --Steve

jwgda:
Thanks Steve.  I'm using the MS4W 4.0 package build which contains: MapServer 7.4.0-dev, GDAL 2.4.0, Apache 2.4.38, PHP 7.2.16, MapCache 1.7dev.

I also posted to the MS4W User mailing list a couple weeks ago, but no response yet (https://lists.ms4w.com/pipermail/ms4w-users/2019-April/thread.html).  If there's another list I can post to, let me know.

sdlime:
[...]

Issue #5840, opened 2019-07-25
SWIG - Send image to stdout


richie21:
Hi. This request is to fill a gap between PHPMapScript and SWIG.

With PHP MapScript, the [saveImage([string filename, MapObj oMap])](https://mapserver.org/fr/mapscript/php/phpmapscript.html#id15) method with no argument exists and allows to send image to stdout.

```
 int saveImage([string filename, MapObj oMap])

    Writes image object to specified filename. Passing no filename or an empty filename sends output to stdout
```

In MapServer 7.4, the equivalent with SWIG seems to be [write( [ FILE file=NULL ] )](https://mapserver.org/fr/mapscript/mapscript.html#imageobj-methods). However this method is only supported by Python and C# according to the documentation.

Workaround :

```php
$tmpFilename = tempnam($mapObj->web->imagepath, 'tmp');

// save the image to a temporary file on the hard disk
$imageObj->save($tmpFilename);

// read temporary file and writes it to the output buffer. 
readfile($tmpFilename);

// delete temporary file
unlink($tmpFilename);
```

geographika:
@richie21 - the Python `write` function extends the `imageObj` in MapScript. You can see the code [here](https://github.com/mapserver/mapserver/blob/master/mapscript/python/pyextend.i#L324). 

A similar extend function would need to be added to `mapscript/phpng/php7module.i`. 

You could maybe base it on the PHPMapScript [write function here](https://github.com/mapserver/mapserver/blob/master/mapscript/php/image.c#L258). Although I have no idea about PHP and C integration. You may want to look at the [SWIG PHP docs](http://www.swig.org/Doc4.0/Php.html).

jmckenna:
This has been just fixed by @bragef (#5943) and merged into master.  Please re-try your tests with today's git-master; I confirmed the fix on Windows with PHP  7.2.

richie21:
Thanks @bragef. `echo $image->getBytes();` works. That writes the image data directly to stdout.

PR #5943, opened 2019-12-05
Add PHP typemap for gdBuffer. 

bragef:
This patch adds a SWIG typemap to return gdBuffer as a string in PHP, similar to the Python Mapscript behaviour (ref. RFC-16).  This fixes the bug reported in #5798, where PHP Mapscript msIO_getStdoutBufferBytes() returns a resource handle instead of the actual buffer.

### DIFF
--- a/mapscript/phpng/php7module.i
+++ b/mapscript/phpng/php7module.i
@@ -3,3 +3,11 @@
   php_info_print_table_row(2, \"MapServer Version\", msGetVersion());
   php_info_print_table_end();
 "
+
+
+/* To support imageObj::getBytes */
+%typemap(out) gdBuffer {
+    RETVAL_STRINGL((const char*)$1.data, $1.size);
+    if( $1.owns_data )
+       msFree($1.data);
+}


jmckenna:
So far I cannot get this to work, on Windows (msIO_getStdoutBufferBytes returns nothing).  Will do more tests with your pull request though, maybe it is my test steps here.

bragef:
I have no Windows system to test mapserver on (only Linux or MacOS), so I have no opportunity to test this on Windows. 

Using the example natural earth mapfile on https://live.osgeo.org/en/quickstart/mapserver_quickstart.htm, this example should draw a map of Europe with the patch, and return an empty png without the patch.


```
require_once('mapscript.php');

$params = ['LAYERS' =>  'Countries','FORMAT' => 'image/png',
           'SERVICE'=> 'WMS','CRS' => 'EPSG:4326',
           'VERSION'=> '1.3.0','REQUEST'=> 'GetMap',
           'BBOX' => '33,-15,72,42','WIDTH' => 500,
           'HEIGHT'=> 500];

$map = new mapObj($mapfile);

$request = new OWSRequest();
foreach($params as $k=>$v) {
    $request->setParameter($k,$v);
}
msIO_installStdoutToBuffer();
$map->owsDispatch($request);
$contenttype = msIO_stripStdoutBufferContentType();
header('Content-type: '.$contenttype);
echo msIO_getStdoutBufferBytes();
msIO_resetHandlers();
```

bragef:
The typemap also fixes getBytes(), and makes it possible to write the image buffer directly to stdout without creating a temporary file (ref. #5840).

```
$map = new mapObj($mapfile);
$image = $map->draw();
header('Content-type: image/png');
echo $image->getBytes();
```

jmckenna:
@bragef I have both tests working on Windows with your changes.  Thank you so much!  This fix was much needed.  Kudos to you!

Issue #5919, opened 2019-11-12
Support group in WMS GetMap request with FILTER parameter


adube:
MapServer supports receiving the Vendor-Specific OGC FILTER parameter in WMS requests, which was implemented in RFC 118.

If a group is used instead of a layer in the url, then MapServer may return an error such as this:

```
<ServiceExceptionReport xmlns="http://www.opengis.net/ogc" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.3.0" xsi:schemaLocation="http://www.opengis.net/ogc http://schemas.opengis.net/wms/1.3.0/exceptions_1_3_0.xsd">
<ServiceException code="InvalidParameterValue">
msWMSApplyFilter: WMS server error. Wrong number of filter elements, one filter must be specified for each requested layer.
</ServiceException>
</ServiceExceptionReport>
```

This is due to the fact that behind the scene, MapServer explodes the group into layer names and the number of filters given in the url does not match the number of layers. Let's say that `groupA` has 2 layers, then:

|**Url snippet**| **Output** |
|------|------|
|`&LAYERS=groupA&FILTER=(...)`| Error|
|`&LAYERS=groupA&FILTER=(...)(...)`| Okay|

## Solution suggestion

So, the goal would be to make MapServer apply the filter given to the group to all the layers in that group.

It is expected that the filter must be compatible with all the layers of the group. For example, the fields in the filter must be defined in each layer. If something is invalid, then MapServer would return an error.

## References

 * Original MapServer RFC: https://mapserver.org/development/rfc/ms-rfc-118.html
 * Original Pull Request for MapServer: https://github.com/mapserver/mapserver/pull/5480

jratike80:
I agree with Andrea Aime https://stackoverflow.com/questions/34474852/how-to-filter-geoserver-layer-group-using-cql-filter#34479406 :

> The general answer is simple: you don't filter a group with CQL filter.
> 
> Getting into specifics, you can apply a single filter to the layer group, but it has to be valid for each and every layer in the group, which often is not the case. When you want to filter, it's best to split up the group into its individual layers, and apply a CQL_FILTER list (CQL filters separated by ;, one filter for each of them.
>

adube:
Thanks for the quick reply.

The idea main idea behind this is to avoid requiring to duplicate the same filter over and over.  In a scenario where the a group would have (let's say) 10 layers, each sharing the same properties, then the URL would need to have 10 times the exact same filter.

[...]

PR #5945, opened 2019-12-06
#5919 - Add support of group layer with filter in WMS query

charles-plante:
We can now apply a filter to a group of layers without having to match the numbers of filter.

The layer params should contains the same numbers of params as the filter


charles-plante:
@rouault One of the AppVeyor tests fail but all the Travis ones are okay. Do you have an idea why ?

mapows.h
```diff
@@ -128,8 +128,8 @@ typedef struct {
 
 /* owsRequestObj: Represent a OWS specific request with its enabled layers */
 typedef struct {
-  int numlayers;
-  int *enabled_layers;
+  int numlayers, numwmslayerargs;
```
rouault: please declare each member on a separate line

mapwms.c
```diff
@@ -376,7 +372,7 @@ int msWMSApplyFilter(mapObj *map, int version, const char *filter,
       continue;
 
     /* Skip empty filters */
-    if (paszFilters[curfilter][0] == '\0') {
+    if (paszFilters[ows_request->layerwmsfilterindex[lp->index]][0] == '\0') {
```
rouault: Use curfilter = ows_request->layerwmsfilterindex[lp->index] and reuse it aftward

mapwms.c
```diff
@@ -1079,8 +1075,13 @@ int msWMSLoadGetMapParams(mapObj *map, int nVersion,
       isUsedInNestedGroup = (int*)msSmallCalloc(map->numlayers, sizeof(int));
       msWMSPrepareNestedGroups(map, nVersion, nestedGroups, numNestedGroups, isUsedInNestedGroup);
 
-      for (k=0; k<numlayers; k++) {
-        layerfound = 0;
+      if (ows_request->layerwmsfilterindex != NULL)
+        msFree(ows_request->layerwmsfilterindex);
+      ows_request->layerwmsfilterindex = (int*)msSmallMalloc(map->numlayers * sizeof(int));
```
rouault: is there a guarantee that all entries of layerwmsfilterindex will be initialized in the following loop ? If not, that might explain the crash you get on some platforms. Perhaps use msSmallCalloc() to make sure that all entries are initialized

rouault:
I'm not super familiar with that part of the code. Anyway there's still an issue you have to fix. The Windows CI consistently crash on ows_test.py. Perhaps try running that test under Valgrind

sdlime:
I'll to try and review too, hopefully @dmorissette can as well.

dmorissette:
Thanks a lot for your review. Yes, I'll be working with Charles to finalize this PR.

mapwms.c
```diff
@@ -347,25 +347,21 @@ int msWMSApplyFilter(mapObj *map, int version, const char *filter,
   if (filter[0] == '(') {
     paszFilters = FLTSplitFilters(filter, &numfilters);
 
-    if ( paszFilters && numfilters > 0 && numlayers != numfilters ) {
-      msFreeCharArray(paszFilters, numfilters);
-      paszFilters = NULL;
-    }
```
dmorissette: Why was this block removed? Removing it creates a possible leak in case of error.

mapwms.c
```diff
@@ -959,9 +956,9 @@ int msWMSLoadGetMapParams(mapObj *map, int nVersion,
   outputFormatObj *format = NULL;
   int validlayers = 0;
   char *styles = NULL;
-  int numlayers = 0;
+  int numwmslayerarg = 0;
```
dmorissette: Ideally renaming this variable would have been in a separate pull request and it would have been called "numwmslayerargs" (plural)... but we'll accept it as is for this time.

mapwms.c
charles-plante: As discussed, we longer want to validate that the number of wms layer match the number of filters. This is so because when we're using a single group of layer, we want only one filter to be apply to all layers inside the group.

dmorissette: Agreed. I had misinterpreted those lines on my first review.

mapwms.c
charles-plante: Changs will be made

PR #5946, opened 2019-12-11
Include mserror.i and call msSetup() for PHP7 Swig mapscript

bragef:
The default exception handling routines in mserror.i is missing if the only mapscript module defined is SWIGPHP7. This causes all mapscript errors to throw an uncathable 'E_ERROR' and quit php instead of throwing an exception.

### DIFF
--- a/mapscript/mapscript.i
+++ b/mapscript/mapscript.i
@@ -197,7 +197,7 @@ typedef struct {
 ============================================================================
 */
 
-#if defined(SWIGCSHARP) || defined(SWIGJAVA) || defined(SWIGRUBY)
+#if defined(SWIGCSHARP) || defined(SWIGJAVA) || defined(SWIGRUBY) || defined(SWIGPHP7)
 %include "../mserror.i"
 #endif
 
--- a/mapscript/phpng/php7module.i
+++ b/mapscript/phpng/php7module.i
@@ -11,3 +11,20 @@
     if( $1.owns_data )
        msFree($1.data);
 }
+
+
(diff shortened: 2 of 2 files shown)


bragef:
Please ignore this for now; will do some more testing here.

jmckenna:
roger, thanks @bragef

bragef:
I think this should be safe to commit. I was a bit hesitant because I got a some exceptions  which did not make sense after enabling the exeption handlers, e.g.  the `getMetaData`  methods trown an exception for a non-existent key. However, looking at the code this was more of a mis-feature than a bug. At any rate, it can be avoided by using acessing the hashObj instead. I also got a couple of exceptions because of a font initialisation bug (will open a new issue for this).

bragef:
The PHP Swig mapscript was missing a call to `msSetup()`, so that the font cache was not properly initialized. 

This causes labels to not appear on maps, or appear erratically. Without exceptions, I got a lot of errors of the type 
> [Wed Dec 11 13:59:21 2019].450696 msGetFontFace(): General error message. Freetype was unable to load font file "arial.ttf" for key "arial"
and with exceptions enabled mapscript threw an exeption when `drawLabelCache()` were called.

This is the same issue as is described in this thread:
http://osgeo-org.1560.x6.nabble.com/PHP-Mapscript-swig-issue-on-macOS-td5414228.html

jmckenna:
I cannot reproduce the initial font errors reported, and still cannot.  However your changes look valid.  Merging...

jmckenna:
thanks again @bragef !!

PR #5948, opened 2019-12-12
Backport pull request #5926 from master.

sdlime:
This fixes a segfault issue with mode=legendicon and certain conditions where the opaque pointer back to the parent mapObj is used, for instance when dealing with truetype symbols along a line.


jmckenna:
mercy buckets steve!

Issue #5952, opened 2019-12-18
Mapscript pointObj::draw() segfaults


bragef:
PHP Swig Mapscript `pointObj::draw()` will often segfault rather than throw an exception or return `MS_FAILURE` when the operation fails.

The PHP example below gives three different ways to produce a segfault.

```
<?php
require_once('./mapscript.php');

$map = mapscript::msLoadMapFromString('                                         
MAP                                                                             
 EXTENT 0 0 90 90                                                               
 SIZE 500 500                                                                   
                                                                                
 SYMBOL                                                                         
  NAME "circle"                                                                 
  TYPE ELLIPSE                                                                  
  POINTS 1 1 END                                                                
  FILLED true                                                                   
 END                                                                            
                                                                                
 LAYER                                                                          
  NAME "punkt"                                                                  
  STATUS ON                                                                     
  TYPE POINT                                                                    
 END                                                                            
                                                                                
END                                                                             
','.');

$image = $map->prepareImage();
$layer = $map->getLayerByName("punkt");

$class = new classObj();
$style = new styleObj();
$style->outlinecolor->setHex('#00aa00ff');
$style->size =  10;
$style->setSymbolByName($map, 'circle');
$class->insertStyle($style);
$class_idx = $layer->insertClass($class);

## Example 1: If the projection is not set, pointObj::draw() segfaults.                 
#$map->setProjection("init=epsg:4326");                                         
#$layer->setProjection( "init=epsg:4326");                                      

## Example 2: If the class is created with new classObj(), and a labelObj() 
## is not added to the class, pointObj::draw() segfaults
#$class->addLabel(new labelObj());                                              

$point = new pointObj(rand(0,90), rand(0,90));
$point->draw($map, $layer, $image, $class_idx, "test");

## Example 3: If given an invalid class index,  pointObj::draw() segfaults  
$point->draw($map, $layer, $image, 1000, "");

$map->drawLabelCache($image);
# Save to file if cli, display if web.                                          
$imgbytes = $image->getBytes();
if(php_sapi_name() === 'cli') {
    echo strlen($imgbytes)." bytes written to file.\n";
    file_put_contents('testpoints.png', $imgbytes);
} else {
    header('Content-type: image/png');
    echo $imgbytes;
}
```

sdlime:
What version?

bragef:
Mapserver 7.4.3 or current HEAD (and PHP 7.2.24).

jmckenna:
[...]

PR #5953, opened 2019-12-19
Add sanity checks to avoid segfaults in Mapscript pointObj::draw()

bragef:
Issue #5952


bragef:
> The only question I'd have is whether we should return an error when labeltext is supplied but there is no label definition or if we should just skip label drawing. Does that specific case cause a segfault? --Steve

It did – but that can be avoided by checking if the label definition is present. I have updated the patch to only throw error on invalid classindex.

sdlime:
Closed/re-opened to trigger CI checks.

sdlime:
Looks ok to me... Requested another set of eyes (@geographika) and then we can then merge.

Issue #5724, opened 2019-01-07
[PATCH] Add option to skip failures on OGR layers


MarcWeustink:
For generating floorplan maps we sometimes receive Autocad drawings. 
We import those ad DXF through a OGR layer. Recently we received some where one feature triggered an error in the GDAL library [https://github.com/OSGeo/gdal/issues/1011] . As a result no map was rendered. In our case having no map is worse than missing one feature.
Therefore I've added a processing option `PROCESSING "SKIP_FAILURES=YES"` so these errors can be ignored.

This is a patch based on mapserver 7.2.0
[mapserver_mwe_ignore_feature2.zip](https://github.com/mapserver/mapserver/files/2733542/mapserver_mwe_ignore_feature2.zip)

sdlime:
Even points to a simpler patch if one wanted to have a go... I did not apply the submitted patch. --Steve

PR #5956, opened 2019-12-28
Issue 5724 - Skip Failures Processing Option for OGR Driver

sdlime:
See #5724 for more information.


sdlime:
Reminder: need to squash before merging...

rouault:
That's reasonable. But a simpler patch would be to not call at all CPLGetLastErrorType(). Most GDAL code that iterates over features just rely on GetNextFeature() == NULL to stop iterations.

sdlime:
> That's reasonable. But a simpler patch would be to not call at all CPLGetLastErrorType(). Most GDAL code that iterates over features just rely on GetNextFeature() == NULL to stop iterations.

Do think it would be easier to start over or edit the patch?

rouault:
> Do think it would be easier to start over or edit the patch?

perhaps start over

sdlime:
Closing. Will update #5724 with notes... --Steve

PR #5964, opened 2020-01-08
Add test for MapScript point.draw method

geographika:
See #5952

### DIFF
--- a/mapscript/python/tests/cases/point_test.py
+++ b/mapscript/python/tests/cases/point_test.py
@@ -28,11 +28,11 @@ import unittest
 import mapscript
 
 
-have_image = 0
+have_image = False
 
 try:
     from PIL import Image
-    have_image = 1
+    have_image = True
 except ImportError:
     pass
 
@@ -125,7 +125,7 @@ class PointObjTestCase(unittest.TestCase):
             pyimg = Image.open(filename)
             assert pyimg.size == (500, 500)
         else:
-            assert 1
+            assert True
 
     def testPoint__str__(self):
         """return properly formatted string"""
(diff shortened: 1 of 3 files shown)


mapscript/python/tests/cases/point_test.py
```diff
@@ -28,6 +28,15 @@
 import mapscript
 
 
+have_image = 0
```
rouault:
```suggestion
have_image = False
```

mapscript/python/tests/cases/point_test.py
```diff
@@ -28,6 +28,15 @@
...
+
+try:
+    from PIL import Image
+    have_image = 1
```
rouault:
```suggestion
    have_image = True
```

mapscript/python/tests/cases/point_test.py
```diff
@@ -70,6 +79,54 @@ def testSetXYZ(self):
...
+
+        filename = 'testDrawPoint.png'
+        with open(filename, 'wb') as fh:
+            img.write(fh)
+
+        if have_image:
+            pyimg = Image.open(filename)
+            assert pyimg.size == (500, 500)
+        else:
+            assert 1
```
rouault: assert 1 is a no-op.

geographika: This was taken from the approach in image_test.py - if PIL isn't installed then the test passes automatically

rouault: This is just useless code. No need to perpetuate bad past patterns :-)

rouault: I mean: you don't need that else: assert 1 (or True) branch at all. It doesn't bring anything. Or I'm missing something...

geographika: I thought it might result in pytest ignoring it as a test. I just checked and it will pass without the assert (assuming it gets to the end of the function) - so removing.

geographika:
@rouault - I've added the suggested review changes - any objections to merging?

rouault:
there are still a few useless assert True remaining, but there were already there. So no objection in merging

PR #5967, opened 2020-01-08
WFS: fix WFS 2.0 numberMatched computation, by opening temporary new connection(s)

rouault:


geographika:
@rouault - I've been testing this along with an implementation of paging in the MSSQL driver by @szekerest - it prevents lots of individual queries being run via WFS giving a huge performance boost. Is this ready to merge? I've not tested with other database drivers.

Issue #5842, opened 2019-07-31
Support WFS Paging at database level for MSSQL Driver


geographika:
## Database Paging

WFS requests support a [STARTINDEX](https://mapserver.org/ogc/wfs_server.html#mapserver-wfs-extensions) parameter which allows results to be paged through in conjunction with MAXFEATURES. 

A sample request would look as follows:

```mapserv -nh "QUERY_STRING=map=test.map&service=WFS&version=2.0.0&request=GetFeature&typeName=MyLayer&outputFormat=geojson&sortBy=Property1%20ASC&startIndex=0&count=20"```

The PostGIS driver supports paging at the [database level](https://github.com/mapserver/mapserver/blob/85e294e8c8ce3c9c455fea0137d4af1ee2feced0/mappostgis.c#L3924). Other drivers fetch all the results and then MapServer carries out the paging in [mapquery.c](https://github.com/mapserver/mapserver/blob/139ed5a40c8957c2c34695f50eec2cd5cf4894e8/mapquery.c#L883). Supporting drivers implement `layer->vtable->LayerGetPaging`. 

Allowing WFS Paging at database level for the MSSQL Driver would greatly improve performance when paging through results. 

MS SQL 2012+ can simulate paging in SQL with [OFFSET and FETCH](https://stackoverflow.com/questions/971964/limit-10-20-in-sql-server). A fallback for older versions (and for Azure hosted SQL Server) is [RowNumber](https://stackoverflow.com/a/47346977/179520). The latter approach may be prereable as it is more universal. 

## Performance Issues

The lack of paging causes a huge overhead and performance hit when providing `&startIndex=0&count=20`. The database runs the full SQL query for the layer (with no paging) 3 times when running a WFS 2.0.0 GetFeature request:

1. In [mapwfs.c](https://github.com/mapserver/mapserver/blob/master/mapwfs.c#L3772) with the `msWFSRetrieveFeatures` function (and then `msLayerWhichShapes` in `mapquery.c` (with no paging so the full query is run). 

2. In [mapwfs.c](https://github.com/mapserver/mapserver/blob/master/mapwfs.c#L3801) - `msWFSComputeMatchingFeatures` which again calls `msWFSRetrieveFeatures` and then `LayerDefaultGetShapeCount` in `maplayer.c` which runs the full query again. 
[...]

PR #5971, opened 2020-01-14
MSSQL: Implement server side paging support (#5842)

szekerest:


mapmssql2008.c
```diff
@@ -2453,6 +2527,142 @@ int msMSSQL2008LayerGetShape(layerObj *layer, shapeObj *shape, resultObj *record
...
+    layerinfo = getMSSQL2008LayerInfo(layer);
+
+    if (!layerinfo) {
+        /* Layer not open */
+        msSetError(MS_QUERYERR, "msMSSQL2008LayerGetShapeCount called on unopened layer (layerinfo = NULL)", "msMSSQL2008LayerGetShapeCount()");
+
+        return MS_FAILURE;
+    }
+
+#ifdef USE_PROJ
```
rouault: USE_PROJ is no longer defined in master

mapmssql2008.c
```diff
@@ -2453,6 +2527,142 @@ int msMSSQL2008LayerGetShape(layerObj *layer, shapeObj *shape, resultObj *record
...
+        // Reproject the passed rect into the layer projection and get
+        // the SRID from the rectProjection
+        msProjectRect(rectProjection, &(layer->projection), &searchrectInLayerProj); /* project the searchrect to source coords */
+    }
+#endif
+
+    if (searchrectInLayerProj.minx == searchrectInLayerProj.maxx || 
+        searchrectInLayerProj.miny == searchrectInLayerProj.maxy) {
+        /* create point shape for rectangles with zero area */
+        sprintf(box3d, "%s::STGeomFromText('POINT(%.15g %.15g)',%s)", /* %s.STSrid)", */
```
rouault:
```suggestion
        snprintf(box3d, sizeof(box3d), "%s::STGeomFromText('POINT(%.15g %.15g)',%s)", /* %s.STSrid)", */
```

mapmssql2008.c
```diff
@@ -2453,6 +2527,142 @@ int msMSSQL2008LayerGetShape(layerObj *layer, shapeObj *shape, resultObj *record
...
+            layerinfo->geom_column_type, searchrectInLayerProj.minx, searchrectInLayerProj.miny, layerinfo->user_srid);
+    }
+    else {
+        sprintf(box3d, "%s::STGeomFromText('POLYGON((%.15g %.15g,%.15g %.15g,%.15g %.15g,%.15g %.15g,%.15g %.15g))',%s)", /* %s.STSrid)", */
```
rouault:
```suggestion
         snprintf(box3d, sizeof(box3d), "%s::STGeomFromText('POLYGON((%.15g %.15g,%.15g %.15g,%.15g %.15g,%.15g %.15g,%.15g %.15g))',%s)", /* %s.STSrid)", */
```

mapmssql2008.c
```diff
@@ -1446,32 +1450,77 @@ static int prepare_database(layerObj *layer, rectObj rect, char **query_string)
...
+  if (layerinfo->paging && (layer->maxfeatures >= 0 || layer->startindex > 0))
+    paging_query = msStringConcatenate(paging_query, "SELECT ");
 
   /* adding items to the select list */
   for (t = 0; t < layer->numitems; t++) {
 #ifdef USE_ICONV
       query = msStringConcatenate(query, "convert(nvarchar(max), [");
       query = msStringConcatenate(query, layer->items[t]);
-      query = msStringConcatenate(query, "]),");
+      query = msStringConcatenate(query, "]) '");
```
rouault: I'm probably missing something but it looks the #ifdef USE_ICONV and the #else branch code are strictly identical, except the initial line. So this could be largely simplified by placeing just it in the #if #else #endif, and have the rest to be common conde

szekerest: yes, that could be simplified in that way. 
However I don't understand the reason of the code change above.

mapmssql2008.c
szekerest: Applied the change

mapmssql2008.c
szekerest: Fixed

PR #5972, opened 2020-01-14
Fix Equal Earth rendering for wide images

### LINKED ISSUE #5958: Incorrectly rendered EPSG:4326 raster in Equal Earth, depending on output image size
When rendering a raster stored in EPSG:4326 to EPSG:8857 (Equal Earth), values are plotted in incorrect locations for certain ouput image sizes.

Mapfile:
```
MAP
  PROJECTION 
    "init=epsg:4326"
  END

  EXTENT -180 -90 180 90        

WEB
  METADATA
  "ows_enable_request" "*"
  "wms_srs"            "EPSG:4326 EPSG:8857" 
  END
END

LAYER
  NAME "rast"
  TYPE RASTER
  STATUS OFF 

  DATA "/tmp/data.tif"

  CLASS
    EXPRESSION ([pixel] >= 0)
    STYLE
      COLORRANGE 0 0 0 0 0 255
      DATARANGE 0 60
      RANGEITEM "pixel"
    END
  END
END

END
```

Raster data: [data.zip](https://github.com/mapserver/mapserver/files/4012953/data.zip)

Expected output:

```
./mapserv -nh "QUERY_STRING=map=/home/dan/dev/mapserver/test.map&service=wms&version=1.3.0&request=GetMap&format=png&width=2162&height=1055&crs=EPSG%3A8857&layers=rast&BBOX=-17243959.06221695,-8392927.59846645,17243959.06221695,8392927.59846645&transparent=TRUE" > ok.png
```
image

Incorrect output (image width increased from 2162 to 2163 pixels):

```
./mapserv -nh "QUERY_STRING=map=/home/dan/dev/mapserver/test.map&service=wms&version=1.3.0&request=GetMap&format=png&width=2163&height=1055&crs=EPSG%3A8857&layers=rast&BBOX=-17243959.06221695,-8392927.59846645,17243959.06221695,8392927.59846645&transparent=TRUE" > bad.png
```
image

```
./mapserv -v
# MapServer version 7.5-dev OUTPUT=PNG OUTPUT=JPEG SUPPORTS=PROJ SUPPORTS=AGG SUPPORTS=FREETYPE SUPPORTS=CAIRO SUPPORTS=ICONV SUPPORTS=FRIBIDI SUPPORTS=WMS_SERVER SUPPORTS=WFS_SERVER SUPPORTS=WCS_SERVER SUPPORTS=FASTCGI SUPPORTS=GEOS SUPPORTS=PBF INPUT=JPEG INPUT=POSTGIS INPUT=OGR INPUT=GDAL INPUT=SHAPEFILE
```

dbaston:
This PR resolves #5958 , at least for my test case. Along the lines of @rouault 's suggested fix, the interpolation test point is taken to be just after the midpoint of the scanline. (I think this is more conservative than taking the point before the middle.) Since `msApproxTransformer` bails early for scanlines with <= 5 points, the middle+1 point will always be available.

### DIFF
--- a/mapgeos.c
+++ b/mapgeos.c
@@ -706,7 +706,7 @@ shapeObj *msGEOSGeometry2Shape(GEOSGeom g)
 #endif
 
 /*
-** Maintenence functions exposed to MapServer/MapScript.
+** Maintenance functions exposed to MapServer/MapScript.
 */
 
 void msGEOSFreeGeometry(shapeObj *shape)
--- a/mapresample.c
+++ b/mapresample.c
@@ -58,7 +58,7 @@ int InvGeoTransform( double *gt_in, double *gt_out )
 
   inv_det = 1.0 / det;
 
-  /* compute adjoint, and devide by determinate */
+  /* compute adjoint, and divide by determinate */
 
   gt_out[1] =  gt_in[5] * inv_det;
   gt_out[4] = -gt_in[4] * inv_det;
@@ -129,7 +129,7 @@ msNearestRasterResampler( imageObj *psSrcImage, rasterBufferObj *src_rb,
       /*
        * We test the original floating point values to
        * avoid errors related to asymmetric rounding around zero.
-       * (Also note bug #3120 regarding nearly redundent x/y < 0 checks).
+       * (Also note bug #3120 regarding nearly redundant x/y < 0 checks).
        */
       if( x[nDstX] < 0.0 || y[nDstX] < 0.0
           || nSrcX < 0 || nSrcY < 0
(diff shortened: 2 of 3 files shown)


rouault:
Why closing ?

dbaston:
Looks like the fix works for a width of 2163 pixels but not 2560 😢 

Also looks like it broke an existing test. And I assume it should been opened against `master`, not `branch-7-4` (the GitHub default until recently, I think?)

rouault:
> Looks like the fix works for a width of 2163 pixels but not 2560

ah ok

> Also looks like it broke an existing test.

Not necessarily a show stopper. Would need to check if the difference is significant or not. The current results are not necessarily an absolute truth...

>  assume it should been opened against master

Indeed, the policy is now to target master, and backport if needed

Issue #5700, opened 2018-11-08
MSVC is showing too many compiler warnings


geographika:
The Appveyor build is currently showing over 800 warnings when compiling MapServer (as are any MSVC builds). This produces a large log output which is hard to review for serious errors. 
Ideally all compiler warnings will be fixed in time. In the meantime hiding a few low-level warnings can highlight more urgent fixes. 

Warnings are divided into 4 levels - see https://msdn.microsoft.com/en-us/library/thxezb7y.aspx

- Level 1 - severe
- Level 2 - significant
- Level 3 - (production quality) warnings
- Level 4 - informational

The following 3 errors account for the majority of these errors:

- [C4267 - Level3](https://docs.microsoft.com/en-gb/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4267?view=vs-2017) - conversion from 'size_t' to 'type', possible loss of data
- [C4267 - Level3-4](https://docs.microsoft.com/en-gb/cpp/error-messages/compiler-warnings/compiler-warning-levels-3-and-4-c4244?view=vs-2017) - 'conversion' conversion from 'type1' to 'type2', possible loss of data
- [C4018 - Level3](https://docs.microsoft.com/en-gb/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4018?view=vs-2017) - 'expression' : signed/unsigned mismatch

geographika:
Closed with #5701

PR #5974, opened 2020-01-15
Initialize status variable for cached WFS results

geographika:
On Windows the following error is thrown in the debugger as the status variable was never set:

`Run-Time Check Failure #3 - The variable 'status' is being used without being initialized.`

This caused OGR output from a WFS cache to return:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<ows:ExceptionReport xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:ows="http://www.opengis.net/ows/1.1" version="2.0.0" xml:lang="en-US" xsi:schemaLocation="http://www.opengis.net/ows/1.1 http://schemas.opengis.net/ows/1.1.0/owsExceptionReport.xsd">
  <ows:Exception exceptionCode="NoApplicableCode" locator="mapserv"/>
</ows:ExceptionReport>
```

This pull request initialises the variable at the top of the function (maybe unnecessary?), and also sets when retrieved from the cache using `msCopyShape`. 

See email thread at https://lists.osgeo.org/pipermail/mapserver-users/2020-January/081530.html

### DIFF
--- a/mapogroutput.c
+++ b/mapogroutput.c
@@ -895,7 +895,7 @@ int msOGRWriteFromQuery( mapObj *map, outputFormatObj *format, int sendheaders )
   /*      Process each layer with a resultset.                            */
   /* ==================================================================== */
   for( iLayer = 0; iLayer < map->numlayers; iLayer++ ) {
-    int status;
+    int status = 0;
     layerObj *layer = GET_LAYER(map, iLayer);
     shapeObj resultshape;
     OGRLayerH hOGRLayer;
@@ -1106,7 +1106,7 @@ int msOGRWriteFromQuery( mapObj *map, outputFormatObj *format, int sendheaders )
       if( layer->resultcache->results[i].shape )
       {
           /* msDebug("Using cached shape %ld\n", layer->resultcache->results[i].shapeindex); */
-          msCopyShape(layer->resultcache->results[i].shape, &resultshape);
+          status = msCopyShape(layer->resultcache->results[i].shape, &resultshape);
       }
       else
       {
--- a/mapogroutput.c
+++ b/mapogroutput.c
@@ -1110,16 +1110,17 @@ int msOGRWriteFromQuery( mapObj *map, outputFormatObj *format, int sendheaders )
       }
       else
       {
-        status = msLayerGetShape(layer, &resultshape, &(layer->resultcache->results[i]));
-        if(status != MS_SUCCESS) {
-            OGR_DS_Destroy( hDS );
-            msOGRCleanupDS( datasource_name );
-            msGMLFreeItems(item_list);
-            msFreeShape(&resultshape);
-            CSLDestroy(layer_options);
-            return status;
-        }
+          status = msLayerGetShape(layer, &resultshape, &(layer->resultcache->results[i]));
       }
+      
+      if(status != MS_SUCCESS) {
+          OGR_DS_Destroy( hDS );
+          msOGRCleanupDS( datasource_name );
+          msGMLFreeItems(item_list);
+          msFreeShape(&resultshape);
+          CSLDestroy(layer_options);
+          return status;
+      }      
 
       /*
       ** Perform classification, and some annotation related magic.


rouault:
Excellent catch @geographika ! For the fix to be complete, the block
```
if(status != MS_SUCCESS) {
            OGR_DS_Destroy( hDS );
            OGR_DS_Destroy( hDS );
            msOGRCleanupDS( datasource_name );
            msOGRCleanupDS( datasource_name );
            msGMLFreeItems(item_list);
            msGMLFreeItems(item_list);
            msFreeShape(&resultshape);
            msFreeShape(&resultshape);
            CSLDestroy(layer_options);
            CSLDestroy(layer_options);
            return status;
            return status;
        }
    }
```
that is currently only taken in the else case (https://github.com/mapserver/mapserver/pull/5974/files#diff-b8cb5eed516c5e213aaccd109187105eR1114), should be applied to both branches. That is move it just after the if else construct

geographika:
@rouault - changes included as suggested. 
Is it good or bad practice to set the `int status = 0;` at declaration?
Are there static analysis tools for C that could find this type of error in the MapServer codebase?

rouault:
> Is it good or bad practice to set the int status = 0; at declaration?

It is generally a good practice.

> Are there static analysis tools for C that could find this type of error in the MapServer codebase?

I'm pretty sure Visual Studio could warn about that, but only if running at a more pedantic level, which would be the super pedantic /W4 . Cf https://docs.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level?view=vs-2019#see-also and https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4701?view=vs-2019

PROJ appveyor.yml includes in the cmake invokation:
```
 -DCMAKE_C_FLAGS="/WX" -DCMAKE_CXX_FLAGS="/WX"
```
But there would be likely a LOT of work to have a clean MapServer build with that.

A possibility would be to enable specifically C4701 at the default warning level 1, with /w14701

rouault:
Cherry-picked in branch-7-4 per 006d4ba5e64f59dbab14cb3409817567bae313ff

sdlime:
I wouldn't mind making a run at a code clean up - or at least looking at the scale. What would be the approach on a mac - adding a couple of flags to cmake, doing a make and fixing issues as identified?

jmckenna:
Through the years I've had to battle this in MapCache (declare all variables at beginning of scope/block)  C89 requirement (example changes: https://github.com/mapserver/mapcache/pull/57/commits/ff5e3b1bf8b136d173ff63329fee92e82c0ee92a) so that it builds on Windows compilers.  I feel this practice should be following by rule, in MapServer and MapCache, so it is more friendly to more compilers.

rouault:
> What would be the approach on a mac - adding a couple of flags to cmake, doing a make and fixing issues as identified?

Unfortunately it seems gcc has no diagnostics for this. clang however has -Wsometimes-uninitialized that does the job, so you could likely try with cmake  -DCMAKE_C_FLAGS="-Wsometimes-uninitialized" -DCMAKE_CXX_FLAGS="-Wsometimes-uninitialized"
cppcheck can also sometimes detect such situations.

> I feel this practice should be followed by rule

You meant initializing variables, right, not declarating at top of scope ? With C99 allowed and encouraged in MapServer, we should avoid as much as possible putting variable declarations at the top of a block.  So instead of doing
```
int foo = 0;
....
foo = bar();
```
it is less error prone to do ``` int foo = bar();```. Of course that doesn't apply to all situations, like the one of this ticket where the variable assignment depend on which branch is taken. Hence in that situation 0 initialization is indeed the best choice.

jmckenna:
Yes you are correct (I had changed my wording from 'declare' to 'initialize' after commenting initially).  Good point.

geographika:
I merged a pull request a year ago to hide some warnings, as there were over 800 and it was hard to see serious issues hidden amongst lots of "possible conversion" errors. See #5700 and #5701 

@rouault - adding in `/w14701` gives 80 warnings which could be manageable to fix as a group effort and could prevent a few hidden bugs.

rouault:
> adding in `/w14701` gives 80 warnings which could be manageable to fix as a group effort and could prevent a few hidden bugs.

We should definitely fix them. That class of warnings is normally quite straightforward & mechanical to fix. Could be just a couple of hours.

Issue #5526, opened 2017-11-28
CLASS expressions are case-sensitive when exporting to SLD


geographika:
It seems as though the logical operators in ``CLASS`` ``EXPRESSION``s will only create SLD Filters when they are uppercase. 
For example in the Mapfile below only the second class will have a filter created:

```
  MAP
    SYMBOL
      NAME 'SQUARE'
      TYPE vector
      FILLED true
      POINTS
        0 0
        0 1
        1 1
        1 0
        0 0
      END
    END 
    LAYER
      STATUS ON
      NAME "Test"
      TYPE POINT
      DATA "sample"
      LABELMAXSCALEDENOM 10000
      CLASS
        NAME "Not Working"
        EXPRESSION ( ( [Field] > 0 ) and ( [Field] <= 5 ) )
        STYLE
          SYMBOL "SQUARE"
          COLOR 0 176 80
          SIZE 10
        END
      END
      CLASS
        NAME "Working"
        EXPRESSION ( ( [Field] > 5 ) AND ( [Field] <= 10 ) )
        STYLE
          SYMBOL "SQUARE"
          COLOR 0 176 80
          SIZE 10
        END
      END
    END
  END
```
It looks like this is due to a case-sensitive comparison in https://github.com/mapserver/mapserver/blob/master/mapogcsld.c#L4317

Does MapServer have a case-insensitive string search utility function that could be used here?

sdlime:
@geographika, is this still an issue? --Steve

geographika:
@sdlime - yes this is still an issue in master, and hopefully addressed in #5980. 
All the operators in the [EXPRESSION docs](https://mapserver.org/mapfile/expressions.html#logical-expressions) are uppercase but the Mapfile docs state [The Mapfile is NOT case-sensitive](https://www.mapserver.org/mapfile/#notes). It is only the SLD output that seems to be affected by case-sensitivity - the filters work fine on other features. 

While looking at the code it appears operators such as "lt" in [msSLDGetComparisonValue](https://github.com/mapserver/mapserver/blob/fdad948e08a7a5c6bd3aa0e421aa9c2ec2277897/mapogcsld.c#L4331) are case-sensitive. Should I correct these (allowing "LT") as part of this?

PR #5980, opened 2020-01-19
Use Case Insensitive Filters for SLD Output

geographika:
See issue at #5526. 
This pull request makes checks for operators case-insensitive so "AND", "and", and "Annd" in a Mapfile expression will all produce an SLD filter. 

The functions `strcasecmp`, `strcasestr`, and `toupper` are used to ensure this. 

A new test Mapfile has been added to ensure these filters are created in each case. 
TODO - add test for brackets e.g. `AND(`

### DIFF
--- a/mapogcsld.c
+++ b/mapogcsld.c
@@ -176,7 +176,7 @@ int msSLDApplySLD(mapObj *map, const char *psSLDXML, int iLayer, const char *psz
   layerObj *pasSLDLayers = NULL;
   int nStatus = MS_SUCCESS;
   /*const char *pszSLDNotSupported = NULL;*/
-  
+
   pasSLDLayers = msSLDParseSLD(map, psSLDXML, &nSLDLayers);
   if( pasSLDLayers == NULL ) {
     errorObj* psError = msGetErrorObj();
@@ -403,7 +403,7 @@ int msSLDApplySLD(mapObj *map, const char *psSLDXML, int iLayer, const char *psz
 
                     snprintf(szTmp, sizeof(szTmp), "%s", "))");
                     pszBuffer =msStringConcatenate(pszBuffer, szTmp);
-                    
+
                     msFreeExpression(&lp->filter);
                     msInitExpression(&lp->filter);
                     lp->filter.string = msStrdup(pszBuffer);
@@ -439,7 +439,7 @@ int msSLDApplySLD(mapObj *map, const char *psSLDXML, int iLayer, const char *psz
 
     }
   }
-  
+
   nStatus = MS_SUCCESS;
 
 #ifdef notdef
@@ -732,7 +732,7 @@ static char* msSLDGetCommonExpressionFromFilter(CPLXMLNode* psFilter,
         pszExpression = FLTGetCommonExpression(psNode, psLayer);
         FLTFreeFilterEncodingNode(psNode);
     }
-    
+
     return pszExpression;
 }
 
@@ -874,7 +874,7 @@ int msSLDParseNamedLayer(CPLXMLNode *psRoot, layerObj *psLayer)
           psFeatureTypeConstraint = psIter;
         } else {
           msSetError(MS_WMSERR, "Only one single FeatureTypeConstraint element "
-                    "per LayerFeatureConstraints is supported", "");  
+                    "per LayerFeatureConstraints is supported", "");
           return MS_FAILURE;
         }
       }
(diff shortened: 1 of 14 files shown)


geographika:
There are some memory leaks showing up in Travis. See https://travis-ci.org/mapserver/mapserver/jobs/639062323?utm_source=github_status
I'm not sure if this is due to my changes, or due to previously untested codepaths now being tested. 

For example:

```
Direct leak of 48 byte(s) in 1 object(s) allocated from:

    #0 0x4bb533 in __interceptor_malloc /local/mnt/workspace/tmp/ubuntu_rel/llvm/utils/release/final/llvm.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cc:67:3
    #1 0x2b28ecb6e84f in FLTCreateFilterEncodingNode /home/travis/build/mapserver/mapserver/mapogcfilter.c:970:25
    #2 0x2b28ecd088b0 in BuildExpressionTree /home/travis/build/mapserver/mapserver/mapogcsld.c:4835:18
    #3 0x2b28ecd0afb4 in msSLDParseLogicalExpression /home/travis/build/mapserver/mapserver/mapogcsld.c:4968:12
    #4 0x2b28ecd0df4c in msSLDGetFilter /home/travis/build/mapserver/mapserver/mapogcsld.c:5166:19
    #5 0x2b28eccfd066 in msSLDGenerateUserStyle /home/travis/build/mapserver/mapserver/mapogcsld.c:4150:21
    #6 0x2b28eccefbb1 in msSLDGenerateSLDLayer /home/travis/build/mapserver/mapserver/mapogcsld.c:4314:9
    #7 0x2b28eccee7d3 in msSLDGenerateSLD /home/travis/build/mapserver/mapserver/mapogcsld.c:3196:18

```

rouault:
> I'm not sure if this is due to my changes, or due to previously untested codepaths now being tested.

Likely the later. I believe that adding CPLDestroyXMLNode(psNode) at https://github.com/mapserver/mapserver/blob/e04350e23843cf3c7749aeac25cc7c63fe7e1180/mapogcsld.c#L5000 should fix it

geographika:
@rouault - thanks for the hint. I added in https://github.com/mapserver/mapserver/pull/5980/commits/547c5b2e191f19ab11dacd362a2c5f5b01d43779 but [leaks still reported](https://travis-ci.org/mapserver/mapserver/jobs/640196399?utm_medium=notification&utm_source=github_status). 

Also failing to [build in Appveyor](https://ci.appveyor.com/project/MapServer/mapserver/builds/30270640), is there some casting that needs to be done?

```
C:\projects\mapserver\mapogcsld.c(4999): error C2220: warning treated as error - no 'object' file generated [C:\projects\mapserver\build\mapserver.vcxproj]
C:\projects\mapserver\mapogcsld.c(4999): warning C4133: 'function': incompatible types - from 'FilterEncodingNode *' to 'CPLXMLNode *' [C:\projects\mapserver\build\mapserver.vcxproj]
```
I removed the erroneous `free(pszTmp);` I added as per your comment.

rouault:
>  is there some casting that needs to be done?

ah my mistake...  `FLTFreeFilterEncodingNode(psNode);` should do it hopefully

geographika:
@rouault - thanks for that - updated. I also needed to free the left and right expression variables as at https://github.com/mapserver/mapserver/pull/5980/commits/79f66b6b8f18ce08ca99fb63ca18f0850dd3f7e7 and now all reported leaks are gone. 

The existing parsing used in the code looks like it could have some edge cases that would cause it to fail e.g. if an expression contained `NOR(` it would be picked up as an `OR` (although this situation may not be possible). There is also an issue (present in the existing code) where attribute values aren't trimmed - I can open a separate issue about this. 

This pull request fixes issues around case-sensitivity (previously no filters were created for these), adds some tests for the SLD creation functions, fixes some comment typos, and fixes a couple of memory leaks.

mapogcsld.c
```diff
@@ -4993,8 +4996,8 @@ char *msSLDParseLogicalExpression(char *pszExpression, const char *pszWfsFilter)
 
       free(pszFLTExpression);
       pszFLTExpression = pszTmp;
+      FLTFreeFilterEncodingNode(psNode);
```
rouault: this should be moved just after the } of line 5000, so that psNode is also freed when != NULL but when pszFLTExpression is NULL

geographika:
@rouault - many thanks for the review and explanations

Issue #5982, opened 2020-01-29
Use SWIG 4 for MapScript Builds


geographika:
SWIG 4.0.1 is available since 2019/08/21. Currently, 3.0.12 is used for testing/CI released on 2017/01/28. This issue tracks issues relating to updating to 4.0.1

## Python

Python fails to build due to 2 issues.

1. The following line in pymodule.i:

```
/* Translates Python None to C NULL for strings */
//%typemap(in,parse="z") char * "";
```

This caused lots of the following errors : `error: expected expression before ‘,’ token if (!PyArg_ParseTuple(args, "Oz:CompositingFilter_filter_set", , &arg2)) SWIG_fail;`

I posted about this to the [SWIG Mailing list](https://sourceforge.net/p/swig/mailman/message/36649157/) and see also the [mailing list thread](https://lists.osgeo.org/pipermail/mapserver-dev/2020-January/016025.html) with @tigerfoot 

The code was a [recommended typemap](https://swig-user.narkive.com/4S9mKKDM/overriding-pyarg-parsetuple-arguments) - but this was 16 years ago so may no longer be needed. The full test suite passes after removing the code. 

2. Use of ``__swig_getmethods__`` and ``__swig_setmethods__`` which no longer exist. These have been changed to use Python properties (which weren't available until Python 2.7?). 

A couple of useful links related to this with an alternative solution using `%attribute` I couldn't get working due to the additional use of typemaps. 

* https://sourceforge.net/p/swig/mailman/message/36698052/
* https://bugs.llvm.org/show_bug.cgi?id=42284
* https://reviews.llvm.org/D63530

## C#

The C# bindings currently run into [this issue](https://github.com/swig/swig/issues/421) with SWIG 4.0.1. 

```
mapserver\mapscript\csharp\swig_csharp_extensions.i(232): error : A deprecated csfinalize typemap was found for colorObj, please remove it and replace all csdestruct, csdestr
uct_derived and csfinalize typemaps by the csdispose, csdispose_derived, csdisposing and csdisposing_derived typemaps. 
```

Commenting out the following allows C# MapScript to build, but I'm unsure if it has any side effects. 
```
%typemap(csdestruct, methodname="Dispose", methodmodifiers="public") SWIGTYPE {
  lock(this) {
      if(swigCPtr.Handle != System.IntPtr.Zero && swigCMemOwn) {
        swigCMemOwn = false;
        $imcall;
      }
[...]

PR #5983, opened 2020-01-29
Updates to build MapScript with SWIG 4.0.1

geographika:
See issue #5982

- Python test suite now passes when built with SWIG 4.0.1 with this pull request. 
- Appveyor checks Python with both SWIG 3.0.12 (in the x32 build) and SWIG 4.0.1
- C# binding code commented out to successfully build, but I'm unsure if there are any unwanted side-effects
- PHP and Perl are built and tested on Travis so this will need updating to check these are working

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,17 +7,21 @@ image: Visual Studio 2017
 cache:
   - '%LOCALAPPDATA%\pip\Cache'
   - '%APPVEYOR_BUILD_FOLDER%\swigwin-3.0.12.zip'
-
+  - '%APPVEYOR_BUILD_FOLDER%\swigwin-4.0.1.zip'
+  
 environment:
   # VS 2017
   VS_VERSION: Visual Studio 15 2017
   matrix:
   - platform: x86
     PYTHON_EXECUTABLE: c:/python27/python.exe
+    SWIG_VERSION: swigwin-3.0.12
   - platform: x64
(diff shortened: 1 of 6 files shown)


geographika:
@szekerest - I've added a conditional statement to only remove the existing C# typemaps which cause SWIG 4 to fail when building with < SWIG 4. You may know better if these can be removed completely. 

@jmckenna reported on the mailing list that the PHP MapScript has been building successfully on SWIG 4 for sometime as part of ms4w. @tigerfoot @sebastic if you get a chance to review please let me know if these changes could cause any openSUSE / Ubuntu build problems. 

The Travis build currently uses SWIG 3.0.12 and all tests are passing so these updates at least are SWIG 3 compatible.

tigerfoot:
@geographika I've made a backport to 7.4.3 for python, and the build work nicely now so +1 on my side
( Can't test C# ) but java,perl,php,python mapscript modules are all build (with 7.4.3 + patches version)

rouault:
This needs to be rebased on latest master due to a conflict in bug_check.py

geographika:
@rouault - I've updated to the latest master (a rebase via TortoiseGit, not sure why the commits are showing up twice, but this pull request can be squashed). 
The `test_pattern` is a Python MapScript related test so I've moved to the MapScript test suite rather than part of msautotest. 
If you are happy with this I can merge?

rouault:
Looks good to me

PR #5993, opened 2020-02-19
Add basic WFS tests for the MSSQL driver

geographika:
Basic tests for WFS 2.0 GetFeature queries. 

Adding the tests also highlighted an issue where multiple ORDER BY statements were added to the SQL query causing a syntax error. This is corrected in the driver, and tested for. 

The tests cities.shp data also requires a .cpg file to correctly load the UTF-8 attributes into SQL Server.


jmckenna:
@geographika ok to merge this, to include in 7.6.0 release?

geographika:
@jmckenna - I would appreciate if anyone could have a quick review of just the C file https://github.com/mapserver/mapserver/pull/5993/files#diff-d4fe27d706e5a7dcb9fc808d413f8f60

Currently ORDER BY is duplicated if in the DATA clause if an additional sort is passed in. This change fixes that.

jmckenna:
Looks fine to me but @szekerest would be better for this review.

geographika:
@jmckenna - if you want to merge for the 7.6 release without a further review I'd be happy enough - the current code throws an error, and the new code and test passes.

jmckenna:
ok let's merge.  Great work and thank you @geographika !!!!

PR #5997, opened 2020-02-24
[WIP] Limit SQL Select to required fields

geographika:
A WFS LAYER can be configured with "gml_include_items" "all" to allow maximum flexibility for the client. A WFS request allows a propertyName parameter to be passed via the query string to only return data for the requested fields e.g. &propertyName=ogr_fid,name will return features with only these 2 attributes. This greatly reduces the amount of data transferred to the client. 
However the database query itself still selects data for all fields for the layer (confirmed in both the MSSQL and PostGIS drivers). 

Selecting only the propertyName fields in the query can have a huge effect on the database query time, especially for layers which have 10s of fields, but where each request may only require 5-6 fields at any one time. In my test case 13,000 records with 50 fields takes 12 seconds compared to running instantly for 5 fields. 

This is a proposed draft approach to limiting the number of fields used in a database select statement. As it currently is it has a number of issues (as reflected in the failing tests):

- The case-sensitivity of fields is affected
- The order fields are returned is affected
- Older style query templates miss data as they do not use the "gml_include_items" metadata value

Further discussions are in the [dev mailing list thread](https://lists.osgeo.org/pipermail/mapserver-dev/2020-February/016082.html). 

I also found this related RFC: https://mapserver.org/id/development/rfc/ms-rfc-33.html

### DIFF
--- a/maplayer.c
+++ b/maplayer.c
@@ -990,7 +990,7 @@ int msLayerWhichItems(layerObj *layer, int get_all, const char *metadata)
       for(i=0; i<n; i++) {
         bFound = 0;
         for(j=0; j<layer->numitems; j++) {
-          if(strcmp(tokens[i], layer->items[j]) == 0) {
+          if(strcasecmp(tokens[i], layer->items[j]) == 0) {
             bFound = 1;
             break;
           }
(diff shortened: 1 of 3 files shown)


geographika:
Cut back the scope of this change to try and limit selection only in the WFS case where propertyName is passed in from the querystring. 

2 issues currently outstanding:

- propertyName fields can be wrapped in brackets e.g. (name) so they are not picked up
- msGeometry is passed to propertyName in several tests, but is not a named field

Will continue looking into this.

rouault:
Superseded by #6131

PR #6000, opened 2020-02-27
Add missing expression test

geographika:
As per comment at https://github.com/mapserver/mapserver/pull/5897#issuecomment-591881359 - the expected result was missing (although shouldn't this have thrown an error?).


rouault:
> (although shouldn't this have thrown an error?).

Yes, that would be welcome. https://github.com/mapserver/mapserver/pull/6001 should detect this

rouault:
>  #6001 should detect this

yes, this missing file is now detected per https://travis-ci.org/mapserver/mapserver/jobs/655978020?utm_medium=notification&utm_source=github_status

PR #6001, opened 2020-02-27
msautotest: in --strict mode, as run by the CI, make absence of expected file an error

rouault:
Should address remark of https://github.com/mapserver/mapserver/pull/6000#issue-381070490


geographika:
Thanks @rouault - looks like my test was the only one missing a result then!

PR #6003, opened 2020-02-28
Allow both <LineSymbolizer> and <PolygonSymbolizer> in same <Rule>

jbo-ads:
This pull request adds support of a combination of `<LineSymbolizer>` and `<PolygonSymbolizer>` tags in same `<Rule>` section of a SLD style sheet. This is described in fifth section of [MS RFC 124: Improving SLD Support in MapServer](https://mapserver.org/development/rfc/ms-rfc-124.html#linesymbolizer-and-polygonsymbolizer-in-rule).


jmckenna:
thanks @jbo-ads !

Issue #6016, opened 2020-03-23
SLD <Label> implementation is non-conformant to OGC SLD spec


jbo-ads:
(For historians, this issue is a follow-up to issues #1857 and #1737)

As users around me start using improved SLD features in MapServer, they find some issues. One of them is related on how `<Label>` contents should be parsed:
- `<Literal>` is ignored
- Raw text is interpreted as implicit `<PropertyName>` instead of implicit `<Literal>`
- Mixted content, a mixture of raw text, `<Literal>`, `<PropertyName>`..., is handled incorrectly
- Users need implementation of `<Function name="round">`, with one argument as per SLD spec.

This issue is addressed by PR #6017.

PR #6017, opened 2020-03-23
SLD Improve <Label> conformance with OGC standard

jbo-ads:
This pull request addresses issue #6016.

Proposed changes are:
- Raw text in `<Label>` is interpreted as implicit `<Literal>` instead of implicit `<PropertyName>`. **This breaks previous MapServer behaviour**.
- `<Literal>`, `<PropertyName>` and `<Function>` tags are handled as well as any mixture of them.
- Regarding `<Function>` tag, all MapServer functions are available, although only the ones returning a number are expected to be functional. Moreover, a SLD-compliant `round(x)` function, with a single argument, has been added; it is equivalent to MapServer's `round(x,1)`.
- Note that arithmetic operations (`<Add>`, `<Sub>`, `<Mul>`, `<Div>`) are not taken into account, except inside a `<Function>` tag.

Associated documentation is submitted in Docs pull request [#318](https://github.com/mapserver/docs/pull/318).


rouault:
@jbo-ads This needs to be rebased on top of current master due to a conflict in mapparser.c

mapogcsld.c
```diff
@@ -1455,50 +1466,81 @@ int msSLDParseOgcExpression(CPLXMLNode *psRoot, void *psObj, int binding,
...
+        (*nexprbindings)++;
+        status = MS_SUCCESS;
+      }
       else if (strstr(ops, psRoot->pszValue)
           && psRoot->psChild && psRoot->psChild->psNext)
       {
         // Parse an arithmetic element <ogc:Add>, <ogc:Sub>, <ogc:Mul>, <ogc:Div>
         const char operator[2] = { *(strstr(ops, psRoot->pszValue)+3), '\0' };
-        char * expressionString = NULL;
+        msStringBuffer * expression = msStringBufferAlloc();
```
rouault, line 1521: may potentially leak if msStringBufferReleaseStringAndFree() is not reached

jbo-ads: Good catch, thank you!

Issue #6015, opened 2020-03-23
7.6.0-beta1 fails to build (error: unknown type name 'siginfo_t)


sebastic:
The Debian package build for MapServer 7.6.0-beta1 fails to build due to an issue with PHP MapScript NG:
```
[ 79%] Building C object mapscript/phpng/CMakeFiles/php_mapscriptng.dir/CMakeFiles/php_mapscriptng.dir/mapscriptPHP7_wrap.c.o
cd "/build/mapserver-7.6.0~beta1/obj-x86_64-linux-gnu/mapscript/phpng" && /usr/bin/cc -DPROJ_VERSION_MAJOR=7 -Dphp_mapscriptng_EXPORTS -I"/build/mapserver-7.6.0~beta1/obj-x86_64-linux-gnu/mapscript/phpng"  -I"/build/mapserver-7.6.0~beta1
/obj-x86_64-linux-gnu" -I"/build/mapserver-7.6.0~beta1/renderers/agg/include" -I"/build/mapserver-7.6.0~beta1/mapscript/v8" -I/usr/include/freetype2 -I"/build/mapserver-7.6.0~beta1/obj-x86_64-linux-gnu/renderers/mvt" -I/usr/include/fribi
di -I/usr/include/harfbuzz -I/usr/include/cairo -I/usr/include/postgresql -I/usr/include/gdal -I/usr/include/librsvg-2.0 -I/usr/include/gdk-pixbuf-2.0 -I/usr/include/libmount -I/usr/include/blkid -I/usr/include/glib-2.0 -I/usr/lib/x86_64
-linux-gnu/glib-2.0/include -I/usr/include/pixman-1 -I/usr/include/uuid -I/usr/include/libpng16 -I/usr/include/libxml2 -I/usr/include/php/20190902 -I/usr/include/php/20190902/main -I/usr/include/php/20190902/TSRM -I/usr/include/php/20190
902/Zend -I/usr/include/php/20190902/ext -I/usr/include/php/20190902/ext/date/lib -I"/build/mapserver-7.6.0~beta1/mapscript/swiginc" -I"/build/mapserver-7.6.0~beta1/mapscript" -I"/build/mapserver-7.6.0~beta1/mapscript/phpng" -I"/build/ma
[...]

PR #6018, opened 2020-03-23
handle gnu_source definition for SWIG mapscript (issue #6015)

jmckenna:


mapscript/mapscript.i
```diff
@@ -33,6 +33,8 @@
 %begin %{
 #ifdef _MSC_VER
 #define SWIG_PYTHON_INTERPRETER_NO_DEBUG
+#else
+#define _GNU_SOURCE 1
```
sebastic: Just `#define _GNU_SOURCE` is enough, it doesn't need a value.

jmckenna: I did at first but it throws several 'redefined' warnings.  Giving it a value prevents the warnings.

sebastic: Then you probably shouldn't define it unconditionally, e.g.:
```C
#ifdef _MSC_VER
#define SWIG_PYTHON_INTERPRETER_NO_DEBUG
#elif !defined(_GNU_SOURCE)
#define _GNU_SOURCE
#endif
```

jmckenna: I had tried that earlier as well.  I tried again, still gives warning "/../main/php_config.h:2535:0: warning: _GNU_SOURCE" redefined".

jmckenna: With all my testing I found the best way was the pass it a value, to remove the redefinition warnings in PHP source.

sebastic: Alright, then just leave it as-is.

jmckenna: But I'm open to other ways.  Let me know.

sebastic: Looking at [codesearch.debian.net](https://codesearch.debian.net/search?q=%23define+_GNU_SOURCE&literal=0&perpkg=1&page=4) there a quite a number of occurences with value, so either seems acceptable.

jmckenna: ok thanks

Issue #6021, opened 2020-03-26
Rendering bug with both RENDERMODE and GEOMTRANSFORM options


jbo-ads:
I want to draw a polygon with one class for selecting its fill and outline colors and another class for placing a symbol at its centre. I use `PROCESSING "RENDERMODE=ALL_MATCHING_CLASSES"` at layer level for rendering both classes and `GEOMTRANSFORM "centroid"` at style level for rendering a single symbol instead of a filling pattern.

Unfortunately, the symbol is not rendered.

This issue is addresed by pull request #6022.

jmckenna:
reviewing and testing proposed change...

jmckenna:
Confirmed the problem with 7.6.0-beta1, and confirmed the proposed pull request.

Here was my local mapfile that I tested with:
```
MAP
  NAME "rendermode-test"
  STATUS ON
  EXTENT -64.25868122196320087 44.35950418003932327 -64.25674251814376703 44.35987545845821955
  SIZE 500 500
  IMAGETYPE png
  IMAGECOLOR 255 255 255
  UNITS DD
  SHAPEPATH "./"
  
  OUTPUTFORMAT
    NAME "png"
    DRIVER AGG/PNG
    MIMETYPE "image/png"
    IMAGEMODE RGB
    EXTENSION "png"
    FORMATOPTION "GAMMA=0.75"
  END
  
  SYMBOL
    NAME "circle"
    TYPE ELLIPSE
    FILLED TRUE
    POINTS
      1 1
    END
  END #symbol  
  
  LAYER
    NAME        "test"
    STATUS      ON
    TYPE        POLYGON
    PROCESSING  "ITEMS=name"
    PROCESSING  "RENDERMODE=ALL_MATCHING_CLASSES"
    FEATURE
      POINTS
        -64.2577617 44.3597553 
        -64.2575793 44.3596671
        -64.2576477 44.3595952
        -64.2578261 44.3596911
        -64.2577617 44.3597553
      END
      ITEMS "building"
    END
    CLASSITEM "name"
    CLASS
      NAME       "class1"
      STYLE
        COLOR    0 255 0
        OUTLINECOLOR 150 150 150
        WIDTH    200
      END
    END
    CLASS
      NAME       "class2"
      STYLE
        SYMBOL "circle"
        GEOMTRANSFORM "centroid"
        COLOR 255 255 255
        OUTLINECOLOR 0 0 0
        SIZE 20
      END
    END
	
  END #layer
  
[...]

PR #6022, opened 2020-03-26
Fix rendering bug with both RENDERMODE and GEOMTRANSFORM options

jbo-ads:
This pull request addresses issue #6021.


jmckenna:
thank you for this fix @jbo-ads !!!

PR #6028, opened 2020-04-01
Minimal C++ification of mapwms.c, mapwcs.c, mapwcs11.c, mapwcs20.c, mapwfs.c and mapwfs11.c

rouault:


rouault:
Changing to 8.0 as I think this is a more reasonable target

jmckenna:
7.6.1 target date is 31 May, but I'm fine with 8.0 target.

PR #6037, opened 2020-04-04
Use GDAL large file API (UTF-8) compatible in a number of places

### LINKED ISSUE #5995: DATA filenames with umlauts & relative path don't work, on Windows
I figured out problems with raster filenames containing umlauts.

To reproduce it, just call

shp2img -m mypath/umlauttest.map -o myImg.png

I tracked it down to the function msTryBuildPath3 and the fopen call with UTF-8 chars.
Every fopen call with  UTF-8 chars ( 2 or more bytes) will fail on windows.
That seems to be a general problem, not only with rasters.
The gdal source handles this in another way, e.g. 

VSIWin32FilesystemHandler::Open
...
 if( CPLTestBool(CPLGetConfigOption( "GDAL_FILENAME_IS_UTF8", "YES" ) ) )
    {
        wchar_t *pwszFilename =
            CPLRecodeToWChar( pszFilename, CPL_ENC_UTF8, CPL_ENC_UCS2 );

        hFile = CreateFileW( pwszFilename, dwDesiredAccess,
                            bShared ? FILE_SHARE_READ | FILE_SHARE_WRITE : 0,
                            nullptr, dwCreationDisposition,  dwFlagsAndAttributes,
                            nullptr );
   ...

[umlauttest.zip](https://github.com/mapserver/mapserver/files/4234943/umlauttest.zip)

rouault:
There are still uses of ANSI FILE* API, but this should hopefully
fix #5995, but untested

Likely not appropriate for backport

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -282,7 +282,7 @@ mapraster.c mapuvraster.c mapdummyrenderer.c mapobject.c maprasterquery.c
 mapwcs.cpp maperror.c mapogcfilter.c mapregex.c mapwcs11.cpp mapfile.c
 mapogcfiltercommon.cpp maprendering.c mapwcs20.cpp mapogcsld.c mapmetadata.c
 mapresample.c mapwfs.cpp mapgdal.c mapogcsos.c mapscale.c mapwfs11.cpp mapwfs20.c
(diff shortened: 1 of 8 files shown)


jmckenna:
my initial umlaut tests through #5995 on Windows succeed, after this change.  Thank you @rouault !!  wow great work.  Will report in other ticket (where likely more testing can occur).

PR #6039, opened 2020-04-06
WMS GetMap/GetFeatureInfo: make STYLES parameter mandatory as per WMS spec (fixes #6012)

### LINKED ISSUE #6012: MapServer treats WMS STYLES param as optional, but the WMS spec says it is mandatory
We got a report that MapServer treats the WMS GetMap STYLES param as optional, but the WMS spec state that it should be mandatory. (I verified and GetMap STYLES is indeed mandatory in WMS 1.1.0, 1.1.1 and 1.3.0)

One could say that this is no big deal, MapServer is just being a bit more friendly than it should, and it's been like this since the initial implementation almost 20 years ago anyway. But actually the side-effect here is that we have server logs showing that there are some clients in the wild that were probably tested only against MapServer and do not pass the STYLES param. Those WMS clients will break if we switch MapServer to another WMS service that enforces the rule. (We are working to place a MapProxy cache in front of an existing MapServer instance, and we found out about this because MapProxy enforces the mandatory STYLES param rule).

There is not much we can do about the past, but what do we want to do about this in MapServer for the future?  

- Option 1, Fix it: Do we want to fix this and make WMS GetMap STYLES mandatory in MapServer in the 8.0 release, which may break existing custom WMS clients that were only tested against MapServer?

- Option 2, Don't touch it:  Or do we prefer to keep the STYLES param as optional in MapServer, which won't break existing clients, but may contribute to the deployment of more broken clients that assume that STYLES is optional if tested only against MapServer?

rouault:
Only for baseline requests: if SLD specific parameters SLD & SLD_BODY are
specified, then do not make STYLES mandatory

If the STYLES parameter is missing and needed, the error message emitted
suggests to define the "wms_allow_getmap_without_styles" "true" MAP.WEB.METADATA
item to get backward compatible behaviour.


jmckenna:
thanks for this @rouault

Issue #5942, opened 2019-12-04
Parse errors within INCLUDEd files leak file handles


alexbrault:
When the parser encounters an error, it immediately exits its loop, closes the current file and returns. If that happens while parsing a file included with the INCLUDE token, all the other files in the include stack are left open. This is especially a problem in MapScript since those files may not be reclaimed by the operating system until the parent process exits.

For example:
leaked.map:
```
map
include 'parse_error.map'
end
```

parse_error.map:
```
invalid token
```

dmorissette:
Reopening... this is related to #5941, but PR#5944 does not address this, my mistake.  The fix will be more involved as it is related to generic handling of INCLUDEs in the parser.

@sdlime, you may be the best candidate to address this one.

jmckenna:
@sdlime, as @dmorissette said, can you add this onto your plate?  I think this one is important.

sdlime:
Yes, will try to find time this week.

sdlime:
I have a pull request out there that aims to address this in master but it needs to be verified. @alexbrault, any chance you could take a look? I've tried with parse errors in the main map file, one include deep and two includes deep and the right code seems to be being called.

--Steve

PR #6050, opened 2020-04-15
Addresses open file handles when parse errors occur in include files  (Issue #5942)

sdlime:


rouault:
An addition in msautotest might be useful

sdlime:
A specific test would need to take the form of something that would show the number of open file handles after a process closed. I'm not sure how that would be done or how it could be included in the test suite.

rouault:
> A specific test would need to take the form of something that would show the number of open file handles after a process closed

I had presumed that the CI config that runs with Address Sanitizer enabled (PHP_7.2_WITH_ASAN) would warn about unclosed files as a memory leak, but on a small test case, I see this isn't the case unfortunately. Valgrind can see this with --leak-check=full --show-leak-kinds=all however
```
==27735== 552 bytes in 1 blocks are still reachable in loss record 1 of 1
==27735==    at 0x4C2DB8F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==27735==    by 0x4EA7CDC: __fopen_internal (iofopen.c:69)
==27735==    by 0x40053C: main (leak.c:4)
```
However I think there's still a merit in having such a test case: at least this makes sure that the error code paths are hit and don't cause crash (I'm sure they don't now :-), but it might if someone later touches it)

sdlime:
I can certainly create a couple of scenarios that generate parse errors 0, 1 and 2 includes deep.

sdlime:
Bump, @dmorissette or @alexbrault, do you have time to test this fix? --Steve

jmckenna:
No response on testing, but am merging this now...

PR #6061, opened 2020-04-29
Wms essential

umeier:
Pull-request re-done to master as requested by @jmckenna

### DIFF
--- a/mapwmslayer.c
+++ b/mapwmslayer.c
@@ -1255,7 +1255,7 @@ int msDrawWMSLayerLow(int nLayerId, httpRequestObj *pasReqInfo,
   if ( !MS_HTTP_SUCCESS( pasReqInfo[iReq].nStatus ) ) {
     /* ====================================================================
           Failed downloading layer... we log an error but we still return
-          SUCCESS here so that the layer is only skipped intead of aborting
+          SUCCESS here so that the layer is only skipped instead of aborting
           the whole draw map.
           If the layer is essential the map is not to be drawn.
      ==================================================================== */
@@ -1275,7 +1275,7 @@ int msDrawWMSLayerLow(int nLayerId, httpRequestObj *pasReqInfo,
    * Check the Content-Type of the response to see if we got an exception,
    * if yes then try to parse it and pass the info to msSetError().
    * We log an error but we still return SUCCESS here so that the layer
-   * is only skipped intead of aborting the whole draw map.
+   * is only skipped instead of aborting the whole draw map.
    * If the layer is essential the map is not to be drawn.
    * ------------------------------------------------------------------ */
   if (pasReqInfo[iReq].pszContentType &&
--- a/mapwmslayer.c
+++ b/mapwmslayer.c
@@ -1224,6 +1224,18 @@ int msDrawWMSLayerLow(int nLayerId, httpRequestObj *pasReqInfo,
   int currentconnectiontype;
   int numclasses;
   char *mem_filename = NULL;
+  const char *pszTmp;
+  int bIsEssential;
+
+  /* ------------------------------------------------------------------
+   * Sometimes a requested layer is essential for the map, so if the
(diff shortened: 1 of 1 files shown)


jmckenna:
thanks @umeier !

jmckenna:
@umeier I'm testing this with the following mapfile, which makes a call to a WMS service that will fail, but the map is still drawn and no error is thrown at the commandline.  Here is the test mapfile (if you change the wms_format to "image/png" it will work:
```
MAP
  NAME "essential-test"
  STATUS ON
  SIZE 400 300
  EXTENT -180 -90 180 90
  UNITS DD
  IMAGECOLOR 255 255 255

  LAYER
   NAME "country_bounds"
    TYPE RASTER
    STATUS ON
    CONNECTION "https://demo.mapserver.org/cgi-bin/wms?"
    CONNECTIONTYPE WMS
    METADATA
      "wms_srs"             "EPSG:4326"
      "wms_name"            "country_bounds"
      "wms_server_version"  "1.1.1"
      "wms_format"          "image/gif"
      "wms_essential"       "1"
    END
  END

END
```

Here is my command:
```
shp2img -m wms-client-essential.map -o ttt.png -map_debug 3
```
and here is the response:
```
msDrawMap(): rendering using outputformat named png (AGG/PNG).
msDrawMap(): WMS/WFS set-up and query, 0.704s
msDrawRasterLayerLow(country_bounds): entering.
msDrawMap(): Layer 0 (country_bounds), 0.022s
msDrawMap(): Drawing Label Cache, 0.000s
msDrawMap() total time: 0.727s
msSaveImage(ttt.png) total time: 0.015s
```

jmckenna:
Ok, so it seems "wms_essential" could be a little misleading, as if you pass incorrect layer name ("wms_name") or image format ("image/gif") no error is thrown, but if you pass a broken connection url it will properly return an error.  I guess I will have to add this to the docs.

for example, this incorrect url properly fails with wms_essential:
```
CONNECTION "https://demo.mapserver.org/cgi-bin/wms2?"
```

rouault:
> as if you pass incorrect layer name ("ttt") or image format ("image/gif") no error is thrown

Running with -all_debug 5 shows the following request:
```
HTTP request: id=0, https://demo.mapserver.org/cgi-bin/wms?LAYERS=country_bounds&REQUEST=GetMap&SERVICE=WMS&FORMAT=image/gif&STYLES=&HEIGHT=300&VERSION=1.1.1&SRS=EPSG:4326&WIDTH=400&BBOX=-180.451127819549,-135.338345864662,180.451127819549,135.338345864662&TRANSPARENT=TRUE&EXCEPTIONS=application/vnd.ogc.se_inimage
```
The important thing is the EXCEPTIONS=application/vnd.ogc.se_inimage . I guess that in "essential" mode this shouldn't be added. Manually adding "wms_exceptions_format" "application/vnd.ogc.se_xml" fixes the issue, but should probably be the default in essential mode (or do not emit EXCEPTIONS at all since XML is the default)

jmckenna:
> The important thing is the EXCEPTIONS=application/vnd.ogc.se_inimage . I guess that in "essential" mode this shouldn't be added.

A working response that returns a valid map image will also include `EXCEPTIONS=application/vnd.ogc.se_inimage`, for example here is the same mapfile with the correct CONNECTION and "wms_format" which returns the proper map image yet does contain the EXCEPTIONS string :

https://demo.mapserver.org/cgi-bin/wms?LAYERS=country_bounds&REQUEST=GetMap&SERVICE=WMS&FORMAT=image/png&STYLES=&HEIGHT=300&VERSION=1.1.1&SRS=EPSG:4326&WIDTH=400&BBOX=-180.451127819549,-135.338345864662,180.451127819549,135.338345864662&TRANSPARENT=TRUE&EXCEPTIONS=application/vnd.ogc.se_inimage

jmckenna:
Ok I think I misunderstood @rouault sorry about that.  I see what you mean.

PR #6065, opened 2020-05-07
for essential WMS layers set EXCEPTIONS to XML by default

jmckenna:
- related to mapserver/mapserver#6061
 - sets EXCEPTIONS to XML for essential WMS layers, specifically for the GetMap request

### DIFF
--- a/mapwmslayer.c
+++ b/mapwmslayer.c
@@ -428,7 +428,9 @@ msBuildWMSLayerURL(mapObj *map, layerObj *lp, int nRequestType,
   int nVersion=OWS_VERSION_NOTSET;
   int bUseStrictAxisOrder = MS_FALSE; /* this is the assumption up to 1.1.0 */
   int bFlipAxisOrder = MS_FALSE;
-
+  const char *pszTmp;
+  int bIsEssential;
+  
   if (lp->connectiontype != MS_WMS) {
     msSetError(MS_WMSCONNERR, "Call supported only for CONNECTIONTYPE WMS",
                "msBuildWMSLayerURL()");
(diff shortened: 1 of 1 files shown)


mapwmslayer.c
```diff
@@ -874,11 +886,25 @@ msBuildWMSLayerURL(mapObj *map, layerObj *lp, int nRequestType,
...
+      if (pszExceptionsParam == NULL) {
+        if (nVersion >= OWS_1_1_0 && nVersion < OWS_1_3_0)
+          pszExceptionsParam = "application/vnd.ogc.se_inimage";
+        else
+          pszExceptionsParam = "INIMAGE";
+      }
+    }
+    /* if layer is essential, set EXCEPTIONS parameter to XML */    
+    else {                        
+      if (nVersion >= OWS_1_3_0)
```
rouault, line 900: I've not reviewed earlier WMS specs, but at least for WMS 1.1.1 and 1.3.0, the default value for EXCEPTIONS when it is not specified is to be in XML format. So a simpler possibility, that would save if else loginc on the WMS version, would be not to emit EXCEPTIONS= at all.

jmckenna: This same check occurs earlier in the file.

jmckenna: This pull request follows your advice in #6061 where you gave 2 options for the solution.

jmckenna: Sorry for making it complicated, I definitely tried.  I have tested this on multiple systems; that said, I'm open to changes/new pull request/closing this and re-opening...

rouault: no need to close&recreate a pull request to update it. I've just added a commit with the code simplification I add in mind. Can you pull and confirm this works OK ?

jmckenna: confirmed, your simplification still returns the XML exception (instead of image).  Thank you again @rouault

mapwmslayer.c
```diff
@@ -785,10 +785,14 @@ msBuildWMSLayerURL(mapObj *map, layerObj *lp, int nRequestType,
...
-  bIsEssential = MS_FALSE;
   if ((pszTmp = msOWSLookupMetadata(&(lp->metadata),
                                     "MO", "essential")) != NULL) {
-    bIsEssential = atoi(pszTmp);
+    if( strcasecmp(pszTmp,"true") == 0
+        || strcasecmp(pszTmp,"on") == 0
+        || strcasecmp(pszTmp,"yes") == 0 )
+      bIsEssential = MS_TRUE;
+    else
+      bIsEssential = atoi(pszTmp);
```
jmckenna, line 795: this follows other settings as "cache_to_disk"

PR #6069, opened 2020-05-16
Remove submodule name from mapscript package

geographika:
This pull request removes the submodule name from the class names. 

Currently all classes are imported into the package root namespace in `__init__.py`. 
However the class module names are still set to the subpackage. This change will make them appear in the root, based on the answer at https://stackoverflow.com/questions/1808522/masquerading-real-module-of-a-class/

Before:

```python
>>> import mapscript
>>> mapscript.classObj.__module__
'mapscript.mapscript'
```

After:

```python
>>> mapscript.classObj.__module__
'mapscript'
```

For Python MapScript users nothing will change. The main benefit is for the ongoing autodoc work with Sphinx.

### DIFF
--- a/mapscript/python/mapscript/__init__.py
+++ b/mapscript/python/mapscript/__init__.py
@@ -1,1 +1,12 @@
+import inspect
 from .mapscript import *
+
+# change all the class module names from mapscript.mapscript to mapscript
+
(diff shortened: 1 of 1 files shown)


jmckenna:
thanks @geographika !

PR #6071, opened 2020-05-19
Add classObj property strings and update MapScript docstrings

geographika:
This pull request builds on https://github.com/mapserver/docs/pull/327 and allows direct links from the planned generated MapScript docs to Mapfile keyword syntax. 

It also adds property docstrings to the classObj in mapserver.h
Initially I added these as Python only docstrings [here](https://github.com/mapserver/mapserver/pull/6071/commits/19d40f2a4572e3eba76e57f1dabc1a9d788114e4), but using the SWIG feature and adding to where classObj is defined allows types to be automatically generated for each property. 

Keeping the docstrings and property declarations together also has more chance of them being maintained.

### DIFF
--- a/mapscript/mapscript.i
+++ b/mapscript/mapscript.i
@@ -251,12 +251,18 @@ typedef struct {
 =============================================================================
 */
 
+%feature("autodoc"); // autodoc all struct properties in the header files
+
 %include "../../mapserver.h"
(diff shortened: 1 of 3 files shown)


jmckenna:
thanks @geographika ! Hope it was ok to merge now :)       Enjoy your weekend!

PR #6073, opened 2020-05-23
Deploy MapScript to PyPI for any release tag

geographika:
This automates deploying to PyPI for any release tag (previously the PS script had assumed this would be on a master branch, and would have deployed to the test PyPI site). 

Release tags are assumed to be in the format `"rel-7-6-0"` which should exclude any release candidates and betas from being deployed. 

```
("rel-7-6-0" -cSplit "-").Count -eq 4
("rel-7-6-0-rc4" -cSplit "-").Count -eq 5
```

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -106,11 +106,14 @@ artifacts:
 
 deploy_script:
   - ps: |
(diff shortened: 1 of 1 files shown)


jmckenna:
thanks @geographika

Issue #6076, opened 2020-05-25
msMSSQL2008LayerGetExtent doesn't consider NATIVE_FILTER or translated filter


szekerest:
The postgis and the oracle drivers limit the extent query within the specified filter (either by using NATIVE_FILTER or translated filter). We should also establish the same option for the mssql driver.

Also msMSSQL2008LayerGetExtent  causes a crash if the queries return null values. ie. when SQLGetData return -1 in the retLen parameter.

PR #6078, opened 2020-05-27
msMSSQL2008LayerGetExtent doesn't consider NATIVE_FILTER or translated filter (#6076)

szekerest:


jmckenna:
thanks @szekerest !

Issue #6083, opened 2020-06-02
Relative URL on SLD ExternalGraphic doesn't work


jbo-ads:
Currently it is not possible to specify a relative URL for a SVG external graphic in SLD, e.g.:

    <ExternalGraphic>
        <OnlineResource xlink:type="simple" xlink:href="assets/foo.svg" />
        <Format>image/svg+xml</Format>
    </ExternalGraphic>

If for instance SLD file is located at <http://example.com/path/to/sld.xml>, that relative URL should resolve to <http://example.com/path/to/assets/foo.svg>. Instead the symbol is not found and thus not rendered.

PR #6084, opened 2020-06-02
Enable relative URL on SLD ExternalGraphic

jbo-ads:
This pull request addresses issue #6083.

Three cases are handled:
- Symbol URL is absolute: *http://example.com/path/to/assets/foo.svg*.
- Symbol URL is relative to SLD URL: *http://example.com/path/to/sld.xml* + *assets/foo.svg*
- Symbol URL is relative to root of SLD server: *http://example.com/path/to/sld.xml* + */path/to/assets/foo.svg*

When SLD is passed as SLD_BODY inline document, symbol URL is considered absolute.


jratike80:
Are there any limits in the relative path or is it possible to follow the path to root and anywhere?

jbo-ads:
@jratike80 The only constraint is the one already existing about *sld_external_graphic* validation, see [Migration guide from 6.4 to 7.0](https://mapserver.org/MIGRATION_GUIDE.html?highlight=sld_external_graphic#mapserver-6-4-to-7-0-migration).

The validation of symbol URL against *sld_external_graphic* entry is done on the resulting URL.

PR #6085, opened 2020-06-04
Development

botulf2000:
A collection of fixes done at Sokigo-GLS


rouault:
This would need to be split in several pull requests for each topic

szekerest:
Also an assigned issue for each pull rq would be required with a brief description what we would like to achieve with the changes.

jmckenna:
Closing, as these have or will be handled by separate pull requests.

Issue #5529, opened 2017-12-02
msGEOSError(): GEOS library error when using MULTILINESTRING


gertcuykens:
msGEOSError(): GEOS library error when using MULTILINESTRING in shapefiles

Try `shp2img -m test.map -o test.png -l test -all_debug 5 -s 500 500`
on any shapefile that has MULTILINESTRING in it

example shapefile feature and using mapserver 7.0.7
```
OGRFeature(test):178

  color (String) = #c6b952

  MULTILINESTRING ((1365999.913051389390603
6668002.688304415903986,1365929.670452698832378
6667981.874043367803097),(1366036.091885895933956
6668063.352393131703138,1365999.913051389390603
6668002.688304415903986),(1366004.811108984053135
6668127.219178147614002,1366036.091885895933956
6668063.352393131703138),(1365934.123232331126928
6668138.427049811929464,1366004.811108984053135
6668127.219178147614002),(1365862.878758222097531
6668117.612443232908845,1365934.123232331126928
6668138.427049811929464),(1365792.190881569404155
6668094.485164736397564,1365862.878758222097531
6668117.612443232908845),(1365384.093628322007135
6667942.914158684201539,1365306.949221200309694
6667931.172859975136817),(1365461.460674423491582
6667952.342789551243186,1365384.093628322007135
6667942.914158684201539),(1365538.939040013123304
6667961.771431289613247,1365461.460674423491582
6667952.342789551243186),(1365616.528725096955895
6667969.065293673425913,1365538.939040013123304
6667961.771431289613247),(1365773.266568133374676
6668049.120308448560536,1365760.910104657290503
6668084.344763254746795),(1365771.819414753001183
6668087.724895685911179,1365792.190881569404155
6668094.485164736397564),(1365760.910104657290503
6668084.344763254746795,1365771.819414753001183
6668087.724895685911179),(1365694.786327125271782
6667973.512773931957781,1365616.528725096955895
6667969.065293673425913),(1365773.043929153587669
6667977.426558562554419,1365694.786327125271782
6667973.512773931957781),(1365851.301531179109588
6667980.272948558442295,1365773.043929153587669
6667977.426558562554419),(1365929.670452698832378
6667981.874043367803097,1365851.301531179109588
6667980.272948558442295))
```

sdlime:
Can you share your test files? --Steve

gertcuykens:
[...]

PR #6086, opened 2020-06-04
Fix issue #5529 about GEOS error with offset on MULTILINESTRING

jbo-ads:


jmckenna:
@jbo-ads I also have the zip test here, will test on Windows now and report in ticket....

sdlime:
A comment against #5529 said this was still an issue w/7.6.4.

PR #6088, opened 2020-06-04
Don't use CIRCULARSTRING when fetching geography data from SQL Server.

botulf2000:
Use CURVEPOLYGON(()) instead of CURVEPOLYGON(CIRCULARSTRING()) when fetching geography data from SQL Server. This will increase the performance.


jmckenna:
thanks @botulf2000

PR #6089, opened 2020-06-04
Fixed rounding error in WMS client

botulf2000:
When using a tile size of 256 in Web Mercator the cascaded WMS request could be computed to a height or width of 257.


rouault:
Rebased on top of latest main in https://github.com/MapServer/MapServer/pull/6252 . @wicksell @botulf2000 Anything you had in mind to put that pull request as Draft ?

PR #6094, opened 2020-06-12
Fix eol of mapscript python examples to LF instead of CRLF

tigerfoot:
Signed-off-by: Bruno Friedmann <bruno@ioda-net.ch>
During packaging on openSUSE rpmlint warn about wrong eol used making script unusable under Linux. Simple fix done with ```sed -i 's,\r,,g' *.py``` 
Perhaps a test where wc -l shoud be always 0 can be installed to avoid this in the future ?
/usr/bin/find mapscript/*/examples/ -iname "*.*" -exec /usr/bin/file {} \; | /usr/bin/grep "CRLF" | wc -l


jmckenna:
thanks @tigerfoot !! (and @geographika for review!)

PR #6103, opened 2020-06-23
Hide some SWIG Java warnings

geographika:
Currently the Travis build with optimisations has a 15,000 line log file: https://travis-ci.org/github/mapserver/mapserver/jobs/699304461 = lines 15357

Many of these are warnings related to the Java MapScript bindings e.g.:

` /home/travis/build/mapserver/mapserver/build/mapscript/java/edu/umn/gis/mapscript/mapscriptJAVA_wrap.c:12271:19: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
`

This is caused by the optimisation flag:

`make cmakebuild MFLAGS="-j2" CMAKE_C_FLAGS="-O2...`

See https://github.com/mapserver/mapserver/blob/master/ci/travis/script.sh#L30

A similar issue arose at https://github.com/microsoft/LightGBM/issues/1556

See SWIG notes at http://www.swig.org/Doc4.0/Java.html#Java_compiling_dynamic

> Important
> If you are going to use optimisations turned on with gcc (for example -O2), ensure you also compile with -fno-strict-aliasing. The GCC optimisations have become more aggressive from gcc-4.0 onwards and will result in code that fails with strict aliasing optimisations turned on. See the C/C++ to Java typemaps section for more details. 

Hiding these warnings takes 5,000 lines off the log which makes it easier to see other errors. E.g. https://travis-ci.org/github/geographika/mapserver/jobs/701402353

### DIFF
--- a/mapscript/java/CMakeLists.txt
+++ b/mapscript/java/CMakeLists.txt
@@ -15,7 +15,7 @@ set (CMAKE_SWIG_OUTDIR "${CMAKE_CURRENT_BINARY_DIR}/edu/umn/gis/mapscript")
 set(CMAKE_SWIG_FLAGS -package edu.umn.gis.mapscript)
 
 # hide warnings
-if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_COMPILER_IS_GNUCC OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID STREQUAL "Clang")
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fno-strict-aliasing")
 endif ()
 
--- a/mapscript/java/CMakeLists.txt
+++ b/mapscript/java/CMakeLists.txt
@@ -4,7 +4,7 @@ include(${SWIG_USE_FILE})
 find_package(JNI)
 find_package(Java)
 if(NOT JNI_INCLUDE_DIRS OR NOT Java_JAVAC_EXECUTABLE OR NOT Java_JAR_EXECUTABLE)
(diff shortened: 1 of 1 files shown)


mapscript/java/CMakeLists.txt
```diff
@@ -14,6 +14,9 @@ include_directories(${PROJECT_SOURCE_DIR}/mapscript/java)
 set (CMAKE_SWIG_OUTDIR "${CMAKE_CURRENT_BINARY_DIR}/edu/umn/gis/mapscript")
 set(CMAKE_SWIG_FLAGS -package edu.umn.gis.mapscript)
 
+# hide warnings
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fno-strict-aliasing")
```
rouault, line 18: This is a gcc specific switch. Should probably be protected by (untested)
```suggestion
if (CMAKE_COMPILER_IS_GNUCC OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fno-strict-aliasing")
endif()
```

geographika: Should CMAKE_COMPILER_IS_GNUCXX also be checked for? 
I've added the same approach as https://github.com/mapserver/mapserver/blob/master/CMakeLists.txt#L188

rouault: Testing CMAKE_COMPILER_IS_GNUCXX is for the C++ compiler, so not relevant when setting CMAKE_C_FLAGS
But thinking about that the test for clang with CMAKE_CXX_COMPILER_ID should be done with CMAKE_C_COMPILER_ID

Issue #5708, opened 2018-11-24
MapScript PHPNG adding false build targets to Python MapScript


geographika:
The following code in the mapscript.i file:

```
#ifdef SWIGPHPNG
%module mapscriptng
#else
%module mapscript
#endif
```

Causes the following to be added to `build/mapscript/python/CMakeFiles/_pythonmapscript.dir/build.make`

```
mapscript/python/mapscriptng.py: mapscript/python/mapscriptPYTHON_wrap.c
	@$(CMAKE_COMMAND) -E touch_nocreate mapscript/python/mapscriptng.py 
```

As this file is never generated it has the following knock-on effect of generating a dependency that is never created in `/build/mapserver/build/mapscript/python/CMakeFiles/_pythonmapscript.dir/DependInfo.Cmake`
   
```
# Pairs of files generated by the same build rule.
set(CMAKE_MULTIPLE_OUTPUT_PAIRS
  "/build/mapserver/build/mapscript/python/mapscriptng.py" "/build/mapserver/build/mapscript/python/mapscriptPYTHON_wrap.c"
  )
```

This causes the Python MapScript bindings to be rebuilt every time `make` or `make install` is called as shown in the logs (for future reference you can see what triggered the rebuild using `make install VERBOSE=1`:

```
make[2]: Entering directory '/build/mapserver/build'
cd /build/mapserver/build && /usr/bin/cmake -E cmake_depends "Unix Makefiles" /build/mapserver /build/mapserver/mapscript/python /build/mapserver/build /build/mapserver/build/mapscript/python /build/mapserver/build/mapscript/python/CMakeFiles/_pythonmapscript.dir/DependInfo.cmake --color=
Deleting primary custom command output "/build/mapserver/build/mapscript/python/mapscriptPYTHON_wrap.c" because another output "/build/mapserver/build/mapscript/python/mapscriptng.py" does not exist.
```

Using the following command - `grep -iRl "mapscriptng" ./` the only place `mapscriptng` is mentioned is in mapscript.i

I'm not sure on how everything fits together, or why this target is generated by CMake. Changing it to the following fixes the issue in Python, but may break the PHPNG build:
```
%module mapscript
#ifdef SWIGPHP
%module mapscriptng
#endif
```

@AlexanderGabriel - maybe you have an idea on what causes this?

AlexanderGabriel:
Hi @geographika ,

sure, you are in the right branch?
[...]

PR #6104, opened 2020-06-27
install PHPNG required file 'mapscript.php'

jmckenna:
- fixes reports of Debian-GIS not including the required file 'mapscript.php' (https://bugs.launchpad.net/ubuntu/+source/mapserver/+bug/1884854)
 - will install the file into the PHP_EXTENSTION_DIR, such as:

     `-- Installing: /usr/lib/php/20170718/mapscript.php`

### DIFF
--- a/mapscript/mapscript.i
+++ b/mapscript/mapscript.i
@@ -38,7 +38,7 @@
 #endif
 %}
 
-#ifndef SWIGPHPNG
+#ifndef SWIGPHP7
 %module mapscript
 #else
 %module mapscriptng
--- a/mapscript/mapscript.i
+++ b/mapscript/mapscript.i
@@ -38,7 +38,7 @@
 #endif
 %}
 
-#ifndef SWIGPHP7
+#ifndef SWIGPHPNG
 %module mapscript
 #else
 %module mapscriptng
--- a/mapscript/phpng/CMakeLists.txt
+++ b/mapscript/phpng/CMakeLists.txt
@@ -84,4 +84,6 @@ set_target_properties(${SWIG_MODULE_mapscript_REAL_NAME} PROPERTIES PREFIX "")
 
 if(NOT WIN32)
   install(TARGETS php_mapscriptng DESTINATION ${PHP_EXTENSION_DIR})
+  # install the required file containing MapServer constants and functions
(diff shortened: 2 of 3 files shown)


sebastic:
The file content is wrong:
```
// Try to load our extension if it's not already loaded.
if (!extension_loaded('mapscript')) {
  if (strtolower(substr(PHP_OS, 0, 3)) === 'win') {
    if (!dl('php_mapscript.dll')) return;
  } else {
    // PHP_SHLIB_SUFFIX gives 'dylib' on MacOS X but modules are 'so'.
    if (PHP_SHLIB_SUFFIX === 'dylib') {
      if (!dl('mapscript.so')) return;
    } else {
      if (!dl('mapscript.'.PHP_SHLIB_SUFFIX)) return;
    }
  }
}
```
It should load `php_mapscriptng.so`.

jmckenna:
@sebastic I can't find the location of the code/file that you mention, it seems to be a test .php file, but I can't find the location of it.  Can you point me to it?

jmckenna:
ah you must mean that the contents of 'mapscript.php' is wrong....checking...

jmckenna:
ok sorry for that, I see what you mean.

jmckenna:
This seems to be related to issue #5708 .   

I can't seem to get SWIG to change the name of the module inside the generated 'mapscript.php' file.

But there is no error when using that file however (and has been leveraged on my systems for 2 years like this), so worst case, the changes in this PR do install the file, and make it visible to packagers.

jmckenna:
@sebastic there is no error on any of the Linux systems I test on, when using the 'mapscript.php' file.

sebastic:
No errors are expected because `return` is executed if loading the extension fails.

Are the classes actually defined when `mapscript.php` is included?

jmckenna:
Correct.  When I say 'works for me' here I meant actually using the mapscript constants and functions defined in that file - I get no errors in my testing on Linux systems.

jmckenna:
Small PHP script to test (if successful no errors returned):

```php
<?php

  // required SWIG include (contains constants for PHP7)
  include("/usr/lib/php/20170718/mapscript.php");

  // open map
  $oMap = new mapObj();

  $oMap = null;
  unset($oMap);
?>
```

PR #6111, opened 2020-07-08
Vagrant and Travis-CI: update to Ubuntu Bionic

### LINKED ISSUE #5775: Switch Travis to use xenial or bionic (trusty EOL in April 2019)
With Ubuntu trusty going EOL in April, the Travis builds should be switched to xenial.

rouault:
Fixes #5775

By using ubuntugis-unstable, this also upgrades to GDAL 3 & PROJ 7

PHP 5.6 and 7.0 Travis-CI configs are no longer tested as not supported on Bionic

Number of expected results have been regenerated due to changes in libraries (with the build in /vagrant/build_vagrant)

Like previously, the PHP 7.1 and 7.2 configs still use PROJ with the legacy proj_api.h, while PHP 7.3 uses PROJ >= 6 proj.h API.
For that later config, a number of .alternative expected results have been created due to subtle differences in coordinate reprojections between PROJ legacy and modern API (with the build in /vagrant/build_vagrant_proj7)


rouault:
CC @dmorissette This may impact your updates for your mapml works. If you haven't started refreshing expected result, I presume merging this PR first would make sense

geographika:
@rouault - wondering how do you get msautotest to use the .alternative results?

rouault:
>  wondering how do you get msautotest to use the .alternative results?

due to https://github.com/mapserver/mapserver/blob/3568fa6f5dbc655afa5e94aea742a2c8d44a441f/msautotest/pymod/testlib.py#L88

geographika:
> due to
> 
> https://github.com/mapserver/mapserver/blob/3568fa6f5dbc655afa5e94aea742a2c8d44a441f/msautotest/pymod/testlib.py#L88

Thanks.  I didn't realise that was already in place. I might look at a similar approach to getting msautotest to run on Appveyor.

rouault:
> at a similar approach to getting msautotest to run on Appveyor.

Hum, I'd be afraid this would result in having a lot of variants, and the issue with AppVeyor/Windows is that it is less easy to update. You actually need to access to the Windows environment to be able to refresh the tests, which would be highly inconvenient from my perspective. At least with Vagrant, I believe anybody is in position to update them.

I was a bit hesitant about introducing this .alternative mechanism to start with. My idea is that at some point we use only the PROJ >= 6 code path, and no longer have a need for alternative expected results

Issue #6114, opened 2020-07-10
GEOMTRANSFORM "labelpnt"with SHADOWCOLOR causes memory corruption


szekerest:
The use case is:

```
CLASS
      NAME "Class (0)"
      LABEL
        FONT "arial"
        SIZE 8
        ALIGN LEFT
        OFFSET 0 0
        POSITION CC
        REPEATDISTANCE -1
        SHADOWCOLOR 255 255 255
        SHADOWSIZE 1 1
        STYLE
          COLOR 0 0 0
          GEOMTRANSFORM "labelpnt"
          SIZE 8
          SYMBOL "Circle"
          WIDTH 2
        END # STYLE
        TEXT "[ASSESSMENT]"
      END # LABEL
      STYLE
        COLOR 206 206 206
        OUTLINECOLOR 147 158 223
        SIZE 8
        WIDTH 0.8
      END # STYLE
      TEMPLATE "query.html"
    END # CLASS
```

PR #6115, opened 2020-07-10
Fix memory corruption when GEOMTRANSFORM and SHADOWCOLOR is used in label simultaneously #6114

szekerest:


jmckenna:
thanks @szekerest

PR #6118, opened 2020-07-14
Add docstrings to all MapScript classes and methods

geographika:
This pull request adds Doxygen-style docstrings to all MapScript classes and methods. This is part of [MS RFC 132: Update MapScript API Docs](https://mapserver.org/development/rfc/ms-rfc-132.html). Changes in this pull request are limited to the SWIG *.i interface files. 
Documentation has been taken from https://mapserver.org/mapscript/mapscript.html where available. 
Other changes include:

- standardising header includes into mapscript.i (some were duplicated)
- tab/space formatting
- typos
- add missing method documentation where possible

I've not had enough time to investigate every undocumented function or TODO from the docs, but hopefully these will build up over time along with expanding the Python MapScript test suite. 

There remain the following tasks which will be added in separate pull requests:

- %feature("shadow")  in Python not yet added (docstrings currently being ignored for these)
- class properties to be documented in the mapserver.h file
- Sphinx doc updates to use the new autodoc approach


sdlime:
This approach does look cleaner - less clutter in the files.

geographika:
@sdlime - is it ok to merge this, and then update the header files with comments for all MapScript accessible properties (using standard C comments) in a separate pull request?

geographika:
If there are no objections I'll merge this in the next couple of days.

sdlime:
No objections from me. Merge away!

PR #6124, opened 2020-07-23
Add a bounding box filter to requests passed down to OGR whenever possible

copierrj:
We observed that our WFS (MapServer + GeoPackage utilizing the OGR backend) is a lot slower when specific spatial operators are used in a filter condition. Although the spatial index in the GeoPackage could be used to accelerate those specific operators it was not.

By adding a bounding box filter to the request passed down to OGR whenever possible (= depending on which operator is used) the spatial index in GeoPackage is utilized for those requests resulting a very significant performance improvement.


jmckenna:
Changes make sense to me.  Thanks @copierrj !

PR #6131, opened 2020-08-18
WFS: limit SQL Select to required fields for faster data retrieval

rouault:
Co-authored-by: Seth G <sethg@geographika.co.uk>

### DIFF
--- a/maplayer.c
+++ b/maplayer.c
@@ -776,6 +776,95 @@ parse_error:
   return MS_FAILURE;
 }
 
+
+static void buildLayerItemList(layerObj *layer)
+{
+  int i, j, k, l;
+  /*
+  ** build layer item list, compute item indexes for explicity item references (e.g. classitem) or item bindings
+  */
+
+  /* layer items */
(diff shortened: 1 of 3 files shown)


rouault:
Supersedes https://github.com/mapserver/mapserver/pull/5997
CC @geographika

geographika:
Thanks @rouault !
Built and tested locally, and all working fine. 
This update greatly improves WFS performance when requesting a subset of layer fields. 

One question/confirmation - any fields referenced in a `CLASS` are selected at the database level even if never used for a WFS request (this was also the case in my previous pull request) - I presume there is no way to get round this due to `buildLayerItemList` also handling fields referenced in `FILTER`, `CLUSTER` etc. ?

rouault:
> any fields referenced in a `CLASS` are selected at the database level even if never used for a WFS request (this was also the case in my previous pull request) - I presume there is no way to get round this due to `buildLayerItemList` also handling fields referenced in `FILTER`, `CLUSTER` etc. ?

yes, msLayerWhichItems() doesn't know the calling context, so there might be things that are irrelevant to WFS. It would require to know which items are needed for which context.

jmckenna:
thanks @rouault & @geographika !!

PR #5088, opened 2015-04-03
Allow OFFSET working on polygon layers

mkofahl:
Allow OFFSET working on polygon layers by changing shape type to line before passing it to geos, which will throw 'BufferBuilder::bufferLineSingleSided only accept linestrings' otherwise (which isn't fully passed to MapServer, actually). Requires #5099.

PR #6132, opened 2020-08-20
Enable offset for polygons

ejn:
Rebased #5088 against current master and added autotest case


ejn:
Note that the `msGEOSError(): GEOS library error.` errors reported by @tbonfort against #5088 seem to have also disappeared in the meantime

ejn:
No idea why the AppVeyor build is failing - looks like some kind of problem with the system.
The Travis build is failing on the PDF output for the test I modified. Is it possible to get the result file from the Travis build in order to compare with the expected result I pushed? I guess it's just a result of systems rather than a problem with the content.

rouault:
Superseded by #6211

Issue #6101, opened 2020-06-22
Add Python MapScript 3.8 support


geographika:
There are currently 2 blockers to adding Python 3.8 support (on Windows): 

1. Python 3.8 uses sqlite version 3.28.0

```
Python 3.8.2 (tags/v3.8.2:7b3ab59, Feb 25 2020, 23:03:10) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import sqlite3
>>> sqlite3.sqlite_version
'3.28.0'
```

The SDK from GISInternals uses sqlite-3.24.0
https://www.gisinternals.com/packageinfo.php?file=release-1911-x64-gdal-mapserver.zip

This causes the following error when attempting to import mapscript:

> ImportError: DLL load failed while importing _mapscript: The specified procedure could not be found.

2. mapserver.dll and its dependencies (proj.dll, geos.dll, gdal300.dll etc.) are currently loaded by adding the /bin folder to the Windows PATH variable. From version 3.8 this is no longer allowed for security reasons. From https://docs.python.org/3/whatsnew/3.8.html#changes-in-the-python-api

> Only the system paths, the directory containing the DLL or PYD file, and directories added with add_dll_directory() are searched for load-time dependencies. **Specifically, PATH and the current working directory are no longer used, and modifications to these will no longer have any effect on normal DLL resolution.** If your application relies on these mechanisms, you should check for add_dll_directory() and if it exists, use it to add your DLLs directory while loading your library.     

To fix this I propose adding a new `MAPSERVER_HOME` variable to the Python MapScript library and adding this folder to the allowed DLLs folders using `os.add_dll_directory` on startup. All MapServer DLLs should be copied to this location prior to running tests. 

On the command line the variable can be set:

```SET MAPSERVER_HOME=D:\Temp\DLLs```

And the mapscript __init__.py file will have the following added:

```python
import sys
import platform

if platform.system() == 'Windows' and (3, 8) <= sys.version_info:
    mapserver_home = os.getenv('MAPSERVER_HOME', '')
    if os.path.isdir(mapserver_home):
        with os.add_dll_directory(mapserver_home):
            import mapscript._mapscript
```

See https://github.com/Toblerity/Fiona/issues/851 which has the same problem.

jmckenna:
+1 on implementing on leveraging a MAPSERVER_HOME environment variable, this will be very useful for Windows environments.

dmorissette:
[...]

PR #6136, opened 2020-09-13
Test py3 MapScript Builds on Travis

geographika:
This pull request adds Python3 building and testing to the Travis Linux CI, using an updated Python MapScript build process. 
Python MapScript is tested on Python 2.7 and Python 3.7. msautotest is tested on Python 2.7, 3.6, and 3.7.

Appveyor and the CMake build process now use the `Python_ROOT_DIR` variable rather than `PYTHON_EXECUTABLE` to force a Python version. `PYTHON_EXECUTABLE` did not work correctly with Travis in picking up Python directories in initial attempts. In addition `Python_ROOT_DIR` is one of the standard [HINTs](https://cmake.org/cmake/help/v3.12/module/FindPython.html) for the FindPython module. 

From the issue at https://gitlab.kitware.com/cmake/cmake/-/issues/19492:

> Overriding Python_EXECUTABLE is not possible to avoid any inconsistency between various components which can be searched for.

With the update of the [Travis image to Bionic](https://github.com/mapserver/mapserver/pull/6111) we now have CMake 3.12. 
The [FindPython](https://cmake.org/cmake/help/v3.12/module/FindPython.html) CMake module seems to work fine, so there is no longer a need for the previous hacks to work with older versions of CMake. Users wishing to build Python modules can use their default Python or set the `Python_ROOT_DIR` variable to point to a non-default Python installation. 

Unfortunately the `Python_ROOT_DIR` approach did not work correctly on Travis (FindPython never found the hinted version). I eventually found out it is due to Travis using [pyevn](https://github.com/pyenv/pyenv) to handle the various Python environments. There is a useful introduction to pyenv at [Real Python](https://realpython.com/intro-to-pyenv/). 

All msautotest test code now runs under both py2 and py3. One test case was modified to work under py3 (a byte/string issue), and the inbuilt Python server differs between py2 and py3. The remaining code had already been made Python3 compatible thanks to the work by @rouault on [updating to pytest](https://github.com/mapserver/mapserver/pull/5984). 

In summary each of the Travis CI builds now runs the following:

- `BUILD_NAME=PHP_7.1` - builds and tests Python MapScript for Python 2.7, and msautotest using Python 2.7
- `BUILD_NAME=PHP_7.2_WITH_ASAN` - no Python MapScript is tested in this build, msautotest run using Python 3.6
- `BUILD_NAME=PHP_7.3_WITH_PROJ7` - tests Python MapScript for Python 3.7, and msautotest using Python 3.7

Please add any issues, comments, suggestions to this pull request. 

@tomkralidis - MapScript for Python 3.7 seems fine on Linux and Windows. It may be worth getting this latest CMake approach, and setting `Python_ROOT_DIR` to point to your Conda install to see if it makes a difference.


geographika:
Does anyone have objections to merging this, or any reservations?
I'll take a look at adding py3,8 support for MapServer 8.0 separately (see #6101). I see there are now PHP 7.4 and 8.0 releases we could add builds for.  

@rouault - would it make sense to add the Python dependencies `cpp-coveralls pyflakes lxml` to `msautotest/requirements.txt`? 
`lxml` is used for msautotest but not the others, so we could keep them separate.

Issue #5418, opened 2017-04-06
fix WFS GetFeature featureid queries


tomkralidis:
Testing against master:

## Mapfile
```
MAP
 NAME test
 IMAGETYPE PNG
 STATUS ON
 EXTENT -180 -90 180 90
 MAXSIZE 4096
 SIZE 500 300
 SYMBOL
  NAME 'circle'
  TYPE ELLIPSE
  FILLED TRUE
  POINTS
   1 1
  END
 END
 CONFIG "MS_ERRORFILE" "/tmp/woudc-ows-bug.log"
 DEBUG 5
 PROJECTION
  "init=epsg:4326"
 END
 TRANSPARENT ON
 QUERYMAP
  STATUS OFF
  SIZE 400 300
  STYLE HILITE
  COLOR 255 255 0
 END
 WEB
  METADATA
    "ows_enable_request" "*"
    "ows_title" "foo"
    "ows_onlineresource" "http://localhost/ows"
    "wfs_srs" "EPSG:4326"
  END
 END
LAYER
 NAME "stations"
 STATUS ON
 DEBUG 5
 CONNECTION "host=localhost dbname=foo user=foo password=foo"
 CONNECTIONTYPE POSTGIS
 PROCESSING "CLOSE_CONNECTION=DEFER"
 DATA "the_geom from stations using unique platform_id using srid=4326"
 TYPE POINT
 DUMP TRUE
 TEMPLATE "ttt.html"
 PROJECTION
  "init=epsg:4326"
 END
 METADATA
  "ows_title" "foo"
  "ows_abstract" "foo"
  "gml_featureid" "platform_id"
  "gml_include_items" "all"
  "gml_types" "auto"
  "wfs_srs" "EPSG:4326"
 END
CLASS
 NAME "stations"
 GROUP "default"
 STYLE
  SYMBOL "circle"
  SIZE 8
  COLOR 255 0 0
  OUTLINECOLOR 0 0 0
 END
END
END
END
```

## Data (PostgreSQL/PostGIS GeoJSON dump via ogr2ogr)
```json
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {
        "date_from": "1952\/06\/01 00:00:00+00",
        "date_to": "2017\/02\/01 00:00:00+00",
        "gaw_id": "SIS",
        "platform_id": "043",
        "platform_name": "Lerwick",
        "acronym": "UKMO",
        "platform_type": "STN",
[...]

Issue #5408, opened 2017-03-13
Cannot override GML attribute types when using "auto"


geographika:
In a layer's METADATA I have the following GML settings:

```
    "gml_types"	"auto"
    "gml_MyDate_type" "date"	
```

I would expect that all field types are automatically set, but that the MyDate type is overriden as a GML "date" type (useful if creating layers based on views that have string dates). 

However the "gml_MyDate_type" is ignored. I have run into this issue using the MSSSQL driver, but it appears it would be the same for all other drivers implementing `"gml_types" auto`

Looking at the code it seems to check for a "gml_type" value, which as I understand it will never exist. 

```
      snprintf( md_item_name, sizeof(md_item_name), "gml_%s_type", buffer );
      if( msOWSLookupMetadata(&(layer->metadata), "G", "type") == NULL )
        msInsertHashTable(&(layer->metadata), md_item_name, gml_type );
```

I believe the fix would be to use the following, but I am not certain what the current desired behaviour is. 

```
      snprintf( md_item_name, sizeof(md_item_name), "gml_%s_type", buffer );
      if( msOWSLookupMetadata(&(layer->metadata), "G", md_item_name) == NULL )
      // or maybe
      // snprintf( md_item_name2, sizeof(md_item_name), "%s_type", buffer );
      // if( msOWSLookupMetadata(&(layer->metadata), "G", md_item_name2) == NULL )
        msInsertHashTable(&(layer->metadata), md_item_name, gml_type );
```
See [msOWSLookupMetadata](https://github.com/mapserver/mapserver/blob/master/mapows.c#L809)

The MSSQL, PostGIS, Oracle, and OGR drivers seem to all use the same approach:

- https://github.com/mapserver/mapserver/blob/21c41fe99d04ec64d1cfbb681d7ef01d20ade5f9/mapmssql2008.c#L890
- https://github.com/mapserver/mapserver/blob/21c41fe99d04ec64d1cfbb681d7ef01d20ade5f9/mappostgis.cpp#L2896
- https://github.com/mapserver/mapserver/blob/21c41fe99d04ec64d1cfbb681d7ef01d20ade5f9/maporaclespatial.c#L2955
- https://github.com/mapserver/mapserver/blob/21c41fe99d04ec64d1cfbb681d7ef01d20ade5f9/mapogr.cpp#L2684

geographika:
Fixed with #6137

PR #6137, opened 2020-09-13
Allow custom override for gml_types "auto"

geographika:
This pull request allows GML types to be set automatically for the MSSQL driver, but still allow individual fields to be overwritten manually if required. See #5408 and #5418
I believe this was the original intention of the code, but from debugging and testing I don't believe this has ever been the case. 

The code currently checks the "gml" namespace for "type", "width", and "precision" using msOWSLookupMetadata. However, these properties only ever exist with a property name e.g. 

```
gml_[item name]_type
gml_[item name]_precision
gml_[item name]_width
```

Therefore, the searches always return 0, and the auto values are always used. 
It looks like the code has been copied from driver to driver. 

As in the test case included in the pull request:

```
    METADATA
        "gml_include_items" "all"
        "gml_types" "auto"
        "gml_ogr_fid_type" "long" # override the default gml type of integer
    END
```

Returns:

```xml
    <element name="ogr_fid" minOccurs="0" type="long"/>
    <element name="name" minOccurs="0" nillable="true" type="string"/>
    <element name="name_ar" minOccurs="0" nillable="true" type="string"/>
```

Please let me know if I am missing something. 

I've also added the update for the following drivers:

- Oracle
- Shapefile
- PostGIS
- OGR

Looks like the original code is from 2010 as part of https://github.com/mapserver/mapserver/issues/3570 implementing https://mapserver.org/development/rfc/ms-rfc-62.html#gml-types-auto

### DIFF
--- a/mapows.h
+++ b/mapows.h
@@ -150,8 +150,8 @@ MS_DLL_EXPORT const char * msOWSLookupMetadata2(hashTableObj *pri,
     const char *namespaces,
     const char *name);
 
-MS_DLL_EXPORT int updateGMLFieldMetadata(layerObj *layer, const char *field_name, const char *gml_type, const char *gml_width, 
-                        const char *gml_precision, const short nullable);
+MS_DLL_EXPORT int msUpdateGMLFieldMetadata(layerObj *layer, const char *field_name,
+    const char *gml_type, const char *gml_width, const char *gml_precision, const short nullable);
 
 void msOWSInitRequestObj(owsRequestObj *ows_request);
 void msOWSClearRequestObj(owsRequestObj *ows_request);
--- a/mapows.h
+++ b/mapows.h
@@ -149,7 +149,10 @@ MS_DLL_EXPORT const char * msOWSLookupMetadata2(hashTableObj *pri,
     hashTableObj *sec,
     const char *namespaces,
     const char *name);
-    
+
+MS_DLL_EXPORT int updateGMLFieldMetadata(layerObj *layer, const char *field_name, const char *gml_type, const char *gml_width, 
+                        const char *gml_precision, const short nullable);
+
 void msOWSInitRequestObj(owsRequestObj *ows_request);
 void msOWSClearRequestObj(owsRequestObj *ows_request);
 
--- a/mappostgis.cpp
+++ b/mappostgis.cpp
@@ -2891,20 +2891,8 @@ msPostGISPassThroughFieldDefinitions( layerObj *layer,
       gml_type = "DateTime";
     }
 
-    char md_item_name[256];
-    snprintf( md_item_name, sizeof(md_item_name), "gml_%s_type", item );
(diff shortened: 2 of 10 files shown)


rouault:
Would be good to move this copied&pasted logic in a function :-)

geographika:
@rouault - ok I gave it a try!

Unsure about the following:

- should I return a TRUE value from the function (not used) or simply make it void?
- md_item_name[256] was smalled [64] mapshape but this should have no impact
- unsure if function is located in best place
- I think i updated the associated .h file correctly

Nullable is only currently used in MSSQL driver, but it having it in the function should make it easier for other drivers to implement if required.

mappostgis.cpp
```diff
@@ -2891,20 +2891,8 @@ msPostGISPassThroughFieldDefinitions( layerObj *layer,
       gml_type = "DateTime";
     }
 
-    char md_item_name[256];
-    snprintf( md_item_name, sizeof(md_item_name), "gml_%s_type", item );
-    if( msOWSLookupMetadata(&(layer->metadata), "G", "type") == NULL )
-      msInsertHashTable(&(layer->metadata), md_item_name, gml_type );
-
-    snprintf( md_item_name, sizeof(md_item_name), "gml_%s_width", item );
-    if( !gml_width.empty()
```
geographika, line 2900: Might need to check this outside of the updateGMLFieldMetadata function?

mapows.h
```diff
@@ -149,7 +149,10 @@ MS_DLL_EXPORT const char * msOWSLookupMetadata2(hashTableObj *pri,
     hashTableObj *sec,
     const char *namespaces,
     const char *name);
-    
+
+MS_DLL_EXPORT int updateGMLFieldMetadata(layerObj *layer, const char *field_name, const char *gml_type, const char *gml_width,
```
rouault, line 153: no need to export this
```suggestion
int updateGMLFieldMetadata(layerObj *layer, const char *field_name, const char *gml_type, const char *gml_width, 
```

rouault (commented):
Thanks ! LGTM with the small suggestion

mappostgis.cpp
geographika: According to https://stackoverflow.com/questions/8103342/calling-c-str-of-empty-string the c_str() will reurn an empty C string so no need to do a further check

mapows.h
geographika: @rouault - I tried this, but as the MSSQL is a separate plugin/DLL I think this is required https://ci.appveyor.com/project/MapServer/mapserver/builds/35196629/job/t24q8bmyf2kc0kob#L428

`mapmssql2008.obj : error LNK2019: unresolved external symbol _updateGMLFieldMetadata referenced in function _columnName `

rouault: ah, good point. Could you then rename the function as ```msUpdateGMLFieldMetadata``` ?

geographika: Done

geographika:
@rouault - thanks for the review and feedback.

Issue #5192, opened 2015-11-01
NULL shapes not supported in WFS


jratike80:
While studying https://github.com/mapserver/mapserver/issues/3487 which was about a rendering bug when shapefile contains NULL shapes I noticed that NULL shapes lead to wrong result in WFS. They are not included in the data that is requested with GetFeature.

I edited a states shapefile by updating one geometry into NULL with ogrinfo. Feature count is still 49 as reported by ogrinfo:

ogrinfo states.shp -al -so
Layer name: states
Metadata:
  DBF_DATE_LAST_UPDATE=2015-11-01
Geometry: Polygon
Feature Count: 49

WFS DescribeFeatureType allows missing geometries for the feature type as can be seen from the schema:
<complexType name="states_shpType"><complexContent><extension base="gml:AbstractFeatureType"><sequence><element name="msGeometry" type="gml:GeometryPropertyType" minOccurs="0" maxOccurs="1"/>

However, the feature with empty geometry (Illinois in my test) is missing totally from the WFS GetFeature resultset that comes with
http://localhost:8060/cgi-bin/mapserv.exe?map=c:/ms4w/apps/null_test.map&service=WFS&version=1.1.0&request=getfeature&typename=states_shp

This happens both with native shapefile connection and through OGR connection.

Odd thing is that if I use OGR connection and input data as OpenJUMP JML instead of shapefile then Illinois is still missing from the plain GetFeature. However, in this case I can read data of Illinois by making a direct featureid query:

http://localhost:8060/cgi-bin/mapserv.exe?map=c:/ms4w/apps/null_test_jml.map&service=WFS&version=1.1.0&request=getfeature&typename=states_shp&featureid=states_shp.17

[...]

PR #6138, opened 2020-09-16
Allow null features in OGR output

geographika:
This pull request removes the checks for `numlines < 1` allowing null features to be output to an OGR format, for example GeoJSON which would return these features as `mygeom: null`. Pull request related to #5192 (although the issue from @jratike80 is a slightly different case). A full set of tests has been added for the MSSQL Driver (attempts to add WKT tests via CONNECTION OGR were unsuccessful as empty geometries were converted to normal geometries). 

To retrieve these features requires `"wfs_use_default_extent_for_getfeature" "false"` to be set on the layer or the null features would be filtered out with an extent query. 

Several drivers have checks for NULLs, for example the MSSQL driver at https://github.com/mapserver/mapserver/blob/ec347870be17fdaa0bb4a2d68e80c677a3c55d31/mapmssql2008.c#L2400L

Supplying an empty geometry rather than NULL gets around these checks:

`geometry::STGeomFromText('POINT EMPTY', 3857) mygeom`

The use case is to return all features as part of a WFS in GeoJSON even if they have no geometry. The feature attributes can still be very useful (and null features are still included in counts which can be confusing if they are never returned). 
A current workaround involve setting fake geometries, but then functions such as zooming to features are incorrect, as are summary lengths and areas etc.
 
As this function is used for all OGR outputs I'm not certain of possible side effects when outputting to all other formats. 
Currently, MapServer would log an error and the feature would not be written to the output. 

With this pull request `hGeom` remains NULL and is passed to `OGR_F_SetGeometryDirectly( hFeat, hGeom )`

There is a check further on to see if writing the feature failed, so I believe any errors with empty geometries for a particular driver would be handled here:

https://github.com/mapserver/mapserver/blob/9bb624f155b8fafbe680de218571f582b1e28347/mapogroutput.cpp#L523

When testing with CSV the output using this pull request is:

```
WKT,fid
,"1"
```


geographika:
Issues to address are as follows:

1. null polygons are being output as follows, rather than `geometry: null` (same for NULL multipolygons):

https://github.com/mapserver/mapserver/blob/06c0b8782369b43df9f3f3167a3cfe809d74b5ac/msautotest/mssql/expected/null_polygon.json#L6

From https://tools.ietf.org/html/rfc7946:

> A GeoJSON Geometry object of any type other than
> "GeometryCollection" has a member with the name "coordinates".
> The value of the "coordinates" member is an array.  The structure
> of the elements in this array is determined by the type of
> geometry.  GeoJSON processors MAY interpret Geometry objects with
> empty "coordinates" arrays as null objects.
> 
> A Feature object has a member with the name "geometry".  The value
> of the geometry member SHALL be either a Geometry object as
> defined above or, in the case that the Feature is unlocated, a
> JSON null value.

There is a [discussion at mapnik](
https://github.com/mapnik/mapnik/issues/2083#issuecomment-28972062) about empty coords or nulls. Nulls was the preferred option. Use of null would make MapServer consistent across all geometry types. 


2. multipolygons are always flattened to polygons:

https://github.com/mapserver/mapserver/blob/06c0b8782369b43df9f3f3167a3cfe809d74b5ac/msautotest/mssql/expected/multipolygon.json#L6

This is due to: https://github.com/mapserver/mapserver/blob/1ca57ae8f9bca01e80dbe61185d919996a8886b0/mapogroutput.cpp#L442

`wkbMultiPolygon !== wkbPolygon` so always flattened. 

Note I used MSSQL for the tests as empty geometries from WKT using CONNECTION OGR always got converted to geometries, e.g. an empty point would become POINT (0, 0), I believe due to the following line:

https://github.com/mapserver/mapserver/blob/1ca57ae8f9bca01e80dbe61185d919996a8886b0/mapogr.cpp#L227

PR #6141, opened 2020-09-21
add check-crlf action

jmckenna:
- enables check-crlf Github action to verify that all files have a LF linefeed


jmckenna:
For this to work with our source we need to exclude directories or filetypes (there is a [PR in check-crlf](https://github.com/erclu/check-crlf/pull/1) for this, will monitor there).  For now closing this PR.

PR #6144, opened 2020-09-23
change to LF

jmckenna:
- change more files from CRLF to LF
- in preparation to implement the check-crlf Github action in the future
  - once check-crlf is enhanced to exclude directories or filetypes (such as our msautotest .pdf or .tab files)


jmckenna:
oops more issues, sorry for noise...

PR #6150, opened 2020-09-25
Fixes for using none ascii characters in Sql Server connection strings, queries and results

wicksell:
There has been an encoding problem in the SQL Server plugin. Queries used SQLExecDirect instead of SQLExecDirectW making the plugin dependent on that the systems locale matched the encoding of the map-file. For Swedish characters on Swedish Windows the map file needed to be in ISO-8859-1 but other connections such as PostGIS reguire the map-file to be in UTF-8. So combining different type of connections was a problem. The plugin was also using a bad pattern when reading attributes from SQL Server, see https://docs.microsoft.com/en-us/sql/relational-databases/native-client/features/odbc-driver-behavior-change-when-handling-character-conversions?view=sql-server-ver15

### DIFF
--- a/mapstring.cpp
+++ b/mapstring.cpp
@@ -1900,7 +1900,7 @@ wchar_t* msConvertWideStringFromUTF8 (const char* string, const char* encoding)
 {
 #ifdef USE_ICONV
   wchar_t* output = NULL;
-  char* errormessage = NULL;
+  const char* errormessage = NULL;
   iconv_t cd = NULL;
   size_t nStr;
   size_t nInSize;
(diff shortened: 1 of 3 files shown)


mapstring.cpp
```diff
@@ -1896,6 +1896,86 @@ char* msConvertWideStringToUTF8 (const wchar_t* string, const char* encoding)
...
+    }
+
+    cd = iconv_open(encoding, "UTF-8");
+
+    nOutSize = nBufferSize;
+    if ((iconv_t)-1 != cd) {
+      nInSize = sizeof (char)*nStr;
+      pszUTF8 = string;
+      pwszWide = output;
+      iconv_status = iconv(cd, &pszUTF8, &nInSize, (char **)&pwszWide, &nOutSize);
```
szekerest, line 1932: For the linux builds the warnings are treated as error, so we need to handle the cast from const char* to char* in the second argument

szekerest (changes requested):
As soon as all builds are passed, the changes seem to be correct and can be merged

jmckenna:
thanks @wicksell !!

PR #6151, opened 2020-09-25
Make it possible to send scale_independent to drawLegend via MapScript

wicksell:
This make it possible to create a legend for all layers even if they are not within scale range.


geographika (approved):
Looks good to me. The new MapScript docs process should pick up the new parameter automatically.

geographika:
Thanks @wicksell !

PR #6163, opened 2020-10-04
mapmssql2008.c: fix warnings when building on Linux

rouault:
wchar_t is a 4-byte int whereas SQLWCHAR is a 2-byte int...

@geographika / @szekerest  any of you want to test this on Windows  (with WITH_ICONV=ON) ?


geographika:
@rouault - built and tested the branch locally on Windows and all appears fine. 

For info the MSSQL driver is also tested on Appveyor now using the suite of tests in https://github.com/MapServer/MapServer/tree/master/msautotest/mssql

Issue #6166, opened 2020-10-07
Map metadata used when Layer Metadata is available


geographika:
As per the mailing list post at https://lists.osgeo.org/pipermail/mapserver-users/2020-October/081905.html: with regards to the `GetMetadata` requests:

> Looking at the code it appears the "contact" metadata can only be retrieved
> at the MAP rather than LAYER level [2].
> Is this something that should be available at LAYER level also, perhaps using
> "wms_attribution_title"?

Reply from @tomkralidis:

> Good point.  Perhaps if MAP.LAYER.METADATA.wms_attribution_title is available
we can override accordingly

tomkralidis:
@geographika thanks for filing this issue.  When looking at the [WMS Server documentation](https://mapserver.org/ogc/wms_server.html#layer-object-metadata), I see the following metadata that can be applied from `LAYER.METADATA` with suggested elements they would override:

- `wms_attribution_title`: would override `MAP.WEB.METADATA.*_contactorganization`
- `wms_attribution_onlineresource`: would override `MAP.WEB.METADATA.*_onlineresource`

In addition to these overrides, would we want to turn OFF ALL the `MAP.WEB.METADATA.*_contact_*` metadata if either or both of the above are available?  Or simply override the above two and keep the rest?

geographika:
Tricky one. Initially I would have thought all MAP metadata would apply unless overridden at the LAYER level. 

But could there be cases where a LAYER contact could have the wrong details from the MAP level? I guess this is something users would have to watch out for, and would need to override all relevant settings at the LAYER level. 

Although is this possible? E.g. `wms_contactelectronicmailaddress` is only available at the MAP.WEB level (at least when listed in the docs). Maybe it can be set at the layer level even if not used by `GetCapabilitites`.

tomkralidis:
Correct, contact metadata is only available at the MAP level (and not the LAYER level).  So then are we proposing introducing the `wms_contact*` metadata at the LAYER level?  Would this get much use outside of the Layer Metadata API?  Also, this probably needs an RFC.  Thoughts?

jratike80:
But is there place for layer lever contact information in the GetCapabilities schema http://schemas.opengis.net/wms/1.3.0/capabilities_1_3_0.xsd?

tomkralidis:
[...]

PR #6169, opened 2020-10-12
Metadata: use attribution_title if available (#6166)

tomkralidis:
Implements #6166 

In addition, removes erroneous default date output.

### DIFF
--- a/mapmetadata.c
+++ b/mapmetadata.c
@@ -528,6 +528,11 @@ xmlNodePtr _msMetadataGetIdentificationInfo(xmlNsPtr namespace, mapObj *map, lay
 
   psDNode = xmlNewChild(psCINode, namespace, BAD_CAST "date", NULL);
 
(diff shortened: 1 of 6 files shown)


rouault (commented):
I'm not super qualified to comment on this.
Just 2 remarks;
- the mention "In addition, removes erroneous default date output." should be added to the commit message itself. (ideally this would have been a separate commit)
- I guess we would need extending an existing test to set "attribution_title" and see its effect in the XML output

tomkralidis:
Thanks for the feedback.  Split into two commits with updated test.

jmckenna:
thanks @tomkralidis Again, although it is only case, maybe rename your local repository from 'mapserver' to 'MapServer', just in case it makes the other Github tools like backport-bot choke down the road (although this PR will not be backported).   thanks again!

tomkralidis:
Thanks again @jmckenna, done: https://github.com/tomkralidis/MapServer

PR #6180, opened 2020-11-19
OGC API Proof-of-Concept

sdlime:
Initial work for OGC API support via the MapServer API following the implementation notes described in https://github.com/mapserver/mapserver/wiki/OGC-API-implementation.

--Steve


sdlime:
For draft RFC discussion see: https://github.com/mapserver/mapserver/wiki/OGC-API-RFC-Draft

cgiutil.h
```diff
@@ -70,6 +70,12 @@ typedef struct {
   char *postrequest;
 
   char *httpcookiedata;
+
+  // todo, SWIG?
+  const char *path_info;
```
geographika, line 75: Looks like this is breaking the Appveyor/Windows build with `C:\projects\mapserver\mapscript\v8\..\swiginc\..\..\cgiutil.h(75): warning 451: Setting const char * member may leak [C:\projects\mapserver\build\mapscript\python\_pythonmapscript.vcxproj]`

sdlime: It's just a pointer to an environment variable so const is appropriate (I think), but I'll change it and see what happens.

sdlime: No effect... Warning is gone but it was just a warning. Dying with mapscript w/both Travis and Appveyor. I didn't think any of this would have impacted mapscript.

third-party/include_nlohmann_json.hpp
```diff
@@ -0,0 +1,38 @@
...
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ ****************************************************************************/
+
+#if defined(__GNUC__)
+#pragma GCC system_header
+#endif
+
```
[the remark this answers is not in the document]
sdlime: The latter looks more straight forward isn't it?

cgiutil.h
sdlime: Resolved the warning but mapscript isn't building. Odd...

cgiutil.h
```diff
@@ -70,6 +70,12 @@ typedef struct {
   char *postrequest;
 
   char *httpcookiedata;
+
+  // todo, SWIG?
```
geographika, line 74: Wrap these in a ifndef statement for now?

#ifndef SWIG
  char *path_info;
  int api_path_length;
  char **api_path;
#endif

cgiutil.h
geographika: Any char **params need a typemap to handle them I think. It looks like the C# bindings are failing on this:

OWSRequest.cs(176,14): error CS0029: Cannot implicitly convert type 'OSGeo.MapServer.SWIGTYPE_p_p_char' to 'string[]' [C:\projects\mapserver\build\mapscript\csharp\mapscript.vcxproj]

Maybe easiest to hide from SWIG for now?

cgiutil.h
sdlime: I think you're right. Param names and values are also hidden...

third-party/include_nlohmann_json.hpp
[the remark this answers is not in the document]
rouault: > ou think it's worth allowing a system version?

that could be done later. Adding the ``#define nlohmann ms_nlohmann`` trick however will avoid any potential clash with PROJ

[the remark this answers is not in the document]
sdlime: Do mapogcapi.cpp and inja.hpp then need to use the ms_nlohmann namespace explicitly? So:

`using json = ms_nlohmann::json;`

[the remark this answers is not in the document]
rouault: if you add #define nlohmann ms_nlohmann in include_nlhmann_json.hpp, then you shouldn't have anyparticular to do in other files. But you need to make sure that inja.hpp is included after it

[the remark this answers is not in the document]
sdlime: Perfect, I made the change... thanks!

jmckenna:
An early view of the (features/WFS equivalent) collection, as most readers will not have seen this yet...
ogcapi-view1

ogcapi-view2

mapogcapi.cpp
```diff
@@ -0,0 +1,1117 @@
...
+          { "href", api_root + "/collections/" + std::string(id_encoded) + "?f=json" }
+        },{
+          { "rel", format==OGCAPI_FORMAT_HTML?"self":"alternate" },
+          { "type", OGCAPI_MIMETYPE_HTML },
+          { "title", "This collection as HTML" },
+          { "href", api_root + "/collections/" + std::string(id_encoded) + "?f=html" }
+        },{
+          { "rel", "items" },
+          { "type", OGCAPI_MIMETYPE_JSON },
+          { "title", "Items for his collection as JSON" },
```
jmckenna, line 531: minor spelling mistake, "Items for this collection as JSON"

jmckenna:
@sdlime I can't seem to get `collections/ocean/items?f=json` return anything but an error, even though I set `gml_featureid`, testing with various formats I get the same error (SpatiaLite, GeoJSON, shp) :
```
{"code":"ServerError","description":"Error getting feature. Missing required featureid metadata."}
```
I'm wondering if this is my local issue (or if this should be working at all, yet).

mapogcapi.cpp
sdlime: Fixed!

sdlime:
> I'm wondering if this is my local issue (or if this should be working at all, yet).

It should be working but I neglected to check the gml namespace - just oga, ows and wfs. If you use any of those it should work. I also just pushed a fix.

jmckenna:
Confirmed the fixes, thanks!  A similar problem still exists for `collections/ocean/items?f=html` :
```
{"code":"ConfigError","description":"Template rendering error. [inja.exception.render_error] (at 30:110) variable 'feature.id' not found (collection-items.html)."}
```

sdlime:
> Confirmed the fixes, thanks! A similar problem still exists for `collections/ocean/items?f=html` :
> 
> ```
> {"code":"ConfigError","description":"Template rendering error. [inja.exception.render_error] (at 30:110) variable 'feature.id' not found (collection-items.html)."}
> ```

I believe that means the item you reference in gml_featureid doesn't exist in the data. Should be an explicit error but that's how I was able to trigger the exact error message.

jmckenna:
> I believe that means the item you reference in gml_featureid doesn't exist in the data. Should be an explicit error but that's how I was able to trigger the exact error message. Will add that as a todo...

Your last change now at least gives me the same error for JSON and HTML ("Error getting feature. Feature id not found."}  This same metadata works for WMS/GetFeatureInfo, for a long time, but I will do more testing here now with WFS and then with OGCAPI, with this metadata.  Thanks!

sdlime:
> > I believe that means the item you reference in gml_featureid doesn't exist in the data. Should be an explicit error but that's how I was able to trigger the exact error message. Will add that as a todo...
> 
> Your last change now at least gives me the same error for JSON and HTML ("Error getting feature. Feature id not found."} This same metadata works for WMS/GetFeatureInfo, for a long time, but I will do more testing here now with WFS and then with OGCAPI, with this metadata. Thanks!

There certainly could be differences between the API and WFS. The check I added simply looks to see if the value referenced by gml_featureid is in the list of items associated with the layer (case-insensitive comparison). If not, then the error you see is generated. The query associated with /collections/{collection}/items is a bbox filter so the error isn't trapped until the first feature is processed (as things sit now).

jmckenna:
> 
> There certainly could be differences between the API and WFS. The check I added simply looks to see if the value referenced by gml_featureid is in the list of items associated with the layer (case-insensitive comparison). If not, then the error you see is generated. The query associated with /collections/{collection}/items is a bbox filter so the error isn't trapped until the first feature is processed (as things sit now).

You're right, it seems to be a data issue, but your changes helped.  Will try with other data before I go crazy here ha.

jmckenna:
Thanks Steve, maybe it was a 'comedy of errors' on my side.  It works now! (this is SpatiaLite)   (also, I enjoy the 'map goes here' in the code, ha)
ogcapi-view3

ogcapi-view4

sdlime:
> Thanks Steve, maybe it was a 'comedy of errors' on my side. It works now! (this is SpatiaLite) (also, I enjoy the 'map goes here' in the code, ha)

Cool... There are still bits and pieces to add (or steal from @tomkralidis and pygeoapi).

jmckenna:
> Cool... There are still bits and pieces to add (or steal from @tomkralidis and pygeoapi).

Absolutely agreed (the pygeoapi 'master' demo is my go-to), I especially like exposing the provider metadata on the landing page.

rouault:
@sdlime  In this pull request / your branch, I've just merged main to fix a conflict. And I've issued https://github.com/sdlime/mapserver/pull/2 against your branch with a number of fixes & enhancements + testing. With them, the GDAL OAPIF driver works against MapServer. With those fixes, I think we should consider merging/squashing this branch into main. The main missing point for conformance is probably the /api endpoint

sdlime:
> @sdlime In this pull request / your branch, I've just merged main to fix a conflict. And I've issued [sdlime#2](https://github.com/sdlime/mapserver/pull/2) against your branch with a number of fixes & enhancements + testing. With them, the GDAL OAPIF driver works against MapServer. With those fixes, I think we should consider merging/squashing this branch into main. The main missing point for conformance is probably the /api endpoint

Thanks so much for the pull request - huge help getting over the hump. I'll review ASAP. Yes, the /api or /schema or whatever we want to call it endpoint is on my "todo" list. Along with:

1. updating the inja .hpp file, there was a bug fix that simplifies some of the templating
2. adding tests that stress the implementation - bad endpoints, signatures, template errors, etc...

That work can be added post-merge though.

--Steve

rouault:
/api endpoint implementation available in https://github.com/sdlime/mapserver/pull/3

rouault:
With the latest changes in https://github.com/sdlime/mapserver/pull/3, CITE testing with the msautotest/api/ogcapi.map passes ! Using the Docker image mentioned at https://cite.opengeospatial.org/teamengine/about/ogcapi-features-1.0/1.0/site/

mapserver.h
```diff
@@ -361,7 +361,7 @@ extern "C" {
 
 
   /* #define MS_VALID_EXTENT(minx, miny, maxx, maxy)  (((minx<maxx) && (miny<maxy))?MS_TRUE:MS_FALSE) */
-#define MS_VALID_EXTENT(rect)  (((rect.minx < rect.maxx && rect.miny < rect.maxy))?MS_TRUE:MS_FALSE)
+#define MS_VALID_EXTENT(rect)  ((((rect).minx < (rect).maxx && (rect).miny < (rect).maxy))?MS_TRUE:MS_FALSE)
```
jratike80, line 364: What things does this validate? Could there be some use case for using a point when in the bbox min=max?

rouault: > Could there be some use case for using a point when in the bbox min=max?

that's a good point (if I might say...). I raised that in https://github.com/opengeospatial/ogcapi-features/issues/288#issuecomment-576326563 and the answer was "A degenerated bbox is a point and should be evaluated as a point."

sdlime: Agreed. We could add a MS_VALID_SEARCH_EXTENT macro that uses <=, sound ok?

geographika (commented):
A couple of issues when testing under IIS/Windows

mapservutil.c
```diff
@@ -165,34 +165,44 @@ mapObj *msCGILoadMap(mapservObj *mapserv)
...
-  if(i == mapserv->request->NumParams) {
+  if(map_value == NULL) {
     if(ms_mapfile == NULL) {
       msSetError(MS_WEBERR, "CGI variable \"map\" is not set.", "msCGILoadMap()"); /* no default, outta here */
       return NULL;
     }
     ms_mapfile_tainted = MS_FALSE;
   } else {
-    if(getenv(mapserv->request->ParamValues[i])) { /* an environment variable references the actual file to use */
+    if(getenv(map_value)) { /* an environment variable references the actual file to use */
```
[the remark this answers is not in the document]
sdlime: I'll look at these. There should be no getenv() in that function. I must have mucked up the merge with main a bit... *sigh*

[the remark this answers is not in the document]
sdlime: You're correct on needing the config file for IIS. We really need something other than a full-path mapfile for this to work.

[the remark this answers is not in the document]
sdlime: Standardizing configuration across platforms and web servers is the primary benefit of a config file. IIS couldn't use environment variable references to mapfiles for regular processing either.

geographika:
@sdlime - with a couple of hacks I got results back. Many thanks for implementing this - it opens up the whole of MapServer to a new nice clean web-ready API 🥇 

I've a few notes/comments that may be of use to anyone testing the new API. I'll add them here, but happy to put them into a new issue if required.

1. Layer extents.

http://localhost/mapserver/features/mymap/ogcapi/collections?f=html was returning:

`{"code":"ConfigError","description":"Error getting collection.Unable to get collection bounding box."}`

- as noted in the code: `throw std::runtime_error("Unable to get collection bounding box."); // might be too harsh since extent is optional` if any layer is missing a `"ows_extent"` then this error is thrown. Would defaulting to the MAP extent be an option?

2. I had some trouble with setting the template directory (I assumed it was a permissions issue). All requests were returning `{"code":"ServerError","description":"General template handling error."}

In mapogcapi.cpp I added an extra try block:

```c
    outputError(OGCAPI_CONFIG_ERROR, "Template rendering error. " + std::string(e.what()) + " (" + std::string(filename) + ").");
    return;
  }
  catch (const inja::InjaError& e) {
      outputError(OGCAPI_CONFIG_ERROR, "InjaError error. " + std::string(e.what()) + " (" + std::string(filename) + ").");
      return;
  } catch(...) {
    outputError(OGCAPI_SERVER_ERROR, "General template handling error.");
    return;
  }
```

I could then get the actual error which showed it was a path issue. I'm not sure if this additional error should be added for security reasons (it is giving away the path), but it might be good to add additional MapServer logging here. 

{"code":"ConfigError","description":"InjaError error. [inja.exception.file_error] failed accessing file at 'C:/MapServer/apps/html-plainlanding.html' (landing.html)."}

The issue was I had the following setting:

`"oga_html_template_directory" "C:/MapServer/apps/html-plain"`

I had to change this to the following path (note the escaped slash):

`"oga_html_template_directory" "C:/MapServer/apps/html-plain//"`

3. One of the hyperlinks in the HTML pages was set to http://localhost/mapserver/features/mymap/ogcapi/collections/Projects/collections/Projects?f=json rather than http://localhost/mapserver/features/mymap/ogcapi/collections/Projects?f=json
I'll try and find which template to fix and add a pull request.

sdlime:
@geographika, thanks so much for the detailed review - you should be able to run the sdlime/config branch as well for testing.  We should probably turn 1 and 2 into tickets...

1. This could warrant discussion but I can see what WFS does.
2. I think there's a todo in the code about adding a trailing slash if it's missing. Can I assume your setting example is from a metadata block and not an environment variable?
3. If your want to look quickly that's fine, otherwise I can find it. I assuming the issue exists in both the bootstrap 4 and plain template sets.

PR #6187, opened 2020-12-02
remove USE_PROJ check for mode=tile

jmckenna:
- this removes the USE_PROJ check in maptile.h, so that mode=tile can be enabled
 - also adds a dedicated test in msautotest for mode=tile
 - tested on Ubuntu, Windows for master and branch-7-6 (both the MODE=tile normally, and also through msautotest)
 - fixes MapServer/MapServer/issues/6178


jmckenna:
credit @sdlime for the memory link find (!).

Issue #6208, opened 2021-01-18
Wfs is rounding the output decimals to 6 numbers


pascal-codetaal:
We have a postgres database with some geom data, when importing the wfs to qgis we can see a small offset.

![alt text](https://i.ibb.co/dWgW1PW/Screenshot-2021-01-18-at-13-59-54.png)

This is the geom in the postgres database
`SRID=4326;MULTIPOLYGON(((4.346174668764422 51.0581301240931,0 4.34616114951401,51.05812865886134 0,4.346153577981363 51.05815530900463,0 4.346167288348252)))`

When calling the wfs the decimals are rounded and truncated to 6 decimals.
`<gml:posList srsDimension="2">51.058130 4.346175 51.058129 4.346161 51.058155 4.346154 51.058157 4.346167 51.058130 4.346175 </gml:posList>`

When we ask for the output OGRGML we can pass the LCO:COORDINATE_PRECISION and those are correct.
But with the default wfs we aren't able to change the precision

so following map we are using
```md
MAP

    OUTPUTFORMAT
        NAME "OGRGML"
        DRIVER "OGR/GML"
        FORMATOPTION "LCO:COORDINATE_PRECISION=10"
      END

    OUTPUTFORMAT
        NAME "wfs"
        DRIVER "OGR/GML"
        MIMETYPE "application/json; subtype=geojson; charset=utf-8"
        FORMATOPTION "STORAGE=stream"
        FORMATOPTION "FORM=SIMPLE"
        FORMATOPTION "USE_FEATUREID=true"
        FORMATOPTION "LCO:COORDINATE_PRECISION=15"
    END

    LAYER
        NAME "begraafplaatsen_qgis3__dd84521e_f3b2_4763_9d85_b5a6b41ad756"
        STATUS ON
        TYPE polygon
        EXTENT 4.346118494380637 51.057783128543726 4.34689792795098 51.058230468179445
        METADATA
          "ows_title" "begraafplaatsen_qgis3__dd84521e_f3b2_4763_9d85_b5a6b41ad756"
          "ows_abstract" "begraafplaatsen_qgis3__dd84521e_f3b2_4763_9d85_b5a6b41ad756"
          "wfs_getfeature_formatlist" "wfs, ogrgml"
          "gml_featureid" "id"
          "gml_include_items" "all"
          "gml_types" "auto"
          "wfs_srs" "EPSG:4326" 
        END
        CONNECTIONTYPE POSTGIS
[...]

PR #6209, opened 2021-01-20
WFS precision for geometry

jaspervercnocke:
adding wfs_geometry_precision parameter to metdata of wfs (web and layer based).

Related issue: #6208


msautotest/wxs/wfs_postgis_10f_precision.map
```diff
@@ -0,0 +1,71 @@
...
+  #
+  
+  LAYER
+    NAME popplace
+    INCLUDE "postgis.include"
+    DATA "the_geom from (select * from popplace order by gid) as foo using srid=3978 using unique gid"
+    METADATA
+      "wfs_title"         "popplace"
+      "wfs_description"   "Cities"
+      "gml_include_items" "auto"
```
jmckenna, line 60: this should be "gml_include_items" "all"  (auto is invalid)

msautotest/wxs/wfs_postgis_4f_precision.map
```diff
@@ -0,0 +1,100 @@
...
+  #
+  
+  LAYER
+    NAME popplace
+    INCLUDE "postgis.include"
+    DATA "the_geom from (select * from popplace order by gid) as foo using srid=3978 using unique gid"
+    METADATA
+      "wfs_title"         "popplace"
+      "wfs_description"   "Cities"
+      "gml_include_items" "auto"
```
jmckenna, line 61: this should be "gml_include_items" "all"  (auto is invalid)

msautotest/wxs/wfs_postgis_4f_precision.map
```diff
@@ -0,0 +1,100 @@
...
+
+  LAYER
+    NAME popplace_10
+    INCLUDE "postgis.include"
+    DATA "the_geom from (select * from popplace order by gid) as foo using srid=3978 using unique gid"
+    METADATA
+      "wfs_title"         "popplace"
+      "wfs_description"   "Cities"
+      "wfs_geometry_precision" "10"
+      "gml_include_items" "auto"
```
jmckenna, line 81: this should be "gml_include_items" "all"  (auto is invalid)

msautotest/wxs/wfs_postgis_10f_precision.map
jaspervercnocke: thanks for the remark. I modified it in the last commit

msautotest/wxs/wfs_postgis_4f_precision.map
jaspervercnocke: thanks for the remark. I modified it in the last commit

msautotest/wxs/wfs_postgis_4f_precision.map
jaspervercnocke: thanks for the remark. I modified it in the last commit

msautotest/wxs/wfs_postgis_10f_precision.map
```diff
@@ -0,0 +1,71 @@
+#
+# Test rounding geometry floating point number to 10 decimal places support in PostGIS input and GML/WFS output
+#
+# REQUIRES: SUPPORTS=WFS INPUT=POSTGIS
+#
+#
+# RUN_PARMS: wfs_postgis_10f_precision_gml2.xml [MAPSERV] QUERY_STRING="map=[MAPFILE]&SERVICE=WFS&VERSION=1.0.0&REQUEST=GetFeature&TYPENAME=popplace" > [RESULT]
```
rouault, line 7: you could possibly add &MAXFEATURES=1 in the requests to keep the size of the resulting files smaller (having more than 1 features doesn't add more to the testing)

geographika:
Looks good to me @jaspervercnocke - thanks. 

For reference (and possibly to add to the docs), as I understand it:

1. This update will only affect WFS requests using the built-in GML output. Any WFS requests with a different `OUTPUTFORMAT` (e.g. GeoJSON using OGR) will ignore the new `wfs_geometry_precision` setting. 
2. Other `LAYER` `METADATA` GML geometry settings use the format `gml_[geometry name]_XXXX`. For an example of reading these values see https://github.com/MapServer/MapServer/blob/6883f0227d3ffa1b3f15f201c6ef9dca80c134e7/mapgml.c#L959
 However, `WEB` `METADATA` doesn't have any `gml_` settings, so there would be inconsistencies with either syntax approach.

msautotest/wxs/wfs_postgis_4f_precision.map
```diff
@@ -0,0 +1,100 @@
...
+      "gml_include_items" "all"
+      
+     END
+    TYPE POINT
+    STATUS ON
+    PROJECTION
+      "init=epsg:3978"
+    END
+  
+    DUMP TRUE
```
geographika, line 90: The `DUMP TRUE` has been deprecated since version 6.0 so can be removed from the test Mapfiles

msautotest/wxs/wfs_postgis_10f_precision.map
jaspervercnocke: thanks for the remark. I modified the test in the last commit

msautotest/wxs/wfs_postgis_4f_precision.map
jaspervercnocke: thanks for the remark. I modified the test in the last commit

PR #6211, opened 2021-01-20
Enable offset for polygons

rouault:
Rebase of #6132


jmckenna:
Thanks @rouault for also including the tests

PR #6222, opened 2021-01-30
Add MapScript Python 3.8 Support

geographika:
As per #6101

This adds Windows support for Python MapScript 3.8, and testing under Appveyor. 
Also tests PHP 7.4 and Python 3.8 on Ubuntu with Travis. 

CI build times will increase with this pull request. Python 3.6 is EOL at the end of 2021 so perhaps this could be dropped from the CI?


jmckenna:
thanks @geographika !

Issue #4673, opened 2013-06-26
Track RFC98: Text Placement Refactoring


tbonfort:
Track RFC98 implementation
- [x] remove GD support (RFC99)
- [x] remove ANNOTATION layers
- [x] implement a global freetype glyph/font cache
- [x] agg and cairo renderer glyph handling rewrite
- [x] truetype symbol marker placements
- [x] truetype line markers (gap != 0)
- [x] baseline adjustments
- [x] iconv/bidi
- [x] harfbuzz shaping
- [x] labelcache speedups
- [x] leader labels
- [x] text alignment
- [x] bitmap labels
- [x] non labelcache'd layers (kml)
- [ ] #4408 labelpoly scalefactor

NathanW2:
@tbonfort I was just wondering for this item "implement a global freetype glyph/font cache" does that mean that there should be a single global font cache for all threads?  Seems the code only uses it if not using `USE_THREAD`

```
ft_cache* msGetFontCache() {
102
#ifndef USE_THREAD
103
  return &global_ft_cache;
104
#else
```

szekerest:
@NathanW2 I assume this has been implemented to avoid the cost of thread locking in exchange for the cost of duplicating the cache elements (for the same file). It may also be OS dependent that keeping the file open or limit the lifetime of the open handles would be more sufficient. I think on a production server with high load (ie. high number of thread in a process and high number of drawing sessions per second) limiting the lifetime of the open handles would be more scalable.

NathanW2:
The locking be from the cache being read/write vs just read right?

szekerest:
Not sure what you mean, we might either read or write the cache structure in a single operation.

PR #6224, opened 2021-02-01
Document labelCacheObj and associated objects (RFC 132)

geographika:
This pull request documents labelCacheObj and associated objects properties, which can then be used to generate SWIG API docs as part of [RFC 132](https://mapserver.org/development/rfc/ms-rfc-132.html)

Other changes are as follows:

- The current MapScript API docs don't currently reflect changes made to the labelling related objects  since the [RFC98 changes](https://github.com/MapServer/MapServer/issues/4673). There is a 'new' (from 2013) `labelCacheSlotObj` which now has several of the properties previously associated with `labelCacheObj`: cachesize, markercachesize, numlabels, nummarkers. 

- A deprecated property of `labelCacheMemberObj` and related code has been removed as annotation layers were dropped several versions ago. There is mention of the [include_deprecated](https://github.com/MapServer/MapServer/search?q=include_deprecated) anywhere else in build scripts or the codebase. 

```
#ifdef include_deprecated
    int shapetype; /* source geometry type, can be removed once annotation layers are dropped */
#endif
```

- labelCacheMemberObj > textSymbolObj - hidden as array would need typemaps for access

### DIFF
--- a/mapfile.c
+++ b/mapfile.c
@@ -6042,10 +6042,6 @@ int msFreeLabelCacheSlot(labelCacheSlotObj *cacheslot)
       }
       msFree(cacheslot->labels[i].textsymbols);
 
-#ifdef include_deprecated
-      for(j=0; j<cacheslot->labels[i].numstyles; j++) freeStyle(&(cacheslot->labels[i].styles[j]));
-      msFree(cacheslot->labels[i].styles);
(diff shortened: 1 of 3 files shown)


jmckenna:
thanks @geographika !

PR #6225, opened 2021-02-03
Issue #5508 - Revived ANTIALIAS keyword in STYLE section

rouault:
https://github.com/MapServer/MapServer/pull/5514 rebased on top of latest master with tab fixes and addition of expected results


jmckenna:
thanks @rouault for also including the new tests

geographika:
I had this listed for removal in https://mapserver.org/development/rfc/ms-rfc-133.html - I guess this should not be the case. 

[Docs](https://mapserver.org/mapfile/style.html#mapfile-style-antialias) currently talk about the GD renderer:

`Should TrueType fonts be antialiased. Only useful for GD (gif) rendering. Default is false. Has no effect for the other renderers (where anti-aliasing can not be turned off).`

Would the following be correct?

`Should antialiasing be applied to the style? For line rendering only`

rouault:
> Would the following be correct?
> 
> `Should antialiasing be applied to the style? For line rendering only`

looks good to me

jmckenna:
thanks @rouault and @erik-h11 !!

rouault:
> thanks @rouault and @erik-h11 !!

I would have used the "squash & merge" button to have a cleaner commit history

PR #6226, opened 2021-02-04
Oracle/MSSQL/OGR/PostGIS backends: avoid potential buffer overflows …

rouault:
…on large values of type double in expression translation

Based on initial work by @sergmorev done in https://github.com/MapServer/MapServer/pull/5416


jmckenna:
thanks @rouault & @sergmorev !!

PR #6229, opened 2021-02-08
When using OFFSET -1 -99 the shifted geometry can be a multiline containing multilines

wicksell:
While generating tiles for a line layer having a style containing "OFFSET -1 -99" we often get the following error.

msGEOSError(): GEOS library error. Geometry must be a Point or LineString

After some investigation we found that this is because the shifted GEOS geometry can be a multiline containing multilines which is not taken into account when converting it into a shape.

The issue can be reproduced with the following command and MAP-file.

```
shp2img -m test.map -o test.png -all_debug 5
msLoadMap(): 0.000s
msDrawMap(): rendering using outputformat named png24 (AGG/PNG).
msDrawMap(): WMS/WFS set-up and query, 0.000s
msGEOSError(): GEOS library error. Geometry must be a Point or LineString
```

```
MAP
    NAME "Test"
    EXTENT 387200 6540064 399488 6552352
    SIZE 1536 1536
    IMAGETYPE "png24"
    STATUS ON
    DEFRESOLUTION 96
    RESOLUTION 96
    UNITS METERS
  
    OUTPUTFORMAT
        NAME "png24"
        MIMETYPE "image/png"
        DRIVER "AGG/PNG"
        EXTENSION "png"
        IMAGEMODE RGBA
        TRANSPARENT TRUE
    END # OUTPUTFORMAT

    PROJECTION
        "proj=utm"
        "zone=33"
        "ellps=GRS80"
        "towgs84=0,0,0,0,0,0,0"
        "units=m"
        "no_defs"
    END # PROJECTION
  
    LAYER
        NAME "Test"
        PROJECTION
            "proj=utm"
            "zone=33"
            "ellps=GRS80"
            "towgs84=0,0,0,0,0,0,0"
            "units=m"
            "no_defs"
        END # PROJECTION
        STATUS ON
        TYPE LINE
        FEATURE
            POINTS
                387184.7628 6544765.9766500005
                387189.844 6544759.04235
                387215.864 6544756.3572
                387212.7508 6544765.81515
                387183.058 6544782.9486
                387186.706 6544811.9778
            END # POINTS
        END # FEATURE
  
        CLASS
            STYLE
                COLOR 255 0 0
                OFFSET -1 -99
            END # STYLE
        END # CLASS
    END # LAYER
END # MAP
```


jmckenna:
thanks @wicksell   Please also add a new test into msautotest to handle this case (follow steps at https://mapserver.org/development/tests/autotest.html) as part of your pull request.

(thanks for using inline features for the test case, great!)

jmckenna:
Thanks @wicksell for including the test.  This will be available as part of the upcoming 8.0 release    Thanks for contributing.

PR #6234, opened 2021-02-15
Filter out any invalid PATHs when loading Python MapScript

geographika:
This avoids import errors on invalid or empty `MAPSERVER_DLL_PATH` settings.

Also add a Troubleshooting section to the ReadMe that will be displayed at https://pypi.org/project/mapscript/


jmckenna:
thanks @geographika !

PR #6237, opened 2021-02-21
Add labelObj and layerObj comments to mapserver.h

geographika:
This pull request adds comments to the `layerObj` as part of https://mapserver.org/development/rfc/ms-rfc-132.html
Struct variables have been reordered based on SWIG or not SWIG and immutable/mutable. This reduces the many `#ifdef` scattered throughout the declaration (any issues with this please let me know). 

The following have been made immutable as they are custom types that would require getters/setters to be added to work correctly:

```
expressionObj utfdata;
LayerCompositer *compositer;
hashTableObj connectionoptions;
```

If `connectiontype`  is not set using the SWIG function `setConnectionType` the vtable is not initialized properly - however it is left as mutable as it can be used for INLINE layers and there are tests around this. 

Additional changes in the pull request:

- Remove the `**immutable**` flags from the comments - this information can be retrieved automatically when generating the Sphinx API docs
- add a header comment for `labelCacheSlotObj`

### DIFF
--- a/mapserver.h
+++ b/mapserver.h
@@ -1314,6 +1314,9 @@ typedef struct labelObj labelObj;
   /************************************************************************/
   /*                          labelCacheSlotObj                           */
   /************************************************************************/
+  /**
+  A cache slot to hold labels grouped by priority
+  */
   typedef struct {
     int numlabels; ///< Number of label members
     int cachesize; ///< TODO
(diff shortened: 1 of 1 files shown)


jmckenna:
looks good here.  A lot cleaner for sure.  Thanks @geographika !

PR #6238, opened 2021-02-23
Document legendObj and lineObj

geographika:
Documents the legendObj and lineObj structs to make available for autogenerated SWIG docs as part of https://mapserver.org/development/rfc/ms-rfc-132.html

No hidden surprises or changes, although legendObj has a couple of properties not available to SWIG that can be removed when deprecated).


jmckenna:
thanks @geographika !!

PR #6240, opened 2021-02-24
drop PHP 7.1 + Python 2.7 from TravisCI builds

jmckenna:
- PHP 7.1 was end-of-life in 2018, and ended security fixes at end of 2019
- Python 2.7 ended security releases at end of 2019

Remaining tests are:
- PHP 7.4, 7.3, 7.2
- Python 3.8, 3.7, 3.6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,6 @@ language: php
 matrix:
   include:
 
-    - php: 7.1
-      env:
-        - BUILD_NAME=PHP_7.1
-        - PYTHON_VERSION=2.7
-
     - php: 7.2
       env:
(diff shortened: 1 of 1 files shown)


geographika:
@jmckenna - is this to speed up the Travis builds or part of dropping Python 2.7 support?
I think we can support Python 2.7 as long as SWIG does. 
Note Python 2.7 will still be tested on Appveyor, so maybe that is enough.

jmckenna:
@geographika my goal here was to remove long end-of-life versions from the list of tests on TravisCI.  A nice benefit is that it cleans up the total number of tests, and takes a little shorter to complete as you mentioned.  My goal was not to 'officially' remove support for any versions.

PR #6242, opened 2021-02-28
Update mapObj comments and docstring fixes

geographika:
Documents the mapObj struct to make available for autogenerated SWIG docs as part of https://mapserver.org/development/rfc/ms-rfc-132.html
Reorders properties to reduce the `#ifdef SWIG` occurrences. 

Also fixes a few of the method docstrings in the .i SWIG files. 

No changes to immutable etc. as part of this pull request, although a few properties should be investigated further:

- `height` and `width` can be set by SWIG. However the Python bindings have custom methods added to set these by the size property: https://github.com/MapServer/MapServer/blob/e0372c34f5bfb3f539d2737fd98c9f68f9de7d2a/mapscript/python/pyextend.i#L276
- `int *layerorder;` can be set via SWIG but I think should be at least immutable and propbably accessed via methods - again this is the case for the Python bindings: https://github.com/MapServer/MapServer/blob/e0372c34f5bfb3f539d2737fd98c9f68f9de7d2a/mapscript/python/pyextend.i#L256

### DIFF
--- a/mapscript/python/pyextend.i
+++ b/mapscript/python/pyextend.i
@@ -240,9 +240,9 @@ def fromstring(data, mappath=None):
 
 %extend mapObj {
   
-    /* getLayerOrder() extension returns the map layerorder as a native
-     sequence */
-
+    /**
+    \**Python MapScript only**  - returns the map layer order as a native sequence
+    */
     PyObject *getLayerOrder() {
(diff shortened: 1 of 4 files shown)


jmckenna:
thanks @geographika !

PR #6244, opened 2021-03-02
Document outputFormatObj, projectionObj and map primitives

geographika:
Documents properties for `outputFormatObj`,`projectionObj`, and the map primitives `rectObj`, `lineObj`, `pointObj` and `shapeObj` as part of https://mapserver.org/development/rfc/ms-rfc-132.html


jmckenna:
thanks @geographika !

PR #6245, opened 2021-03-04
Document webObj, styleObj and referenceMapObj

geographika:
Document `webObj`, `styleObj` and `referenceMapObj` as part of https://mapserver.org/development/rfc/ms-rfc-132.html

Going through these led to noticing a couple of missing Mapfile keywords in the docs:

- https://github.com/MapServer/MapServer-documentation/issues/463
- https://github.com/MapServer/MapServer-documentation/issues/462
- https://github.com/MapServer/MapServer-documentation/issues/461


jmckenna:
thanks @geographika !

PR #6246, opened 2021-03-05
msGetFriBidiEncodedString(): error out on multi-line paragraph

rouault:
Refresh of https://github.com/MapServer/MapServer/pull/5535 on top of current main


jmckenna:
thanks @rouault and @qwef !!

PR #6254, opened 2021-03-09
OGR output: do not force MultiPolygon to Polygon for a TYPE POLYGON layer, 

### LINKED ISSUE #6235: WFS MultiPolygon & GeoJSON
Hi,

Version: Mapserver 7.6
Protocol: WFS
Output format: GeoJSON

Issue: Multipolygons are replaced  by Polygons with multiparts. This is not compliant with GeoJSON specifications and it causes many errors when you want to use the geometries in computing libraries like TurfJS or others.

Thank you.

rouault:
…unless wfs_geomtype is set to Polygon (fixes #6235)


jmckenna:
great enhancement, thanks @rouault !

Issue #6258, opened 2021-03-12
Performance Issues when using WMS Filters


geographika:
WMS filters were added in [RFC 118](https://mapserver.org/development/rfc/ms-rfc-118.html). Allowing WMS requests to be filtered, for example:

```
mapserv -nh "QUERY_STRING=map=my.map&SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&FORMAT=image/png&TRANSPARENT=true&STYLES=Type&TILED=false&LAYERS=MyLayer&ORDERBY=StartTime&FILTER=(<Filter><PropertyIsEqualTo><PropertyName>MyField</PropertyName><Literal>My Property</Literal></PropertyIsEqualTo></Filter>)&CRS=EPSG:3857&WIDTH=1571&HEIGHT=843&BBOX=-1230789.0294479083,6622962.752797382,-750458.7437038606,6880708.412174996"
```

For a typical WMS request the `msLayerWhichItems` function gathers all the fields required to display the WMS e.g. for classes, labels, filters etc. and does not request any additional fields.

https://github.com/MapServer/MapServer/blob/20b7671f788952c6097ae154b667dd564cf70764/maplayer.c#L873

When a `FILTER` is added then *all* fields for the layer are requested, even though they may not be used:

https://github.com/MapServer/MapServer/blob/2bc7f625d7ec45d8401aeb66ae908b5210acf83f/mapquery.c#L835

I have a layer with ~50 fields. When displaying with just the required fields it takes < 1s. When all fields are requested it takes over 6 minutes. This is a similar issue to #6131 for WFS. 

I propose that if in the `msWMSApplyFilter` a metadata property is set so that returning all fields can be skipped later. Trying to pass a flag from this function would have to be passed through 6 functions to reach to the `msQueryByFilter` so modifying the metadata would be less intrusive. `msWMSApplyFilter` is not called when using `GetFeatureInfo`, so these requests would be unaffected (and continue to return all fields, which is less of an issue as this typically only has to query a handful of features). 

https://github.com/MapServer/MapServer/blob/207594b6685047269d8c2f1289c26644eec36787/mapwms.cpp#L303

geographika:
I've been running this on staging servers for a couple of weeks without issues (apart from a huge performance boost). 
If there are no objections I'd like to merge this soon.

jratike80:
+1 for the boost.

geographika:
Thanks all for the feedback/comments! Closed with #6259

PR #6259, opened 2021-03-12
Improve WMS Filter performance

geographika:
Fix for #6258

All works fine in my setup and all tests are passing in CI, but I'd appreciate any reviews prior to merging, in case there are unforeseen implications of the update. 

- It does add a new `#include "mapows.h"` to `maplayer.c`. 
- `gml_wmsfilter_flag` was the best I could come up with for this temporary metadata item. It is added and then removed once the query has run. 

Note the metadata is only updated in the `msWMSApplyFilter` which is only used when filters are run on WMS so other code paths should be unaffected.

### DIFF
--- a/maplayer.c
+++ b/maplayer.c
@@ -32,7 +32,7 @@
 #include "mapogcfilter.h"
 #include "mapthread.h"
 #include "mapfile.h"
-
+#include "mapows.h"
 #include "mapparser.h"
 
 #include <assert.h>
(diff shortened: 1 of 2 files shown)


jmckenna:
thanks @geographika !!

rouault:
I should have commented before: has it been verified this doesn't affect GetFeatureInfo requests ?

geographika:
> I should have commented before: has it been verified this doesn't affect GetFeatureInfo requests ?

I can confirm `GetFeatureInfo` requests are unaffected by this change.

rouault:
> I can confirm `GetFeatureInfo` requests are unaffected by this change.

ok, I meant the comination of GetFeatureInfo + WMS filter. I see msautotest/wxs/wms_filter.map only tests GetMap

geographika:
> ok, I meant the comination of GetFeatureInfo + WMS filter. I see msautotest/wxs/wms_filter.map only tests GetMap

To expand further - if a WMS is filtered then a feature that is filtered out will not be returned by `GetFeatureInfo` (which is the existing and unchanged behaviour). If a WMS feature on a filtered layer is selected via `GetFeatureInfo` then all fields in the layer are returned as currently. 

I'll look at adding a test for the combination of GetFeatureInfo + WMS filter.

rouault:
> If a WMS feature on a filtered layer is selected via `GetFeatureInfo` then all fields in the layer are returned as currently.

yes, that was the part I was unsure about your patch.

PR #6260, opened 2021-03-12
WFS: Only display cache debug messages for MS_DEBUGLEVEL_V

geographika:
Currently, they are displayed even when the Mapfile has `DEBUG 0`. 
With this update they are displayed for `DEBUG 3` and higher to avoid cluttering logs when reporting errors only.


jmckenna:
thanks @geographika !

PR #6263, opened 2021-03-16
Use PointZM throughout MapServer

geographika:
This pull request removes the build option `WITH_POINT_Z_M` and all code paths using `USE_POINT_Z_M` so that PointZMs are used throughout MapServer. All CI and build scripts updated to remove this flag. 

See previous mailing list discussion [here](https://lists.osgeo.org/pipermail/mapserver-dev/2019-February/015610.html), [here](https://lists.osgeo.org/pipermail/mapserver-dev/2019-March/015611.html)  and the related issue #5456

Advantages:
- Simpler code paths, and easier to test (`WITH_POINT_Z_M` was used for all CI testing so use of non-PointZM was untested)
- Fixes issues with MapScript bindings and testing where the output is not guaranteed due to the compile options and could lead to memory corruption issues (from @szekerest)

Disadvantages:
- Up to ~2% decrease in performance compared to non-PointZM as noted on [this 2017 post](https://lists.osgeo.org/pipermail/mapserver-users/2017-July/080102.html) from @rouault. Note most builds have been using `WITH_POINT_Z_M` `ON` for some time so there will be no performance implications for any applications using these. 

This pull request also fixes a previously undiscovered bug in MapScript (Z and M values for points added to lines were not being set).


jmckenna:
MS4W builds have set this on by default for many years, because of so many requests by users.  Thanks for the detailed examination @geographika

PR #6264, opened 2021-03-16
Remove DUMP keyword from LAYER - RFC133

geographika:
First pull request related to https://mapserver.org/development/rfc/ms-rfc-133.html
This removes the `DUMP` [keyword](https://mapserver.org/mapfile/layer.html#mapfile-layer-dump) from `LAYER`.
This was deprecated in version 6.0 and ` LAYER METADATA` is used instead, so `DUMP` has no currently has no effect, and can be removed for clarity. 
If merged a pull request will be added to the docs.


jmckenna:
@geographika let's give this one until end-of-day tomorrow, which will be 2 full business days from your PSC reminder.  thanks again and have a great long weekend!

jmckenna:
thanks @geographika !  Let's proceed...

PR #6265, opened 2021-03-19
WIP: Reduce log clutter from non-errors

geographika:
Add back #5707 now tests are fixed.


geographika:
Ok it is a little more complicated. The original pull request checks the `LAYER`>`DEBUG` level, but there can be multiple layers and they are closed in the previous loop. Updated so the messages are reported based on the `MAP`>`DEBUG` level. 
Willl also update Mapfiles so msautotest passes.

geographika:
@rouault - just looking through changes to msautotests and came across https://github.com/MapServer/MapServer/commit/58bbb51a9978ff48a5181b1817786077905c79ac - I'm guessing the `LAYER DEBUG 3` settings were removed as they were only added for the reverted commit?
The no result "error" messages only seem to be useful for the tests. On a live server any WMS GetFeatureInfo which misses a feature generates a record in the logs even with `DEBUG OFF` at the moment.

rouault:
@geographika  Upon reflection, removing those if( lp->debug >= MS_DEBUGLEVEL_V ) was wrong. I think they should be indeed re-added back, and the results of the few failed tests where they are found adjusted. A grep shows that they are just wxs/expected/wfs_200_exception_getfeaturebyid_nofeature.xml and wxs/expected/wfs_200_exception_getfeature_resourceid_inexisting.xml.
But we probably don't need to re-add DEBUG statements in map files.

geographika:
Replaced with #6268 - using a log level to set errors seems to be the wrong approach. Queries that return 0 results should be seen as valid queries IMO.

Issue #4852, opened 2014-01-22
Replace OperationProcessingFailed by NotFound in a WFS 2.0 GetFeatureById request without result


rouault:
Currently we emit a "403 Server processing failed" HTTP status code and a "OperationProcessingFailed" exception code when a GetFeatureById request has no match for the requested ID.
It would seem much more appropriate to emit a "404 Not Found" itself with a "NotFound" exception code.
This is accepted by the invokeGetFeatureByIdWithUnknownID() test of the WFS 2.0 CITE suite, but not explicitely by the WFS 2.0 spec.
Waiting for feedback from the WFS SWG.

PR #6268, opened 2021-03-20
Return MS_SUCCESS rather than MS_FAILURE for queries with 0 results

geographika:
An update/alternative to #6265. This changes a response to a query which returns no results from `MS_FAILURE` to `MS_SUCCESS` but with 0 results. I'm unsure what the WxS specs say, but returning server errors for a query with no results causes issues in client applications, and currently fills up MapServer logs with "errors" which could be caused simply by a user filtering a dataset. 

Logging added for capturing these "no results" if required. Tests have been updated to reflect the changes. 

For `wfs_200_exception_getfeaturebyid_nofeature` the issue at #4852 and commit referenced in the issue about changing the response to a 404 and `Not Found` error rather than `OperationProcessingFailed` make a lot of sense to me.


rouault:
Looks good to me

geographika:
Thanks @rouault for the review.
Is there any reason why https://github.com/rouault/mapserver/commit/89919a719ea2bb8a477590e0fb4d2e04f65ad2d6 wasn't merged? It seems a good approach - I can rebase/check tests if it makes sense to add it.

rouault:
My memories rarely go beyond 7 hours, so 7 years ago... :-) From the ticket, it seemed I was waiting from feedback from the WFS SWG. I guess I never got. The commit makes sense AFAICS

jmckenna:
Makes sense to me also.  Thanks @geographika !

Issue #6239, opened 2021-02-24
FindPostgreSQL.cmake finds different library than pg_config reports.


offroad-bis:
macOS 10.15.7
MapServer 7.6.2
cmake version 3.19.5

A bug in FindPostgreSQL.cmake?  if PG_CONFIG is found, then variables are set in cmake to the include and lib dirs specified by pg_config.  If not found, it issues a warning that says it will try some defaults. 

```
if (PG_CONFIG)
   exec_program( ${PG_CONFIG} ARGS "--includedir" OUTPUT_VARIABLE PG_INC_PATH )
   exec_program( ${PG_CONFIG} ARGS "--libdir" OUTPUT_VARIABLE PG_LIB_PATH )
else()
   message(WARNING "pg_config not found, will try some defaults")
endif()
```

But what it does, is try the defaults anyway, which may not be what pg_config reports.  On my machine, pg_config outputs the following.
```
INCLUDEDIR = /usr/local/include
LIBDIR = /usr/local/lib
```
But cmake finds this instead, which I assume comes from the default search that cmake uses prior to using HINTS and PATHS.

`* POSTGIS: /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/lib/libpq.tbd`

Was the intended action to use the library and include directories specified by the pg_config executable?  If so, should the  if block for PG_CONFIG be changed so that, when pg_config is found, it searches only the libraries and include paths output by pg_config.  And if pg_config is not found, then search the default paths.  Without the NO_DEFAULT_PATH directive, the find_path will search the [default paths](https://cmake.org/cmake/help/latest/command/find_path.html) before ever searching the PATHS specified by pg_config.  

```
if (PG_CONFIG)
   exec_program( ${PG_CONFIG} ARGS "--includedir" OUTPUT_VARIABLE PG_INC_PATH )
   exec_program( ${PG_CONFIG} ARGS "--libdir" OUTPUT_VARIABLE PG_LIB_PATH )
   find_path(POSTGRESQL_INCLUDE_DIR libpq-fe.h
      PATHS
      ${PG_INC_PATH}
      NO_DEFAULT_PATH  # Stops cmake from looking in default places.
   )
   find_library(POSTGRESQL_LIBRARY NAMES pq libpq
      PATHS
      ${PG_LIB_PATH}
      NO_DEFAULT_PATH # Stops cmake from looking in default places.
   )
else()
   message(WARNING "pg_config not found, will try some defaults")
   
[...]

PR #6269, opened 2021-03-21
Issue #6239. If pg_config is found, cmake searches paths specified by pg_config.

offroad-bis:
Issue #6239. This changes the IF block for PG_CONFIG.  When pg_config is found, cmake searches only the libraries and include paths output by pg_config.  If pg_config is not found, then it searches the defaults and other listed paths in the find_path and find_library commands.


jmckenna:
Thanks for this effort @offroad-bis ! It will be included in the upcoming 8.0 release.

PR #6270, opened 2021-03-21
msautotest: add test for UNION layer

rouault:


jmckenna:
Great catch @rouault !

Issue #6274, opened 2021-03-21
DEFRESOLUTION 0 causes GEOS errors


geographika:
As noted in #5748 if [DEFRESOLUTION](https://mapserver.org/mapfile/map.html#mapfile-map-defresolution) is set to 0 a divide by 0 can occur causing further errors and strange results e.g. `msGEOSError(): GEOS library error. @ÞÓV}` 

https://github.com/MapServer/MapServer/blob/ce2169f872009018c7bff1efa8853d7554669d23/mapimagemap.c#L275

`msDrawLineSymbol` uses the NaN and this eventually ends up as the `offset` parameter sent to GEOS which causes a library error. 

https://github.com/MapServer/MapServer/blob/d7d84dd072b4451beaf0e7505da1d9b383f4f210/mapgeos.c#L849

geographika:
Closed with #6275

PR #6275, opened 2021-03-21
Ensure DEFRESOLUTION > 0

geographika:
Possible fix for #6274
I'm unsure how much validation is/should be done on Mapfile parameters - I think this is the correct to place it though if deemed useful.


sdlime:
Seems ok to me - pretty straight forward check that follows an existing pattern. _Note:_ at some point we should really have getDouble() and getInteger() return MS_SUCCESS|MS_FAILURE for consistency - not part of this change. Good catch Seth!

geographika:
Thanks for the review @sdlime

Issue #6277, opened 2021-03-27
GetMetadata crashes on a CLUSTER layer


geographika:
Tested with a MSSQL layer with clustering. 

The `GetMetadata` command calls `msLayerGetExtent`, this in turn calls `layer->vtable->LayerGetExtent`. However on a layer with `CLUSTER`  `msClusterLayerOpen` is used rather than the standard layer opening - the source layer is copied to a new layer. Later on `layerinfo->conn` is `NULL` causing a crash when attempting to make database calls to get the extent. 

If the cluster layer has an `EXENT` already set there is no issue as there are no database calls.

geographika:
Crash fixed with #6278, although fixing the root cause of the issue was not fully resolved (see comments).

PR #6278, opened 2021-03-27
Add Cluster Layer GetMetadata fix with test

geographika:
Fix for `GetMetadata` calls on a database layer with no `EXTENT`. See #6277. 
Adds in a test used to recreate the error (with an additional cluster image test) for a MSSQL layer. 

The fix simply removes any clustering in `msMetadataGetLayerMetadata`. As clustering won't have an effect on the source metadata, and the function is only called by a `GetMetadata` this should have no side-effects. 

Attempting to modify `msClusterLayerOpen` to try and copy across connection objects etc. would be an alternative, and more complicated solution, but I'm unaware of any similar issues with other service calls and cluster layers.


jmckenna:
thanks @geographika !

mapmetadata.c
```diff
@@ -687,6 +687,11 @@ xmlNodePtr msMetadataGetLayerMetadata(mapObj *map, metadataParamsObj *paramsObj,
     if(strcasecmp(GET_LAYER(map, i)->name, paramsObj->pszLayer) == 0) {
         layer_found = MS_TRUE;
         layer = GET_LAYER(map, i);
+        // when checking a layer with clustering msLayerGetExtent does not have access
```
rouault, line 690: I have the feeling this isn't the right fix. That wouldn't fix the crash for someone calling msLayerGetExtent() through the C API or Mapscript, would it ?

msautotest/mssql/cluster_mssql.map
```diff
@@ -0,0 +1,51 @@
+#
+# Test a clustered layer with the MSSQL driver plugin
```
rouault, line 2: is the issue specific to using mssql ? Can't the be reproduced with shapefiles ?

mapmetadata.c
geographika: This function is only called using the `GetMetadata` request (see https://mapserver.org/ogc/layer_metadata.html) - there is no mapscript access currently.

msautotest/mssql/cluster_mssql.map
geographika: The issue is around a NULL `layerinfo->conn` in the database driver when trying to get the extent from the database. Shapefiles would be unaffected, but I'd image the PostGIS and Oracle drivers would have the same problem.

mapmetadata.c
rouault: > This function is only called using the `GetMetadata` request

yes, but isn't the issue fundamentally with msLayerGetExtent() called on a clustered layer ?

geographika: True..  Cluster layers are a special case in [msLayerOpen](https://github.com/MapServer/MapServer/blob/4505f7028b20f8d86734419450c2f3299bf76ace/mapcluster.c#L1622). Then a copy is made of the original layer in [msClusterLayerOpen](https://github.com/MapServer/MapServer/blob/4505f7028b20f8d86734419450c2f3299bf76ace/mapcluster.c#L1622), but it is the source layer vtable which is opened, but then the copied layer doesn't have the original connection - I had a hard time trying to follow exactly what is happening here. 
I've not had any cases of `msLayerGetExtent` calls crashing the server apart from through `GetMetadata`, but this may be luck/missing logs. The fix in this pull request prevents a server crash, but I agree it would be better to fix the underlying issue.

PR #6284, opened 2021-04-07
WIP: Require MS_MAP_NO_PATH or MS_MAP_PATTERN to be set.

sdlime:
Patch to improve default CGI security posture by requiring the environment variables MS_MAP_NO_PATH or MS_MAP_PATTERN to be set.


sdlime:
@geographika, @rouault - this will break the CI testing since those environment variables are not set. What do you think the best way to address will be? Could be as simple as setting MS_MAP_PATTERN to "." although it would be better to be more explicit.

It also would make sense to add tests with those environment vars unset or set to various values and then just do a series of "mapserv "QUERY_STRING=map=[value]" tests. Don't actually need a mapfile for those - the error messages are good enough output.

--Steve

rouault:
in msautotest/pytest.ini, you could add in the [pytest] section a new line
```
env =
    MS_MAP_PATTERN=.
```
But I'm wondering if there's not a way to detect that the mapserv binary is called from the command line (in which case we might want to skip the new checks) vs in a CGI environment. In CGI environment, https://tools.ietf.org/html/rfc3875 requires that SERVER_NAME is set.

sdlime:
Good point. Although, then we'd have to do something special then (e.g. set SERVER_NAME) to test this functionality from the command line. I suppose that's the rarer case though and would be less inconvenient for users. I updated the RFC draft I had started with your idea - there's a choice to be made.

I can update the pull request. If nobody weighs in then we can make an executive decision.

--Steve

rouault:
what I've not verified is that the popular CGI implementation s(apache, nginx, iis) do set SERVER_NAME. They *SHOULD* according to RFC3875.

geographika:
I've been testing again with environment variables and IIS/FastCGI, and have the same issues as described on the mailing list in 2015: http://osgeo-org.1560.x6.nabble.com/MapServer-Environment-Variables-and-IIS-td5186363.html

I think I've now resolved the mystery..! From the [MS docs](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/getenv-wgetenv?redirectedfrom=MSDN&view=msvc-160#remarks)

> getenv operates only on the data structures accessible to the run-time library and not on the environment "segment" created for the process by the operating system.

I have `getenv` working in MapCache. Checking the code the `getenv` is in the CGI application itself rather than in the DLL:

https://github.com/MapServer/mapcache/blob/447c0c6848fd101c11e6d54e451037e571117ef1/cgi/mapcache.c#L253

If I add the `getenv` to mapserv.c - the source used to create mapserv.exe then the FastCGI environment variable is picked up correctly - see https://github.com/MapServer/MapServer/commit/3beeff9c7f03ffcb458d1c59bc9701e3b863e364

When the same code is used in maphttp.c (which is used for mapserver.dll) `getenv` returns NULL. 

https://github.com/MapServer/MapServer/blob/cde09201ad2d469325bc911256513c5b9de78694/maphttp.c#L474

In summary `getenv` outside of `mapserver.c` can't be used with IIS and FastCGI.

rouault:
>  Checking the code the `getenv` is in the CGI application itself rather than in the DLL:

Excellent finding @geographika  . Windows is such a hostile territory ! I guess a potential workaround would be to statically link libmapserver inside mapserv binary.
However your findings have still myteries. For example loadParams() is in the cgiutil.c, which is in libmapserver and not mapserv binary, and loadParams() is invoked by main() with a getenv2==NULL, which means that getenv2=&msGetEnv is used, which finally uses getenv(). So REQUEST_METHOD retrieving should fail according to the above theory. Unless there's a difference in how the non-FastCGI env variable CURL_CA_BUNDLE is set and how the FastCGI one REQUEST_METHOD is.

geographika:
It only took me 6 years to understand what was happening :)
@rouault - I think all the [server variables](https://docs.microsoft.com/en-us/previous-versions/iis/6.0-sdk/ms524602(v=vs.90)) would be ok (including `SERVER_NAME` and `REQUEST_METHOD`). 

FastCGI environment variables are declared in `C:\Windows\System32\inetsrv\config\applicationHost.config`, and these are the ones ignored by `getenv`:

```xml
        <fastCgi>
            <application fullPath="C:\MapServer\bin\mapserv.exe" arguments="" monitorChangesTo="" stderrMode="ReturnStdErrIn500" maxInstances="4" idleTimeout="300" activityTimeout="30" requestTimeout="90" instanceMaxRequests="200" signalBeforeTerminateSeconds="0" protocol="NamedPipe" queueLength="1000" flushNamedPipe="false" rapidFailsPerMinute="10">
                <environmentVariables>
                    <environmentVariable name="GDAL_DATA" value="C:\MapServer\bin\gdal-data" />
                    <environmentVariable name="CURL_CA_BUNDLE" value="C:\Mapserver\bin\curl\curl-ca-bundle.crt" />                    
                </environmentVariables>
            </application>
```

I ended up hard-coding `CURL_CA_BUNDLE` in a DLL which needed a custom cert and HTTPS access to a WMS as a workaround in the past. 
@sdlime - sorry for hijacking the thread.. I think your proposal at https://github.com/sdlime/mapserver/wiki/MapServer-8.0-Config-File - loading all variables in mapserv.c via a config file or environ variables would resolve this. 

Without the config file approach however I'm not sure it would be possible to override the `MS_MAP_NO_PATH` variable on IIS/FastCGI.

sdlime:
Wondering if this should be pulled as a stand-alone pull request and just be bundled as part of the config file RFC.

My worry after reading this is that a certain segment of users can't set MS_MAP_NO_PATH or MS_MAP_PATTERN. There has to be a way to address that for current versions or not?

sdlime:
Closing, any changes in this area will be part of larger transition to a config file. --Steve

PR #6286, opened 2021-04-07
[7.6 branch] .travis: update to latest pip version supported by py2.7 to hopefully…

rouault:
… solve later pip install issues


rouault:
Droping this. I'm afrain we have to live with a half broken CI for the 7.6 branch due to using Trusty and deprecated python2.7. What is strange is that some of the CI config work and others don't. Must be related to some pip caching difference

PR #6288, opened 2021-04-10
Only select requested fields for all WFS query types

geographika:
This pull request builds on the work in #6131 to improve WFS GetFeature performance. 

It provides a significant performance boost for any WFS layers that are based on large database views, by only selecting the fields required by the client, set using the `propertyName` parameter. This prevents the database from running joins and returning data which is then later discarded (also similar to #6259). 

So for example previously a WFS GetFeature request with a `FILTER` would select all fields in the database layer - now it only selects a subset. In my test case a query for the top 20 records in a view with 51 fields went from 13.09 seconds to 0.798 seconds, with the number of fields selected dropping from 51 fields down to 17. 

#6131 only applied the performance boost for `msQueryByRect`. This update applies the speedup for all query types - `msQueryByIndex`, `msQueryByFilter`, `msQueryByRect`, `msQueryByFeatures`, `msQueryByPoint`, and `msQueryByShape`. 

`msLayerWhichItems` now checks for fields listed in `propertyName` which populates a metadata item `gml_select_items`. 
If this is populated and no other explicit fields are requested (in the `metadata` parameter, which is also used by WCS queries), then only the required fields are requested from the data source. 

This will apply to all datasources, but the database drivers (Postgres, MSSQL, Oracle) will get the most benefit from this update.

### DIFF
--- a/maplayer.c
+++ b/maplayer.c
@@ -990,11 +990,13 @@ int msLayerWhichItems(layerObj *layer, int get_all, const char *metadata)
       get_all = MS_FALSE;
   }
 
-  // item set by mapwfs.cpp to restrict the number of columns selected
-  metadata = msOWSLookupMetadata(&(layer->metadata), "G", "select_items");
-  if (metadata) {
-      /* get only selected items */
-      get_all = MS_FALSE;
+  if(metadata == NULL){
+      // check item set by mapwfs.cpp to restrict the number of columns selected
+      metadata = msOWSLookupMetadata(&(layer->metadata), "G", "select_items");
+      if (metadata) {
+          /* get only selected items */
+          get_all = MS_FALSE;
+      }
   }
 
   /* always retrieve all items in some cases */
--- a/maplayer.c
+++ b/maplayer.c
@@ -985,10 +985,14 @@ int msLayerWhichItems(layerObj *layer, int get_all, const char *metadata)
   if(layer->utfdata.type == MS_EXPRESSION || (layer->utfdata.string && strchr(layer->utfdata.string,'[') != NULL && strchr(layer->utfdata.string,']') != NULL))
     nt += msCountChars(layer->utfdata.string, '[');
 
+  // if we are using a GetMap request with a WMS filter we don't need to return all items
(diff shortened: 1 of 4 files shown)


sdlime:
So moving code out of msQueryByRect() to a more general location in msLayerWhichItems() - makes sense to me. --Steve

geographika:
Yes, I think originally only `msQueryByRect` was updated to have the minimum impact on the codebase. However `msLayerWhichItems` will only be affected when it is a WFS request with `propertyName` passed in to the querystring.

rouault:
It seems to me that the wmsfilter_flag and select_items metadata items serve the same purpose. Can't they be unified in a single setting ?

geographika:
> 
> 
> It seems to me that the wmsfilter_flag and select_items metadata items serve the same purpose. Can't they be unified in a single setting ?

Yes they effectively do the same thing from 2 different code paths so I'll use the same setting.

geographika:
Actually there are differences between the WMS filter and the WFS filter approaches. 
The WFS filter sets "gml_select_items" with a list of required fields that is then used for the metadata. 
The WMS filter (which sets "gml_wmsfilter_flag") simply has to ensure that `get_all` is set to `MS_FALSE` to avoid selecting all fields.

rouault:
couldn't the gml_wmsfilter_flag be replaced by setting gml_select_items to the empty string then ? Although I'm not super familiar with that part of the code, all that looks like hacks. I'd really prefer we minimize their number...

geographika:
When set via WFS, metadata gets set to a list of fields:

`metadata = msOWSLookupMetadata(&(layer->metadata), "G", "select_items");`

If `gml_select_items` is set to an empty string for a WMS with a filter then metadata will be set to "" and will be true for the condition below, and end up getting all fields from the layer:

https://github.com/MapServer/MapServer/blob/7e7464e06f96fbc7d31f090249ef0ed1602fc626/maplayer.c#L1014

In essence WFS needs a list of fields whereas WMS just needs to know not to get all fields for the layer. 
Maybe the metadata keys could have clearer names, but I can't come up with any.

rouault:
ok, let's go with that then.

Issue #6167, opened 2020-10-09
Mapserver getenv crashes (+ valgrind logs)


hege-li:
Hello,

We use a self-compiled stack of latest versions of mapserver (7.6.1), proj,
gdal, geos etc on RHEL7.  The production server is pretty busy and we see lots of
crashing related to getenv().

I tried running valgrind, had a hard time producing any crashes since it
seems to relate on query speed.  But with 25 concurrent connection flooding
I managed to get few dumps too..

(gdb) where
#0  0x0000000008e262bd in getenv () from /lib64/libc.so.6
#1  0x000000000753e7c1 in CPLGetConfigOption () at cpl_conv.cpp:1717
#2  0x0000000006db696c in GTiffDataset::Finalize (this=0x1c617970) at geotiff.cpp:7744
#3  0x0000000006db6f9a in GTiffDataset::~GTiffDataset (this=0x1c617970, __in_chrg=<optimized out>) at geotiff.cpp:7714
#4  0x0000000006db70c9 in GTiffDataset::~GTiffDataset (this=0x1c617970, __in_chrg=<optimized out>) at geotiff.cpp:7711
#5  0x0000000004f7fab1 in msGDALCleanup () at /home/wms/src/mapserver-7.6.1/mapgdal.c:90
#6  0x0000000004f134b7 in msCleanup () at /home/wms/src/mapserver-7.6.1/maputil.c:2105
#7  0x0000000000401554 in msCleanupOnSignal (nInData=<optimized out>) at /home/wms/src/mapserver-7.6.1/mapserv.c:64
#8  <signal handler called>
#9  0x0000000008eec61d in accept () from /lib64/libc.so.6
#10 0x00000000052f19ea in OS_Accept () from /lib64/libfcgi.so.0
#11 0x00000000052ef9ac in FCGX_Accept_r () from /lib64/libfcgi.so.0
#12 0x00000000052efae5 in FCGX_Accept () from /lib64/libfcgi.so.0
#13 0x00000000052efc6a in FCGI_Accept () from /lib64/libfcgi.so.0
#14 0x0000000000401234 in main (argc=<optimized out>, argv=<optimized out>) at /home/wms/src/mapserver-7.6.1/mapserv.c:137

I have a large package of valgrind logs if someone more experienced could
have a look?  You can download it here:

https://drive.google.com/file/d/1wgl7CgfBhyVyKMkqlOyNoAS6IZc1CEND/view?usp=sharing

Some random examples:

[...]

PR #6290, opened 2021-04-11
WIP: Copy environ variable to FastCGI environ

geographika:
More of a proof-of-concept than a pull request. 

This gets the `CURL_CA_BUNDLE` value from the environment prior to `FCGI_Accept`, and then puts it into the FastCGI environment so it can be accessed later:

https://github.com/MapServer/MapServer/blob/cde09201ad2d469325bc911256513c5b9de78694/maphttp.c#L474

This works ok, but there are lots of MapServer environment variables. Some list/loop could be used?
Alternatively maybe adding the variables to the `mapserv->map` object would make more sense (and perhaps the same approach could be reused as part of https://github.com/sdlime/mapserver/wiki/MapServer-8.0-Config-File).


rouault:
> Some list/loop could be used?

yep, declare "extern char **environ;" close to headers, iterate over it to make a copy of it, and then putenv in the fastcgi loop. But I would have expected the fastcgi code to do that for you...

mapserv.c
```diff
@@ -239,12 +238,33 @@ int main(int argc, char *argv[])
...
 
+  // make a copy of the original environment
+  int i = 0;
+  int j;
+  char **env = environ;
+
+  while (env[i] != NULL)
+      i++;
+
+  char **envCopy = malloc(sizeof(char*) * i + 1);
```
rouault, line 249: no need for +1 here:
```suggestion
  char **envCopy = malloc(sizeof(char*) * num_env_var);
```

mapserv.c
```diff
@@ -239,12 +238,33 @@ int main(int argc, char *argv[])
...
+
+  for (j = 0; j < i; j++) {
+      envCopy[j] = malloc(strlen(env[j]));
```
rouault, line 252: but here you need it
```suggestion
      envCopy[i] = malloc(strlen(env[i])+1);
```

mapserv.c
```diff
@@ -239,12 +238,33 @@ int main(int argc, char *argv[])
   signal( SIGTERM, msCleanupOnSignal );
 #endif
 
+  // make a copy of the original environment
+  int i = 0;
+  int j;
```
rouault, line 243:
```suggestion
```

mapserv.c
```diff
@@ -239,12 +238,33 @@ int main(int argc, char *argv[])
...
+  char **env = environ;
+
+  while (env[i] != NULL)
+      i++;
+
+  char **envCopy = malloc(sizeof(char*) * i + 1);
+
+  for (j = 0; j < i; j++) {
```
rouault, line 251:
```suggestion
  for (int i = 0; i < num_env_var; i++) {
```

mapserv.c
```diff
@@ -239,12 +238,33 @@ int main(int argc, char *argv[])
   signal( SIGTERM, msCleanupOnSignal );
 #endif
 
+  // make a copy of the original environment
+  int i = 0;
```
rouault, line 242:
```suggestion
  int num_env_var = 0;
```

mapserv.c
```diff
@@ -239,12 +238,33 @@ int main(int argc, char *argv[])
...
+  int j;
+  char **env = environ;
+
+  while (env[i] != NULL)
```
rouault, line 246:
```suggestion
  while (env[num_env_var] != NULL)
```

mapserv.c
```diff
@@ -239,12 +238,33 @@ int main(int argc, char *argv[])
...
+      i++;
```
rouault, line 247:
```suggestion
      num_env_var++;
```

mapserv.c
```diff
@@ -239,12 +238,33 @@ int main(int argc, char *argv[])
...
+
+  char **envCopy = malloc(sizeof(char*) * i + 1);
+
+  for (j = 0; j < i; j++) {
+      envCopy[j] = malloc(strlen(env[j]));
+      strcpy(envCopy[j], env[j]);
```
rouault, line 253:
```suggestion
      strcpy(envCopy[i], env[i]);
```

mapserv.c
```diff
@@ -239,12 +238,33 @@ int main(int argc, char *argv[])
...
+
 #ifdef USE_FASTCGI
   msIO_installFastCGIRedirect();
 
   /* In FastCGI case we loop accepting multiple requests.  In normal CGI */
   /* use we only accept and process one request.  */
   while( FCGI_Accept() >= 0 ) {
+
+      // copy across all environ variables to the request environment
+      for (j = 0; j < i; j++) {
```
rouault, line 264:
```suggestion
      for (int i = 0; i < num_env_var; i++) {
```

mapserv.c
```diff
@@ -239,12 +238,33 @@ int main(int argc, char *argv[])
...
+          putenv(envCopy[j]);
```
rouault, line 265:
```suggestion
          putenv(envCopy[i]);
```

rouault (commented):
According to https://www.mkssoftware.com/docs/man5/environ.5.asp,
you should do
```
#include <unistd.h>
extern char **environ; 
```
Windows has likely not unistd.h however. According to https://docs.microsoft.com/en-us/cpp/c-runtime-library/environ-wenviron?view=msvc-140, you may need to use _environ

geographika:
Thanks @rouault for the code review! I've added in the suggestions. I'm fairly sure some variables/memory needs to be freed at some stage but not sure where..

I'm still not sure if it is a good idea to go with this approach. Looking at the FastCGI docs, as I understand it, `getenv` *shouldn't* work on other web servers once in that loop. The FastCGI code uses `fcgi - 2.4.2` from [GISInternals](https://www.gisinternals.com/packageinfo.php?file=release-1928-x64-gdal-mapserver.zip), which is the latest version. Do Linux builds use the same?

Overwriting the env with all variables seems dangerous, but to check every key to see if it exists seems laborious. Messing with `PATH` etc. could cause further issues. 

Creating a config object outside the FastCGI loop may be better?

rouault:
It looks like the approach of modifying the environment in combination with FastCGI is a dead-end
The doc of FCGI_Accept() mentions:
```
 * Side effects:
 *      If the application was invoked as a FastCGI server,
 *      and this is not the first call to this procedure,
 *      FCGI_Accept first performs the equivalent of FCGI_Finish.
 *
 *      On every call, FCGI_Accept accepts the new request and
 *      reads the FCGI_PARAMS stream into an environment array,
 *      i.e. a NULL-terminated array of strings of the form
 *      ``name=value''.  It assigns a pointer to this array
 *      to the global variable environ, used by the standard
 *      library function getenv.  It creates new FCGI_FILE *s
 *      representing input from the HTTP server, output to the HTTP
 *      server, and error output to the HTTP server, and assigns these
 *      new files to stdin, stdout, and stderr respectively.
 *
 *      DO NOT mutate or retain pointers to environ or any values
 *      contained in it (e.g. to the result of calling getenv(3)),
 *      since these are freed by the next call to FCGI_Finish or
 *      FCGI_Accept.  In particular do not use setenv(3) or putenv(3)
 *      in conjunction with FCGI_Accept.
```

geographika:
@rouault - you added some good information in https://github.com/MapServer/MapServer/issues/6167#issuecomment-706116174 as part of #6167 - maybe `getenv` does cause issues in other environs/web servers?

rouault:
ok, so for Apache / FCGI, after experimentations, I have now a good idea of what happens:
- if you add in your apache conf ``SetEnv FOO my_value`` (or ``PassEnv FOO`` and FOO is for example set in /etc/apache2/envvars), then getenv("FOO") only works *after* FCGI_Accept, and is added after the per-request CGI environment variables
- if you add ``FcgidInitialEnv FOO my_value``, then getenv("FOO") only works *before* FCGI_Accept. This is what you'd typically need for MS_DEBUGLEVEL which is read before the fcgi loop. In addition to what you explicitly pass that way, apache also seems to propagate automatically PATH and LD_LIBRARY_PATH

sdlime:
> ok, so for Apache / FCGI, after experimentations, I have now a good idea of what happens:
> 
> * if you add in your apache conf `SetEnv FOO my_value` (or `PassEnv FOO` and FOO is for example set in /etc/apache2/envvars), then getenv("FOO") only works _after_ FCGI_Accept, and is added after the per-request CGI environment variables
> * if you add `FcgidInitialEnv FOO my_value`, then getenv("FOO") only works _before_ FCGI_Accept. This is what you'd typically need for MS_DEBUGLEVEL which is read before the fcgi loop. In addition to what you explicitly pass that way, apache also seems to propagate automatically PATH and LD_LIBRARY_PATH

So I interpret this as Apache CGI and FastCGI can access the environment vars as intended, correct?

rouault:
> So I interpret this as Apache CGI and FastCGI can access the environment vars as intended, correct?

yes, but you have to be careful and rather knowledgeable of mapserver internals to know if you need to define a per-request var or an initial setting one.

geographika:
> 
> 
> ok, so for Apache / FCGI, after experimentations, I have now a good idea of what happens:
> 
>     * if you add in your apache conf `SetEnv FOO my_value` (or `PassEnv FOO` and FOO is for example set in /etc/apache2/envvars), then getenv("FOO") only works _after_ FCGI_Accept, and is added after the per-request CGI environment variables
> 
>     * if you add `FcgidInitialEnv FOO my_value`, then getenv("FOO") only works _before_ FCGI_Accept. This is what you'd typically need for MS_DEBUGLEVEL which is read before the fcgi loop. In addition to what you explicitly pass that way, apache also seems to propagate automatically PATH and LD_LIBRARY_PATH

Thanks for that clarification. It looks like IIS FastCGI only supports the equivalent of `FcgidInitialEnv FOO my_value` with no possibility I can find for setting anything after FCGI_Accept. 

I think this approach is a dead-end in terms of trying to fix IIS and a config approach will be the cleaner/more robust option for setting global MapServer parameters, so closing. Thanks for the code suggestions though @rouault - my C has improved, as has my appreciation for how much easier things are in Python..

rouault:
> with no possibility I can find for setting anything after FCGI_Accept.

so MS_MAP_PATTERN and the like can't be used with IIS in FastCGI, right ?

> my C has improved

that's what I'm slowly converting what I can to C++. std::vector<std::string> is less unfriendly than raw char** :-)

geographika:
> so MS_MAP_PATTERN and the like can't be used with IIS in FastCGI, right ?

I've not been able to, and I've gone through pretty much any IIS/FastCGI doc I've been able to find.
It is however pretty trivial to filter querystring parameters to have the same result. 
The one environ setting that doesn't have a workaround is `CURL_CA_BUNDLE`. 

Is it worth adding this as a `CONFIG` option to a Mapfile and checking with `msGetConfigOption` as well as the environ, or waiting til 8.0?

sdlime:
Could we store MS_MAP_PATTERN, etc... using CPLSetConfigOption() before FCGI_accept() and use that as a simple store in the meantime? Then use CPLGetConfigOption() in place of (or in addition to) getenv(). So no putenv() in the mix.

rouault:
> Could we store MS_MAP_PATTERN, etc... using CPLSetConfigOption() before FCGI_accept() and use that as a simple store in the meantime? Then use CPLGetConfigOption() in place of (or in addition to) getenv(). So no putenv() in the mix.

That should work. At least for anything under MapServer and GDAL codebases (but not for PROJ's own env variables typically)

sdlime:
What GDAL versions so CPLSetConfigOption()/CPLGetConfigOption() apply to, has it been around awhile?

rouault:
> What GDAL versions so CPLSetConfigOption()/CPLGetConfigOption() apply to, has it been around awhile?

Not sure I was born yet when it was there :-) Anyway Mapserver already uses them : https://github.com/MapServer/MapServer/search?q=CPLSetConfigOption

sdlime:
Logical to take advantage then.

sdlime:
Perhaps search for getenv() and use that as a guide for what to grab - the MS_* stuff and things like SERVER_NAME, SERVER_PORT, etc... Then switch to CPLGetConfigOption() as much as possible. I'm assuming the calls in cgiutil.c would remain as would some other areas. Might be ok for both CGI and FCGI.

PR #6291, opened 2021-04-12
Add action to run Coverity static scan.

sdlime:
Initial version - we're allowed 2 scans per day at most so at the moment this needs to be triggered by hand. Eventually (post-8.0 release) this can be scheduled action. Requires a project token set as a secret. --Steve


rouault:
> we're allowed 2 scans per day at most so at the moment this needs to be triggered by hand. Eventually (post-8.0 release) this can be scheduled action

why wait 8.0 ?

You can add the secret at https://github.com/MapServer/MapServer/settings/secrets/actions/new

.github/workflows/coverity-scan.yml
```diff
@@ -0,0 +1,63 @@
...
+  #  - cron: '0 18 * * *' # Daily at 18:00 UTC
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
```
rouault, line 16: to avoid this running (and failing) on clones:
```suggestion
    if: github.repository == 'MapServer/MapServer'
    steps:
```

sdlime:
> > we're allowed 2 scans per day at most so at the moment this needs to be triggered by hand. Eventually (post-8.0 release) this can be scheduled action
> 
> why wait 8.0 ?
> 
> You can add the secret at https://github.com/MapServer/MapServer/settings/secrets/actions/new

I was thinking the initial pass through defects would be part of 8.0 work and would like to save those runs for adhoc use.

.github/workflows/coverity-scan.yml
sdlime: Wouldn't we want clones - those with the token - be able to run this?

rouault:
> I was thinking the initial pass through defects would be part of 8.0 work and would like to save those runs for adhoc use.

ok, then perhaps schedule once a week on sundays for now ?

.github/workflows/coverity-scan.yml
rouault: ah, perhaps there's a way to write a condition on the availability of the ${{ secrets.COVERITY_SCAN_TOKEN }} . Like done in https://github.com/entropia/tip-toi-reveng/commit/2582843896dbbb34581edf2335e3a658d6168747

sdlime:
> > I was thinking the initial pass through defects would be part of 8.0 work and would like to save those runs for adhoc use.
> 
> ok, then perhaps schedule once a week on sundays for now ?

That would be fine w/me. We'd want to make sure the schedule would only apply to MapServer/MapServer an not clones. Guessing we can add:

  `if: github.repository == 'MapServer/MapServer' `

before the on: section. I'll update with those changes... Thanks for the review!

.github/workflows/coverity-scan.yml
sdlime: will take a look...

sdlime:
Seems like specific secrets can't be used in conditionals: https://github.com/actions/runner/issues/520 

Perhaps the best course of action would be limit to MapServer/MapServer as you suggested and then forks could update a local copy of the action in their feature branch if necessary.

rouault:
Looks good to merge to me

sdlime:
Thanks @rouault!

PR #6294, opened 2021-04-14
CI: test mapserv as a CGI and FastCGI binary

rouault:


sdlime:
Excellent addition...

rouault:
yeah, was a bit of a shame that for a server oriented software we didn't had yet testing of the whole client -> webserver -> cgi binary :-)

PR #6295, opened 2021-04-14
Fix resource leak and pointObj initialization errors.

sdlime:
Identified by Coverity scanning... --Steve


jmckenna:
wow great work!

jmckenna:
@sdlime can we try to backport these?

sdlime:
> @sdlime can we try to backport these to branch-7-6 ? (will add the label, until you say nay)

I think I'd hang on for some review first. These were pretty trivial (more testing a process than anything) but better to be safe.

rouault (approved):

PR #6300, opened 2021-04-16
Require url-based symbol values to be pre-defined.

sdlime:


sdlime:
I think this should be back ported...

PR #6303, opened 2021-04-17
WIP: Support for a mapserv config file.

sdlime:
Work in progress to support (and require) a MapServer config file for use by the CGI. Draft RFC is underway at https://github.com/sdlime/mapserver/wiki/MapServer-8.0-Config-File.


sdlime:
@rouault, @geographika - for this pull request I need to update the test suite to have it's own config file for mapserv-based tests. I added that file to the repo but need to reference it properly using an environment variable called MAPSERVER_CONFIG. What's the right way to do that in the context of the test configuration? Value would be something like {MSAUTOTEST DIRECTORY}/etc/mapserv.conf. --Steve

rouault:
> What's the right way to do that in the context of the test configuration?

in msautotest/pytest.ini, you can add environment variables like in https://github.com/MobileDynasty/pytest-env#usage . However as we change the current working directory in each subdirectory, I don't think this can be used, or maybe counter-intutively, by setting MAPSERVER_CONFIG to "../etc/mapserv.conf" .  If that doesn't work, probably hacking into msautotest/conftest.py

sdlime:
Looks like all of the ows POST-based tests are failing for Travis, weird. AppVeyor is puking on lexer globals in a .cpp, not sure how to fix that one.

rouault:
> AppVeyor is puking on lexer globals in a .cpp, not sure how to fix that one.

Maybe put their declaration inside a extern "C" { } block

sdlime:
> > AppVeyor is puking on lexer globals in a .cpp, not sure how to fix that one.
> 
> Maybe put their declaration inside a extern "C" { } block

That was the ticket, thanks. Will take a look at the build action...

sdlime:
@rouault, one of the Travis builds failed with memory leaks detected related to gdal. Does additional cleanup need to be done when using CPLSetConfigOption()? I thought perhaps not since it was already being used in the codebase.

rouault:
> Does additional cleanup need to be done when using CPLSetConfigOption()?

hum, normally not. Not sure what is going on there

msautotest/pytest.ini
```diff
@@ -2,3 +2,5 @@
 python_files = *.py
 testpaths = api misc gdal query renderers wxs sld mspython
 norecursedirs = php mssql
+env =
+    MAPSERVER_CONFIG_FILE=../etc/mapserv.conf
```
geographika, line 6: This file doesn't exist? Should it be `mapserver-sample.conf `?

geographika:
@sdlime - would it be an option to allow a config filename to be passed via the command line for testing?
Looking at the `mapserv.c` code there are warnings about security issues linking to #3485 - would this also apply to adding a new config filename switch?

sdlime:
> @sdlime - would it be an option to allow a config filename to be passed via the command line for testing?
> Looking at the `mapserv.c` code there are warnings about security issues linking to #3485 - would this also apply to adding a new config filename switch?

I don't remember the details of that issue but presumably we'd still want to avoid any additional command arguments. I only added "PATH_INFO=..." as part of OGCAPI support since it parallels the "QUERY_STRING=..." hook. @dmorissette might have some other ideas. Seems safest to add a mechanism to set/override environment vars to the regression test suite.

msautotest/pytest.ini
sdlime: @rouault mentioned this configuration was not exactly intuitive but the file referenced it's actually msautotest/etc/mapserv.conf. That config file is just the bare minimum and only suitable for use with msautotest. The etc/mapserver-sample.conf is different. I included that as a kitchen sink base file for users to start from with all the options and heavily commented. Kinda like you see in an Apache configuration file. Perhaps that sample file could be stored someplace else like the new share directory added with the ogcapi work.

--Steve

sdlime:
@rouault, what do you think the best way to add a config file to the build workflow is?

rouault:
> what do you think the best way to add a config file to the build workflow is?

You mean something that doesn't involve the msautotest/pytest.ini way ?
There's still the possibility to add manual steps towards the end of .github/workflows/start.sh

sdlime:
I figure there must be a way to reference the mapserver.conf config file we have as part of msautotest. We could add MAPSERVER_CONFIG_FILE environment variable declaration to the snippet of apache configuration in start.sh. I'm not sure what the path would be though.

sdlime:
The Build task gets further now and a config file is recognized. CGI query runs but FastCGI crashes.  @rouault, do you know of a way to use a debugger with fastcgi? Doesn't seem possible.

sdlime:
I'm running into a couple of issues with fcgi. I'm testing this using vagrant and a setup similar to what @rouault developed as part of the Build task. I'm testing with bare calls to mapserv.cgi and mapserv.fcgi. They should load the config file and return the standard "... QUERY_STRING is set, but empty." error.

1. When run as a fcgi the MAPSERVER_CONFIG_FILE environment variable isn't being picked up for some reason - always null. Works as expected with cgi.
2. As a result, when msLoadConfig() runs it fails but the error message doesn't get returned to the client and instead ends up in the main apache error log and a 500 error is thrown. Must be an IO issue. I tried moving the msIO_installFastCGIRedirect() function ahead of the configuration loading but then the error message gets completely lost and I still end up with a 500 error.

Any ideas welcome...

--Steve

mapscript/swiginc/config.i
```diff
@@ -0,0 +1,43 @@
...
+   OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+   THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+   FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+   DEALINGS IN THE SOFTWARE.
+   ===========================================================================
+*/
+
+%extend configObj
```
sdlime, line 30: @geographika, I just wanted to make sure you agree with the changes I made here:

1. Default for filename is NULL instead of "" so the environment is checked if necessary.
2. I added a deconstructor since the configObj is not freed when the mapObj is freed (and it doesn't even need to be passed to the mapObj constructor.

msautotest/pytest.ini
geographika: Oops..! It is right above this comment in the GitHub web UI. Thanks for the clarification.

mapscript/swiginc/config.i
geographika: Sounds good. I used an empty string as that was what was in map.i - `mapObj(char *filename="")`

rouault:
This PR would need to be rebased on top of latest master or have master merged into it to have the github action tests run (they have been added after the branch was created)

sdlime:
Yup, last merge with main was fairly recent . Working on fixing the build workflow now - need to create a config file to test against.

sdlime:
Anyone have an idea about the Travis/PHP 7.2 failure? Doesn't seem related to the pull request but I see other pull requests are passing.

rouault:
> Anyone have an idea about the Travis/PHP 7.2 failure? Doesn't seem related to the pull request but I see other pull requests are passing.

it could be caused by a call to CPLGetConfigOption() after GDAL cleanup (based on the "CPLGetTLSList(): pthread_setspecific() failed!" message), which causes new allocations

sdlime:
>> it could be caused by a call to CPLGetConfigOption() after GDAL cleanup (based on the "CPLGetTLSList(): pthread_setspecific() failed!" message), which causes new allocations

I don't see where but I'll keep looking - CPLGetTLSList() isn't present in the source.

rouault:
> CPLGetTLSList() isn't present in the source.

Sure, but it is indirectly called by CPLGetConfigOption()

sdlime:
@rouault, how should PROJ_LIB be handled in the context of a config file? We could handle it like a plugin I guess. That is, if a config obj is present then it can only be set that way.

rouault:
> how should PROJ_LIB be handled in the context of a config file?

I see msSetPROJ_LIB() has subtelties where the PROJ_LIB path can be relative to the mapfile

mapproject.c
```diff
@@ -2341,7 +2341,7 @@ void msProjLibInitFromEnv()
 {
   const char *val;
 
-  if( (val=getenv( "PROJ_LIB" )) != NULL ) {
+  if( (val=CPLGetConfigOption( "PROJ_LIB", NULL )) != NULL ) {
```
rouault, line 2344: you need a ``#include <cpl_conv.h>``

rouault:
It seems most of the tests are skipped

sdlime:
> It seems most of the tests are skipped

So weird, all sorts of _missing some or required components, skip_ messages.

rouault:
> So weird, all sorts of _missing some or required components, skip_ messages.

probably that shp2img no longer outputs the requirements of the tests

sdlime:
It's the only thing to have changed but the skipped tests aren't all shp2img-based. For example, query/query.map is skipped and it's all mapserv-based. All those tests need is INPUT=SHAPEFILE.

jmckenna:
@sdlime the RFC draft currently states "At present, a single environment variable (`MS_CONFIG_FILE`) is referenced to determine what config file to load"   but that does not work (the filename specified at compile-time always is searched no matter what I specify through MS_CONFIG_FILE).  I solved this by instead using the environment variable `MAPSERVER_CONFIG_FILE` in Apache etc which then overrides the filename that was hardcoded at compile time.  Will update draft now.

sdlime:
Good catch - the latter is correct (mirrors mapcache from a naming convention).

geographika:
@sdlime - some feedback from testing the branch on a Windows machine.

- No problems compiling the branch 👍 
- When running `mapserv -v` from the command line an error message is included in the output. This could be confusing/of concern to new users? Not sure of a better message - or maybe surpress it if the `-v` flag is used?

```
C:\MapServer>mapserv -v
Content-Type: text/html

<HTML>
<!-- MapServer version 7.7-dev OUTPUT=PNG OUTPUT=JPEG OUTPUT=KML SUPPORTS=PROJ SUPPORTS=AGG SUPPORTS=FREETYPE SUPPORTS=CAIRO SUPPORTS=SVG_SYMBOLS SUPPORTS=SVGCAIRO SUPPORTS=ICONV SUPPORTS=FRIBIDI SUPPORTS=WMS_SERVER SUPPORTS=WMS_CLIENT SUPPORTS=WFS_SERVER SUPPORTS=WFS_CLIENT SUPPORTS=WCS_SERVER SUPPORTS=SOS_SERVER SUPPORTS=OGCAPI_SERVER SUPPORTS=FASTCGI SUPPORTS=THREADS SUPPORTS=GEOS SUPPORTS=PBF INPUT=JPEG INPUT=OGR INPUT=GDAL INPUT=SHAPEFILE -->
<BODY BGCOLOR="#FFFFFF">
msLoadConfig(): Unable to access file. (/mapserver.conf)
</BODY></HTML>
```

The error can be removed by setting an environment variable pointing to a valid config file: `SET MAPSERVER_CONFIG_FILE=C:\MapServer\bin\mapserver.conf`

- When running through IIS and no environment variable is set then HTTP 500 errors are returned with no indication of what is wrong. No IIS logs are generated, or MapServer logs. I'm not sure if this is just the case with IIS? An HTTP 200 response with the same details (and `msLoadConfig(): Unable to access file. (/mapserver.conf)` message) would be very helpful for users trying to setup MapServer. 
- When setting the environment variable all works fine in my setup. 

image

## Questions

1. Will the following minimal conf file always allow MapServer users with no security requirements (e.g. setting up on a local machine for testing) always work?

```
CONFIG
  ENV
    MS_MAP_PATTERN "."
  END
END
```

2. (Just a suggestion that can be ignored!) If no config file is found, could the above config be hardcoded so MapServer would run out-of-the-box with no config file/environment variable setting? This might go completely against the rationale of the RFC, but I can imagine lots of new users struggling with setting environment variables on web servers, and if it is just to point to the above config then it may overcomplicate things. 
3. The RFC mentions the `PLUGINS` section - I don't believe this is yet implemented as part of the pull request. Is this still planned for the v8 release?

jmckenna:
For the record, from my earlier testing on Windows (to generate the RFC and document it), the response from `mapserv -v` and its error message returned was critical for me as a user, to see that the `ms4w.conf` file was not found - hiding that message would likely prevent myself and other users noticing that the file was not found by MapServer.

sdlime:
> @sdlime - some feedback from testing the branch on a Windows machine.

> * When running `mapserv -v` from the command line an error message is included in the output. This could be confusing/of concern to new users? Not sure of a better message - or maybe surpress it if the `-v` flag is used?

Agreed... I'll add a new comment with a task list.

> * When running through IIS and no environment variable is set then HTTP 500 errors are returned with no indication of what is wrong. No IIS logs are generated, or MapServer logs. I'm not sure if this is just the case with IIS? An HTTP 200 response with the same details (and `msLoadConfig(): Unable to access file. (/mapserver.conf)` message) would be very helpful for users trying to setup MapServer.

I'll test on MacOS as well. I don't recall offhand.

> ## Questions
> 1. Will the following minimal conf file always allow MapServer users with no security requirements (e.g. setting up on a local machine for testing) always work?
> 
> ```
> CONFIG
>   ENV
>     MS_MAP_PATTERN "."
>   END
> END
> ```

Yes. That's what's used as part of msautotest.

> 1. (Just a suggestion that can be ignored!) If no config file is found, could the above config be hardcoded so MapServer would run out-of-the-box with no config file/environment variable setting? This might go completely against the rationale of the RFC, but I can imagine lots of new users struggling with setting environment variables on web servers, and if it is just to point to the above config then it may overcomplicate things.

Personally I think we should be forcing proper configuration and would be -1 on a default.

> 2. The RFC mentions the `PLUGINS` section - I don't believe this is yet implemented as part of the pull request. Is this still planned for the v8 release?

Still needed. I'm unable to test though so if I could find a guinea pig to assist... (wink, wink)

jmckenna:
Most users only execute a 'mapserv -v' to check installation.  It is a critical command, and it is nice that MapServer now also reports it here when the config file is not found.

jmckenna:
Maybe it is all of the html text returned that is scarey, at commandline?  Sure, I can see that, some formatting improvements, but for me (as a user trying this, likely the first ha) that error message from 'mapserv -v' of '...*.conf not found* was exactly what I needed to see.      Put another way: users do not issue 'mapserv' at commandline without the "-v" (I mean users, not power users).  So this is why the config file should be searched for and error displayed for 'mapserv -v'.

jratike80:
For being even more useful could the message be enhanced then
"MS_CONFIG_FILE: [path to file, or the default location] : Found/Not found"

geographika:
@sdlime - happy to test the plugins config (it would be on Windows with the MSSQL driver). Would probably also need someone to test on Linux with the Oracle driver?

With regards to running `mapserv -v` at the command line, even if `MAPSERVER_CONFIG_FILE` is set as a Windows environment variable, the value would be ignored when running under the web server - at least on IIS/FastCGI. The `-v` would then always return the error or users would need to set the same variable twice. 

When the `MAPSERVER_CONFIG_FILE` is not found (e.g. the path is wrong) the web server simply exits with a generic HTTP 500 error, with no logs or further details. I tried adding various things to the code block below (`fflush`, `goto end_request;`, changing the error message to a simple string) but no luck. Is this the same on other servers? Maybe under FastCGI the exit needs to be handled in the FAST_CGI loop?

```
  config = msLoadConfig(NULL); // is the right spot to do this?
  if(config == NULL) {
    msCGIWriteError(mapserv);
    exit(0);
  }
```

geographika:
@sdlime - do you think this branch is ready to merge with master, and address any remaining issues and plugin functionality later? It would make it easier to test the OGC Features once the config file is available for setting map environment variables.

sdlime:
> @sdlime - do you think this branch is ready to merge with master, and address any remaining issues and plugin functionality later? It would make it easier to test the OGC Features once the config file is available for setting map environment variables.

Pretty close, really do need the plugin bit. You can use my branch that the pull request is based on for testing since it is up-to-date with main and does have the OGC feature support integrated.

sdlime:
Regarding the -v command-line option. Personally I think we need to move handling that option ahead of running msLoadConfig() and msSetup(). So 'mapserv -v' would always kick out the version number and the supported components and would be consistent across binaries (e.g. mapserv, shp2img/map2img, etc...). A bare 'mapserv' is where you'd see error messages of various types start to manifest themselves. --Steve

sdlime:
> @sdlime - happy to test the plugins config (it would be on Windows with the MSSQL driver). Would probably also need someone to test on Linux with the Oracle driver?

@geographika, I just added code to treat the value with

```
LAYER
  ...
  PLUGIN "value"
  ...
END
```

as a key to the config->plugins hash. If the config file is set (it has to be with any CGI/FCGI request) then the value must resolve or a parsing error is thrown. I tested generically and it behaves as expected. I'm not setup to configure a real plugin layer though. 

--Steve

geographika:
@sdlime - I'll give this a test with the MSSQL plugin, and update Appveyor builds to use a config file. Can we rebase this branch back onto main?

geographika:
@sdlime - good news is that the PLUGINS works fine with Mapfiles and MapServer.

A few notes for future docs:

- the error below is shown if the config is missing (might be worth adding the key name to this?):

```
loadCustomLayerDLL: General error message. Failed to load dynamic Layer LIB: (null)
msGetSymbol(): General error message. Can&#39;t load requested dynamic library: (null)
loadLayer(): General error message. Plugin value not found in config file.
```
- they error will occur for any layer without a plugin (even if they layer is not requested) as the plugins are checked when loading the Mapfile

The bad news (unless I'm missing something) - `msLoadConfig` is only called in mapserv.exe. Some of the msautotests are using `shp2img` - these never load the config and so can never load the plugin. This is also true for anyone wanting to use shp2img to test a Mapfile. Should `msLoadConfig` be added to shp2img along with a new command line parameter?

geographika:
@sdlime - see https://github.com/sdlime/mapserver/pull/9 for suggested fix for tests and loading config in `shp2img`.

sdlime:
> * the error below is shown if the config is missing (might be worth adding the key name to this?):
> 
> ```
> loadCustomLayerDLL: General error message. Failed to load dynamic Layer LIB: (null)
> msGetSymbol(): General error message. Can&#39;t load requested dynamic library: (null)
> loadLayer(): General error message. Plugin value not found in config file.
> ```

If the PLUGIN line is before CONNECTIONTYPE and CONNECTION you only get the loadLayer() error. I'm not quite sure how the execution path even gets to loadCustomLayerDLL to be honest. I can't remember, is the last error in the list the last one encountered?

> * they error will occur for any layer without a plugin (even if they layer is not requested) as the plugins are checked when loading the Mapfile

I only see the loadCustomLayerDLL error when a layer has CONNECTIONTYPE PLUGIN.

> The bad news (unless I'm missing something) - `msLoadConfig` is only called in mapserv.exe. Some of the msautotests are using `shp2img` - these never load the config and so can never load the plugin. This is also true for anyone wanting to use shp2img to test a Mapfile. Should `msLoadConfig` be added to shp2img along with a new command line parameter?

You're correct, good catch. I think your pull request is probably sufficient for the moment. Either a user will need to set an environment variable before running shp2img or they'll be relying on the value set at compile-time. We may need to add an argument.

sdlime:
@geographika, I did some testing related to the error messages and output is last in, first out. So the sequence:

```
loadCustomLayerDLL: General error message. Failed to load dynamic Layer LIB: (null)
msGetSymbol(): General error message. Can&#39;t load requested dynamic library: (null)
loadLayer(): General error message. Plugin value not found in config file.
```

indicates that some code related to plugins is being executed **after** the plugin value not found error. Note that I only see this sequence or errors with a layer definition like:

```
CONNECTIONTYPE PLUGIN
CONNECTION "..."
PLUGIN "myplugin"
```
If I move the PLUGIN statement before the CONNECTIONTYPE **or** change CONNECTIONTYPE to something other than PLUGIN then you don't see the loadCustomLayerDLL() error. There's still some intervention necessary.

--Steve

geographika:
@sdlime - on debugging the issue is the following function appears to return `NULL`

https://github.com/MapServer/MapServer/blob/1c14821bc4c6fc10e966fd05230ad71c05817c7d/mappluginlayer.c#L74-L75

But it actually returns an "empty" `VTFactoryItemObj` so the following check is true:

https://github.com/MapServer/MapServer/blob/1c14821bc4c6fc10e966fd05230ad71c05817c7d/mappluginlayer.c#L192

I added in a simple NULL check to the function - see https://github.com/sdlime/mapserver/pull/10
Is this good enough?

Any idea why the Travis build is failing with memory errors?

geographika:
@sdlime - rebasing onto main with https://github.com/MapServer/MapServer/pull/6420 should hopefully fix (all?) the Travis errors.

mapfile.c
```diff
@@ -4273,10 +4275,22 @@ int loadLayer(layerObj *layer, mapObj *map)
...
-                                      map);
+        if(map->config) { // value *must* represent a config key
+          char *value = NULL;
+          const char *plugin_library = NULL;
+
+	  if(getString(&value) == MS_FAILURE) return(-1);
+          plugin_library = msConfigGetPlugin(map->config, value);
+	  msFree(value);
+          if(!plugin_library) {
+            msSetError(MS_MISCERR, "Plugin value not found in config file. See mapserver.org/config_file.html for more information." , "loadLayer()");
```
geographika, line 4286: In https://github.com/MapServer/MapServer-documentation/pull/613 the URL would be mapserver.org/mapfile/config.html (seemed to fit alongside other Mapfile keyword docs - happy to change though)

sdlime:
It's just a placeholder, we can definitely update the URL in the messages.

mapfile.c
sdlime: The URL is just a placeholder so I'm not set on it and we can certainly change it. However, the config file is distinct from a mapfile so perhaps they should live at the same level? That is, perhaps /mapfile and /config_file (or /configfile) as sub-directories, so the URL referenced in code could be mapserver.org/config_file/ which would allow for multiple documents to be stored in that section.

geographika: Although distinct I couldn't think of much else to write, or split into more than one page: https://github.com/MapServer/MapServer-documentation/blob/090a233ba2f81156b74860470e8a441414b833c1/en/mapfile/config.txt
Could maybe be moved to the /installation/ folder?

sdlime:
I could see moving a couple of pages into that section, for example:

  https://mapserver.org/optimization/environment_variables.html
  https://mapserver.org/optimization/limit_mapfile_access.html

Those shouldn't be under optimization when they're core concepts.

--Steve

mapfile.c
sdlime: I could see moving a couple of pages into that section, for example:

  https://mapserver.org/optimization/environment_variables.html
  https://mapserver.org/optimization/limit_mapfile_access.html

Those shouldn't be under optimization when they're core concepts.

--Steve

sdlime:
> @sdlime - rebasing onto main with #6420 should hopefully fix (all?) the Travis errors.

Looks like we need to do more to clean up the config in shp2img, still detecting leaks.

PR #6304, opened 2021-04-18
WIP: Use CPLSetConfigOption/CPLGetConfigOption for some CGI/FastCGI-related env vars.

sdlime:
Potential solution for IIS/FastCGI where we push some high-value environment variables into the CPL environment (using CPLSetConfigOption) and then reference them there (CPLGetConfigOption). This is against **branch-7-6** rather than main since the config file work would replace this.

Should be ok for non-IIS/FastCGI as well.

--Steve

### DIFF
--- a/maphttp.c
+++ b/maphttp.c
@@ -39,7 +39,7 @@
 #include "mapthread.h"
 #include "mapows.h"
 
-
+#include "cpl_conv.h"
 
 #include <time.h>
 #ifndef _WIN32
@@ -471,7 +471,7 @@ int msHTTPExecuteRequests(httpRequestObj *pasReqInfo, int numRequests,
    * If set then the value is the full path to the ca-bundle.crt file
    * e.g. CURL_CA_BUNDLE=/usr/local/share/curl/curl-ca-bundle.crt
    */
-  pszCurlCABundle = getenv("CURL_CA_BUNDLE");
+  pszCurlCABundle = CPLGetConfigOption("CURL_CA_BUNDLE", getenv("CURL_CA_BUNDLE")); // fall back to env
 
   if (debug) {
     msDebug("HTTP: Starting to prepare HTTP requests.\n");
--- a/maphttp.c
+++ b/maphttp.c
@@ -471,7 +471,7 @@ int msHTTPExecuteRequests(httpRequestObj *pasReqInfo, int numRequests,
    * If set then the value is the full path to the ca-bundle.crt file
    * e.g. CURL_CA_BUNDLE=/usr/local/share/curl/curl-ca-bundle.crt
    */
-  pszCurlCABundle = CPLGetConfigOption("CURL_CA_BUNDLE", getenv("CURL_CA_BUNDLE")); // fall back to env
+  pszCurlCABundle = CPLGetConfigOption("CURL_CA_BUNDLE", NULL);
 
   if (debug) {
     msDebug("HTTP: Starting to prepare HTTP requests.\n");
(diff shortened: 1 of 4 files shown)


maphttp.c
```diff
@@ -471,7 +471,7 @@ int msHTTPExecuteRequests(httpRequestObj *pasReqInfo, int numRequests,
    * If set then the value is the full path to the ca-bundle.crt file
    * e.g. CURL_CA_BUNDLE=/usr/local/share/curl/curl-ca-bundle.crt
    */
-  pszCurlCABundle = getenv("CURL_CA_BUNDLE");
+  pszCurlCABundle = CPLGetConfigOption("CURL_CA_BUNDLE", getenv("CURL_CA_BUNDLE")); // fall back to env
```
rouault, line 474:
```suggestion
  pszCurlCABundle = CPLGetConfigOption("CURL_CA_BUNDLE", NULL);
```
CPLGetConfigOption() already fallbacks to getenv

mapserv.c
```diff
@@ -162,6 +162,17 @@ int main(int argc, char *argv[])
   if(msGetGlobalDebugLevel() >= MS_DEBUGLEVEL_TUNING)
     msGettimeofday(&execstarttime, NULL);
 
+  /* push high-value ENV vars into the CPL global config - primarily for IIS/FastCGI */
+  const char *value = getenv("CURL_CA_BUNDLE");
```
rouault, line 166: Perhaps replace this by
```
const char* const apszEnvVars[] = { 
  "CURL_CA_BUNDLE", "MS_MAPFILE", "MS_MAP_NO_PATH", "MS_MAP_PATTERN",
  NULL /* guard */ };
for( int i = 0; apszEnvVars[i] != NULL; ++i ) {
   const char* value = getenv(apszEnvVars[i]);
   if(value) CPLSetConfigOption(apszEnvVars[i], value);
}
```

mapserv.h
```diff
@@ -41,6 +41,9 @@
 #include "maptile.h"
 
 #include "cgiutil.h"
+
+#include "cpl_conv.h"
```
rouault, line 45: Not really a good practice to increase the number of includes in a global header, when itself doesn't require it. Should be the responsibility of each .c to include the needed headers

```suggestion
```

jmckenna:
This will be a great help Steve, as MS4W has had to do workarounds for this issue for several years, so that users never had to deal with this problem (also affects MapCache at */mapcache/lib/http.c* ...will file that ticket now for MapCache).

maphttp.c
sdlime: Incorporated, thanks.

sdlime:
Will wait on merging until folks actually confirm it helps in various environments - mostly Windows CGI and FastCGI.

geographika:
@sdlime - many thanks for this! I've built and tested on a fresh Windows 10/ IIS 10 install. All environ variables are now read correctly, with cascading WMS over HTTP now possible thanks to being able to use the `CURL_CA_BUNDLE` variable. 

I've updated the IIS docs in preparation for this being merged: https://github.com/MapServer/MapServer-documentation/pull/517

geographika (approved):

sdlime:
Should I redo this against main? Seems like it and then we'd backport to wherever.

geographika:
> 
> 
> Should I redo this against main? Seems like it and then we'd backport to wherever. I've not done a rebase before though...

@sdlime - I cherry-picked these onto main in #6305 if that helps?

PR #6306, opened 2021-04-23
Fix most of remaining Coverity scan warnings with high priority

rouault:


sdlime:
+1 - Thanks @rouault!

jmckenna:
Great work @rouault !!

sdlime:
Should these (and other recent commits related to scanning) be pushed to older branches as well?

rouault:
> Should these (and other recent commits related to scanning) be pushed to older branches as well?

I'm afraid we can't realisticly pretend to be able to maintain secure more than one release branch at a time..

jmckenna:
My personal opinion (I wrote this in some response, but I can't remember if it was public): I'd like to see our efforts (which are unfunded of course, and limited) to sticking with supporting the most recent release branch (branch-7-6), I'm not for supporting older unmaintained branches (such as branch-7-4 etc).

PR #6311, opened 2021-04-24
Another batch of coverity scan fixes

rouault:
Should hopefully fix the false positive about the last remaining one in 'high' category and a few 'medium' ones. Most of them are false positives or without any real consequence (dead code, useless check etc), except the "agg_font_freetype.cpp: fix incorrect use of memcpy() (CID 1503528)" commit


sdlime:
Thanks Even, ok with me to apply... --Steve

jmckenna:
thanks again @rouault !

Issue #6292, opened 2021-04-13
Trailing Whitespace in MapServer EXPRESSION


glk-24:
## Expected behavior and actual behavior.

Trailing spaces are removed from the string attribute in the `EXPRESSION`.
So the feature isn't displayed in the returned image.
Leading spaces working fine.
The same behaviour occur with a GetFeatureInfo-Request. (leading spaces transfered while trailing spaces get trimmed).

Tested with a PostGIS layer.

e.g.

**db:**
feature | test
1 | "abc "
2 | " abc"

**mapserver mapfile:**
`EXPRESSION ('[test]' eq 'abc ')` - feature 1 not displayed
`EXPRESSION ('[test]' eq 'abc')` - feature 1 displayed
`EXPRESSION ('[test]' eq ' abc')` - feature 2 displayed
`EXPRESSION ('[test]' eq 'abc')` - feature 2 not displayed

## Operating system

CentOS Linux release 7.6.1810 (Core)

## MapServer version and installation method

7.6.2

sdlime:
@glk-24, out of curiosity does the same thing happen with a shapefile? --Steve

glk-24:
@sdlime Unfortunaly i can't create or edit a shapefile with trailing whitespace. Global Mapper, QGIS and even Open Office Calc trim leading and trailing whitespace on attributes when saving the dbf-file.

jratike80:
Isn't dbf fields padded with spaces till full width? Then trailing whitespace would not make sense really.

jmckenna:
Small test case, includes SpatialiteDB & mapfile & commands.txt, to reproduce: https://gatewaygeomatics.com/dl/ticket-6292-trailing-whitespace.zip  

Similar to @glk-24's report:
```
       EXPRESSION "Lake Superior " #map is generated
       EXPRESSION "Lake Superior"  #no map generated  

       *where the attribute value was modified to include a space in the database table, such as "Lake Superior "
```

jmckenna:
Ah, I made it too fast to realize now that the SpatialiteDB case seems to be working as expected...  will try now with a small PostGIS test...

jmckenna:
@glk-24 I've updated the small test case, to include a mapfile with 3 layers (PostGIS, SpatiaLite, Shapefile), and included a dump of the PostGIS 'lakes' database (5 lakes) that you can restore locally, and use with that mapfile. (the spatialite & shapefiles are also included)

[...]

PR #6312, opened 2021-04-28
WIP: Selectively trim blanks when populating shape items in PostGIS (#6292)

sdlime:
It doesn't look like that call is necessary and it can remove legitimate trailing blanks in the data. However, if there are data types that are padded w/blanks then this likely isn't the right solution. Will need some review/discussion.

--Steve


rouault:
Superseded per https://github.com/MapServer/MapServer/pull/6349

PR #6317, opened 2021-05-03
Removal of styling keywords from a Mapfile CLASS

geographika:
This pull request removes deprecated keywords from a Mapfile [CLASS](https://mapserver.org/mapfile/class.html) as part of [MS RFC 133: Mapfile Syntax Cleanup](https://mapserver.org/development/rfc/ms-rfc-133.html). 

The removed keywords provided a shortcut of placing style keywords directly on a `CLASS`. 
The following:

```
 CLASS COLOR 0 0 255 END
```

Must now be rewritten:

```
 CLASS STYLE COLOR 0 0 255 END END
```

The pull request removes the code to map these to styles, and also updates the many msautotest Mapfiles using this approach. 
**Note no changes have been made to the lexer as part of this pull request**. 

The aims of this update are:

- Removes ambiguity on which approach to take
- Avoid supporting deprecated Mapfile keywords indefinitely
- Allows a simpler syntax to be supported by various Mapfile generators
- May make it easier to move to [named styles](https://mapserver.org/mapfile/style.html#style) in the future


sdlime:
Aren't there a bunch of keywords (e.g. OVERLAYBACKGROUNDCOLOR) that can be removed from the lexer and mapfile.h, From the latter...

```
#define OVERLAYBACKGROUNDCOLOR 1106
#define OVERLAYCOLOR 1107
#define OVERLAYMAXSIZE 1108
#define OVERLAYMINSIZE 1109
#define OVERLAYOUTLINECOLOR 1110
#define OVERLAYSIZE 1111
#define OVERLAYSYMBOL 1112
```

geographika:
@sdlime - I was aware, but have so far managed to avoid the lexer/parser side of MapServer development..

As I understand it (and there is a good [2003 post](https://lists.osgeo.org/pipermail/mapserver-users/2003-December/047546.html) by yourself which gives a good overview):

**maplexer.l** is edited manually, and then the **flex** tool is run on the command line to generate maplexer.c which is then checked in. 

This all seems to be scripted in the [vagrant build script](https://github.com/MapServer/MapServer/blob/main/scripts/vagrant/mapserver.sh) - so I presume this is the easiest way to regenerate maplexer.c?

I'm going through the [MapServer vagrant docs](https://mapserver.org/development/tests/vagrant.html) at the moment to try and get this setup locally.

sdlime:
> This all seems to be scripted in the [vagrant build script](https://github.com/MapServer/MapServer/blob/main/scripts/vagrant/mapserver.sh) - so I presume this is the easiest way to regenerate maplexer.c?

Yup, that's what I end up doing to get a version to commit.

BTW There are a couple of other keywords to remove as part of RFC 133 - TEMPLATEPATTERN and DATAPATTERN. Both are deprecated and were replaced by more general validation. The FILEPATTERN symbol (mapfile.h) can also be recovered.

geographika:
I used the vagrant setup to regenerate maplexer.c from an updated [maplexer.l](https://github.com/MapServer/MapServer/pull/6317/files?file-filters%5B%5D=.l&file-filters%5B%5D=.map#diff-56fc276c6ca53fe706a8b4a607f54c6ee136316ed3e1175a5403d3fab8959f27) and added to this pull request. If I understand correctly the keywords aren't tied to particular classes in the lexer, so the other removed `CLASS` keywords, which can also be found in a `STYLE` object e.g. `COLOR`) should be left alone. 
I think the pull request is ready for a final review. 

With regards to `TEMPLATEPATTERN` and `DATAPATTERN` - there is no mention of deprecation on the [MAP](https://mapserver.org/mapfile/map.html) docs page. Do they need to be deprecated prior to removal? Or are they already redundant for the 8.0 release?
I can look at these in a separate pull request.

sdlime:
I can't imagine they are used much, if at all. They've been redundant for years and can be removed IMHO w/documentation.

geographika:
Unless there are any objections I'll merge in the next couple of days to move onto some other keywords (there are lots of msautotest changes so I want to avoid lots of merge conflicts).

Issue #6262, opened 2021-03-14
Querymap fails for polygon layers with labels only (no styles)


szekerest:
Currenly msDrawQueryLayer fails with the following error if the layer type is polygon, the querymap style is MS_HILITE, and the layer class contains only label elements.

> msDrawMap(): Image handling error. Failed to draw layer named '...'.;msDrawQueryLayer(): General error message. Don't know how to draw class Class (0) of layer ... without a style definition.

The same setting works with the point and line layers (where the labels are being highlighted) so I guess it should also work with the polygon layers.

The corresponding code looks like this:

```
 if(layer->type == MS_LAYER_POLYGON) { /* alter BOTTOM style since that's almost always the fill */
        if (layer->class[i]->styles == NULL) {
          msSetError(MS_MISCERR, "Don't know how to draw class %s of layer %s without a style definition.", "msDrawQueryLayer()", layer->class[i]->name, layer->name);
          msFree(colorbuffer);
          msFree(mindistancebuffer);
          return(MS_FAILURE);
        }
```

however the classes with no styles can surely be handled later in the code:

```
} else if (layer->class[i]->numlabels > 0) {
          colorbuffer[i] = layer->class[i]->labels[0]->color;
          layer->class[i]->labels[0]->color = map->querymap.color;
      }
```

jmckenna:
Fixed by @szekerest through https://github.com/MapServer/MapServer/pull/6318

PR #6318, opened 2021-05-03
Fix querymap failure for polygon layers with labels only (no styles) (#6262)

szekerest:


jmckenna:
thanks @szekerest !

jmckenna:
backported to branch-7-6 through https://github.com/MapServer/MapServer/commit/782ee8bacb7f322fd603383fc54040521f995b02

PR #6324, opened 2021-05-12
Try to fallback to using a question mark if a glyph is missing in a font

wicksell:
As it is now MapServer will fail to render if a layer is using characters that's not available in the font. This resolves this by trying to fallback to using a question mark in these situations.

I don't know if this will work as a general solution but it is very annoying that rendering will fail when the content contain characters not supported by the font.


geographika:
Looks a nice addition -thanks @wicksell 
Would you be able to create a test for this? https://github.com/MapServer/MapServer/blob/main/msautotest/renderers/font-fail.map could be used as the basis for this.

wicksell:
Sorry, but I will not have time to create a test for this in the near future.

rouault:
@wicksell I tried your patch but it had no effect on the test case I tried (included in the follow-up commit I've added). This commit patches check_single_font() instead, which works for me.

PR #6329, opened 2021-05-14
Remove MAP TRANSPARENT keyword

geographika:
This pull request removes [TRANSPARENT](https://mapserver.org/mapfile/map.html#mapfile-map-transparent) property of a `MAP` which has been deprecated since version 4.6 as part of  [MS RFC 133: Mapfile Syntax Cleanup](https://mapserver.org/development/rfc/ms-rfc-133.html). 

Transparency should now always be set in the `OUTPUTFORMAT`. So the following:

```
MAP
    IMAGETYPE png
    TRANSPARENT ON
...
```

Becomes:

```
MAP
    IMAGETYPE png
    OUTPUTFORMAT
      NAME "png"
      DRIVER AGG/PNG
      MIMETYPE "image/png"
      IMAGEMODE RGB
      EXTENSION "png"
      FORMATOPTION "GAMMA=0.75"
      TRANSPARENT ON
  END
...
```

The default OUTPUTFORMAT settings are shown at https://mapserver.org/mapfile/outputformat.html
None of these have `TRANSPARENT ON` by default, so a full `OUTPUTFORMAT` needs to be configured. 

If a `MAP` has `TRANSPARENT OFF` set, then this can simply be removed.


jratike80:
In WMS the optional parameter `TRANSPARENT=TRUE|FALSE` has FALSE as default. So in a compliant WMS server configuration the outputformat must not have TRANSPARENT ON by default,  and &TRANSPARENT=TRUE in GetMap should be honored if the server and the format support transparency.

There is nothing about TRANSPARENT in https://www.mapserver.org/ogc/wms_server.html. Do we support it at all?

geographika:
@jratike80 - it looks like this is supported.

Read from parameters:

https://github.com/MapServer/MapServer/blob/9d086cacc2df243d6ea900f0d6785cd0417a34a5/mapwms.cpp#L1160

Then applied to the format:

https://github.com/MapServer/MapServer/blob/9d086cacc2df243d6ea900f0d6785cd0417a34a5/mapwms.cpp#L1365

geographika:
After testing further with the [OUTPUTFORMAT](https://mapserver.org/mapfile/outputformat.html) blocks and default settings, there is no need for users to copy in the large chunks of default settings.

For example to set transparency for the `cairopng` driver setting the `DRIVER` and `TRANSPARENT` option is sufficient:

```
OUTPUTFORMAT
  DRIVER CAIRO/PNG
  TRANSPARENT ON
END
```

However PNG doesn't quite fit into this approach. If the following block is called with shp2img and the `-i png` option it will fail. 

```
OUTPUTFORMAT
  DRIVER AGG/PNG
  TRANSPARENT ON
END
```

The default driver name is set to `png24` for `DRIVER AGG/PNG`:

https://github.com/MapServer/MapServer/blob/9d086cacc2df243d6ea900f0d6785cd0417a34a5/mapoutput.c#L186

The default name however is `png`:

https://github.com/MapServer/MapServer/blob/9d086cacc2df243d6ea900f0d6785cd0417a34a5/mapoutput.c#L106

For tests to pass shp2img must pass `-i png24` to pick up the transparency, or the name must be added to the `OUTPUTFORMAT`:

```
OUTPUTFORMAT
  NAME "png"
  DRIVER AGG/PNG
  TRANSPARENT ON
END
```

Is there a reason for setting the default name to `png24`? Looking at the [code changes](https://github.com/MapServer/MapServer/commit/9d58b294fbcd4bd5152b860b88067de80c906a19#diff-dc05657831f040bd518a4b67c14dffecd832cd4d1cd837c904c8f4f272c23499L188) I think `png` wasn't used as this was used by the (long since removed) GD driver. 
I think it would make more sense to set the default name to `png`.

jratike80:
It is common to use 8 bit paletted png in WMS because for visual use 24 bit is mostly wasting bandwidth. Perhaps that is somehow related?

geographika:
> 
> 
> It is common to use 8 bit paletted png in WMS because for visual use 24 bit is mostly wasting bandwidth. Perhaps that is somehow related?

I think the WMS request would need to specify the format as `png8`. It looks like `png` == `png24` by default on the [AGG specifics page](https://www.mapserver.org/output/agg.html#setting-the-outputformat) and in the code.

mapoutput.c
```diff
@@ -183,7 +183,7 @@ outputFormatObj *msCreateDefaultOutputFormat( mapObj *map,
   }
 
   else if( strcasecmp(driver,"AGG/PNG") == 0 ) {
-    if(!name) name="png24";
+    if(!name) name="png";
```
geographika, line 186: Changing to png makes this consistent with the output default. png24 was used previously as png was mapped to the now removed GD/PNG driver.

PR #6330, opened 2021-05-15
Another batch of Coverity scan fixes

rouault:
Nothing that really deserves backporting. Mostly false positive and cleanups


jmckenna:
thanks @rouault !

PR #6331, opened 2021-05-19
MVT: Sets correct media type of Mapbox vector tiles output

ravhed:
The correct media type is defined as "vnd.mapbox-vector-tile" (see http://www.iana.org/assignments/media-types/media-types.xhtml)

There was some discussion regarding this in the following mail conversation: https://lists.osgeo.org/pipermail/mapserver-dev/2017-July/015172.html

GeoWebCache (and GeoServer) uses the vnd.mapbox-vector-tile media type. This causes some issues when wanting to use vector tiles generated by Mapserver as it doesn't return the media type that GeoWebCache expects. 

Worth to note is that this change might potentially break some existing installations which rely on the application/x-protobuf media type.


mapservutil.c
```diff
@@ -1530,7 +1530,7 @@ int msCGIDispatchImageRequest(mapservObj *mapserv)
     case TILE:
       msTileSetExtent(mapserv);
 
-      if(!strcmp(MS_IMAGE_MIME_TYPE(mapserv->map->outputformat), "application/x-protobuf")) {
+      if(!strcmp(MS_IMAGE_MIME_TYPE(mapserv->map->outputformat), "application/vnd.mapbox-vector-tile")) {
```
rouault, line 1533: you could accept both outputformats here for backward compatibility

mapwms.cpp
```diff
@@ -3770,7 +3770,7 @@ int msWMSGetMap(mapObj *map, int nVersion, char **names, char **values, int nume
   } else {
 
     /* intercept requests for Mapbox vector tiles */
-    if(!strcmp(MS_IMAGE_MIME_TYPE(map->outputformat), "application/x-protobuf")) {
+    if(!strcmp(MS_IMAGE_MIME_TYPE(map->outputformat), "application/vnd.mapbox-vector-tile")) {
```
rouault, line 3773: you could accept both outputformats here for backward compatibility

mapmvt.c
```diff
@@ -544,7 +544,7 @@ int msMVTWriteTile( mapObj *map, int sendheaders ) {
   if( sendheaders ) {
     msIO_fprintf( stdout,
 		  "Content-Length: %d\r\n"
-		  "Content-Type: application/x-protobuf\r\n\r\n",
+		  "Content-Type: application/vnd.mapbox-vector-tile\r\n\r\n",
```
rouault, line 547: instead of hardcoding the value, we could use MS_IMAGE_MIME_TYPE(map->outputformat)

jmckenna:
Thanks @ravhed I agree that this is an important change.  I also agree with @rouault's feedback about handling both media types, if possible.

mapserver.h
```diff
@@ -2807,7 +2807,7 @@ extern "C" {
   MS_DLL_EXPORT outputFormatObj *msSelectOutputFormat( mapObj *map, const char *imagetype );
   MS_DLL_EXPORT void msApplyOutputFormat( outputFormatObj **target, outputFormatObj *format, int transparent, int interlaced, int imagequality );
   MS_DLL_EXPORT const char *msGetOutputFormatOption( outputFormatObj *format, const char *optionkey, const char *defaultresult );
-  MS_DLL_EXPORT outputFormatObj *msCreateDefaultOutputFormat( mapObj *map, const char *driver, const char *name );
+  MS_DLL_EXPORT outputFormatObj *msCreateDefaultOutputFormat( mapObj *map, const char *driver, const char *name, const char *mimetype );
```
ravhed, line 2810: I'm not sure if this is the best solution to change the prototype of this function. I wanted to be able to get the mimetype from the default output format definition somehow. The only other way to accomplish the change I wanted was to match based on output format name here:

https://github.com/ravhed/mapserver/blob/fb57a5527255fb7b7dd00e8e0363205937ced87e/mapoutput.c#L221

I assume I will need to update the SWIG bindings if this is the way to go.

mapwms.cpp
ravhed: I've changed it now so it should be more backwards compatible than before - application/x-protobuf can still be requested successfully. It is also now still advertised in the GetCapabilities as a GetMap format.

The only backwards incompatibility I can think of now is that an output format defined like https://github.com/sdlime/mvt-demo/blob/master/demo.map#L24 will use correct mime type now instead of the old one.

mapserver.h
rouault: > I assume I will need to update the SWIG bindings if this is the way to go.

Perhaps add a msCreateDefaultOutputFormatEx() variant that takes the extra mimetype parameter. No need to map that to SWIG

rouault: actually I misunderstood your message. You probably meant that msCreateDefaultOutputFormat() is used by the SWIG bindings. So yes, maybe the simplest is to update them to use the new signature than creating a msCreateDefaultOutputFormatEx() variant

ravhed: This should be addressed now.

mapmvt.c
ravhed: Fixed, thanks

mapservutil.c
ravhed: Both are accepted now, thanks

PR #6336, opened 2021-05-22
Another batch of Coverity scan fixes

rouault:
We should be close to the end, as normally all remaining warnings should be adressed... (I'd expect a few ones requiring other follow up fixes though)


jmckenna:
thanks @rouault !

PR #6337, opened 2021-05-23
Update doxygen docstrings

geographika:
This pull request updates various MapServer functions, classes, and properties visible to SWIG generated bindings as part of [RFC 132](https://mapserver.org/id/development/rfc/ms-rfc-132.html).

### DIFF
--- a/mapserver.h
+++ b/mapserver.h
@@ -658,13 +658,16 @@ extern "C" {
   /*      used to hold aliases for TRUETYPE fonts                         */
   /************************************************************************/
 
+  /**
+  The :ref:`FONTSET <fontset>` object
+  */
   typedef struct {
 #ifdef SWIG
     %immutable;
 #endif
-    char *filename;
-    int numfonts;
-    hashTableObj fonts;
+    char *filename; ///< The filename of the fonset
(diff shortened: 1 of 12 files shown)


mapserver.h
```diff
@@ -2015,11 +2015,18 @@ void msPopulateTextSymbolForLabelAndString(textSymbolObj *ts, labelObj *l, char
 #endif /*SWIG*/
 
   /* Function prototypes, wrapable */
-  MS_DLL_EXPORT int msSaveImage(mapObj *map, imageObj *img, const char *filename);
-  MS_DLL_EXPORT void msFreeImage(imageObj *img);
-  MS_DLL_EXPORT int msSetup(void);
-  MS_DLL_EXPORT void msCleanup(void);
-  MS_DLL_EXPORT mapObj *msLoadMapFromString(char *buffer, char *new_mappath);
+  MS_DLL_EXPORT int msSaveImage(mapObj *map, imageObj *img, const char *filename); ///< Saves a map image to a file
```
rouault, line 2018: Some of the below makes long lines. using multiline /** */ syntax would be better and more common. That remark applies to maperror.h as well

PR #6341, opened 2021-05-26
Set locator and exception code so CITE WFS2 test will pass

PatrikSylve:
Sets the locator to "resourceid" and exception code to InvalidParameterValue when requested featureid is inconsistent with feature in TYPENAMES. 

From the CITE WFS2.0 test description: 
>If a feature instance identified by the RESOURCEID parameter is not of the type specified by the TYPENAMES parameter, the server shall raise an InvalidParameterValue exception where the "locator" attribute value shall be set to "RESOURCEID".

http://opengeospatial.github.io/ets-wfs20/apidocs/org/opengis/cite/iso19142/basic/filter/ResourceIdFilterTests.html#inconsistentFeatureIdentifierAndType-javax.xml.namespace.QName-


jmckenna:
thanks for this contribution @PatrikSylve !

PR #6342, opened 2021-05-26
Remove INTERLACE, IMAGEQUALITY, OPACITY keywords from SCALEBAR, LEGEND and MAP classes

geographika:
Removes the following keywords as part of [RFC 133](https://mapserver.org/development/rfc/ms-rfc-133.html)

[SCALEBAR>INTERLACE](https://mapserver.org/mapfile/scalebar.html#mapfile-scalebar-interlace)
[LEGEND>INTERLACE](https://mapserver.org/mapfile/legend.html#mapfile-legend-interlace)

[MAP>INTERLACE](https://mapserver.org/mapfile/map.html#mapfile-map-interlace)
[MAP>IMAGEQUALITY](https://mapserver.org/mapfile/map.html#mapfile-map-imagequality)
[MAP>OPACITY](https://mapserver.org/mapfile/map.html#mapfile-map-opacity)

As a result of this the `msApplyOutputFormat` function and its calls could be simplified throughout the codebase. 
msautotests tests have also been updated. 

Initially [SCALEBAR>TRANSPARENT](https://mapserver.org/mapfile/scalebar.html#mapfile-scalebar-transparent) and [LEGEND>TRANSPARENT](https://mapserver.org/mapfile/legend.html#mapfile-legend-transparent) were also going to be removed as they have been deprecated since 4.6. However if they are removed it is no longer possible to have a transparent scalebar or legend embedded within a non-transparent map - the `OUTPUTFORMAT` `TRANSPARENT` setting is applied to both. I think they should be kept and the deprecated status removed. 

The change most likely to impact users is that the following:
```
TRANSPARENCY 33
```

or 

```
OPACITY 33
```

Needs to be rewritten:

```
  COMPOSITE
    OPACITY 33
  END
```

We could keep one of the shortcuts if people think it would be more user friendly. 
OPACITY leave as a shortcut?


dmorissette:
FWIW I agree with keeping  SCALEBAR>TRANSPARENT and LEGEND>TRANSPARENT  since there are no alternatives at the moment and those settings are useful in real life.

geographika:
Merging as-is. `OPACITY` can be added back as a shortcut easily enough if this becomes an issue during 8.0 beta-testing.

PR #6344, opened 2021-06-02
Added loadParamsFromPost to owsrequest

JelmerBaas:
As discussed with Seth Girvin, I need a way to add the (XML) POST to MapScript. Not too sure this is the right way, to be honest.


geographika:
@JelmerBaas - do you have an example of how you are setting up the request and calling it?
I've tried the above in Python MapScript but only get back:

```
Content-Type: text/html

No query information to decode. QUERY_STRING not set.
```

geographika:
@JelmerBaas - #6345 builds on this, and includes a working Python test. The same approach should be available in C# on that branch.

PR #6345, opened 2021-06-06
Allow POST requests from MapsScript

geographika:
The following adds in POST request functionality from within MapScript via a new `loadParamsFromPost` method. Related to #6344.
Note - this will allow POST data to be passed as a string via MapScript, but does not support reading from a buffer. 

The only change to existing CGI functionality is allowing `request->contenttype` to be set directly on a request object rather than always reading from the CGI environment. 

Associated Python test added to the test suite, which can also serve as an example.

### DIFF
--- a/mapows.c
+++ b/mapows.c
@@ -175,8 +175,8 @@ static int msOWSPreParseRequest(cgiRequestObj *request,
     }
 
     /* Get service, version and request from root */
-    ows_request->service = (char *) xmlGetProp(root, BAD_CAST "service");
-    ows_request->version = (char *) xmlGetProp(root, BAD_CAST "version");
+    ows_request->service = msStrdup((char *) xmlGetProp(root, BAD_CAST "service"));
+    ows_request->version = msStrdup((char *) xmlGetProp(root, BAD_CAST "version"));
     ows_request->request = msStrdup((char *) root->name);
 
 #else
--- a/mapows.c
+++ b/mapows.c
@@ -175,11 +175,17 @@ static int msOWSPreParseRequest(cgiRequestObj *request,
     }
 
     /* Get service, version and request from root */
-    char* tmp = NULL;
-    if ((tmp = (char*)xmlGetProp(root, BAD_CAST "service")) != NULL)
-        ows_request->service = msStrdup(tmp);
-    if ((tmp = (char*)xmlGetProp(root, BAD_CAST "version")) != NULL)
-        ows_request->version = msStrdup(tmp);
+    xmlChar* serviceTmp = xmlGetProp(root, BAD_CAST "service");
+    if (serviceTmp != NULL) {
+        ows_request->service = msStrdup((char *) serviceTmp);
+        xmlFree(serviceTmp);
+    }
+
+    xmlChar* versionTmp = xmlGetProp(root, BAD_CAST "version");
+    if (versionTmp != NULL) {
+        ows_request->version = msStrdup((char *) versionTmp);
+        xmlFree(versionTmp);
+    }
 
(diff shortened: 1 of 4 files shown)


mapows.c
```diff
@@ -175,8 +175,14 @@ static int msOWSPreParseRequest(cgiRequestObj *request,
     }
 
     /* Get service, version and request from root */
-    ows_request->service = (char *) xmlGetProp(root, BAD_CAST "service");
-    ows_request->version = (char *) xmlGetProp(root, BAD_CAST "version");
+    char* tmp = NULL;
```
geographika, line 178: Memory issue here?

geographika:
The ASan (Address Sanitizer) is failing in Travis for most of the OWS msautotests with this pull request, although pass successfully without ASan. 

What is the best way to go about debugging these? [memory leaks](https://travis-ci.com/github/MapServer/MapServer/jobs/511476414#L12583) are mentioned in the logs, but not many further details (it says ASan logs are not empty but I don't know how these could be accessed). I'm guessing it is to do with the change to `ows_request->service` but I've tried a few approaches without any luck. 

Is it possible to add ASan to the vagrant builds?

mapows.c
rouault: yes...
Pedantically, xmlGetProp() returns a pointer that should be free'd with xmlFree() (see http://www.xmlsoft.org/html/libxml-tree.html#xmlGetProp), not with msFree(), but both are likely just going through system free(). The issue here is more than you need to call xmlFree(tmp) after setting ->service .

So the correct code is
```
 if ((tmp = (char*)xmlGetProp(root, BAD_CAST "service")) != NULL) {
        ows_request->service = msStrdup(tmp);
        xmlFree(tmp);
   }
    if ((tmp = (char*)xmlGetProp(root, BAD_CAST "version")) != NULL) {
        ows_request->version = msStrdup(tmp);
        xmlFree(tmp);
    }
```

rouault:
> Is it possible to add ASan to the vagrant builds?

we could, but you will get better diagnostics by running the process under Valgrind

geographika:
Thanks @rouault. I installed valgrind to the vagrant VM (`sudo apt-get install valgrind`), and then ran the msautotests with `pytest --valgrind`. 

Results did point to the modified `msOWSPreParseRequest` function:

`mapserver\msautotest\wxs\result\wcs_20_cap_inspire_embed.xml.vgrind.txt.txt`:

```
==29045== 4 bytes in 1 blocks are definitely lost in loss record 2 of 240
==29045==    at 0x4C31B0F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==29045==    by 0x8E0C0BC: xmlStrdup (in /usr/lib/x86_64-linux-gnu/libxml2.so.2.9.4)
==29045==    by 0x4EAEC82: msOWSPreParseRequest (mapows.c:179)
==29045==    by 0x4EAEC82: msOWSDispatch (mapows.c:258)
==29045==    by 0x4EB7BAC: msCGIDispatchRequest (mapservutil.c:1682)
==29045==    by 0x109052: main (mapserv.c:296)
```

mapows.c
geographika: Thanks @rouault - somehow I missed this comment, but (eventually!) came to the same solution. 
I initially copied the approach from: https://github.com/MapServer/MapServer/blob/2f74ad0bb9c26793ea08643576cbbd46aa11cbb6/mapwcs.cpp#L371

I don't see why it works there (there seems to be no `xmlFree(tmp)` in this function), yet the tests pass with no memory leaks.

PR #6346, opened 2021-06-08
Raster classification: better take into account DATARANGE (fixes #6287)

### LINKED ISSUE #6287: DATARANGE and COLORRANGE doesn;t work with GeoTiff with band datatype byte
I use in mapfile:
```
CLASS
	EXPRESSION ([pixel] !=38)
	STYLE
	COLOR 00 00 00 
	END
	
END	
CLASS
	EXPRESSION ([pixel] == 38)
	STYLE
		COLORRANGE 0 0 0 100 0 0
		DATARANGE 0 100
	END
END
```
Rendered pixel has value in image is 100 0 0 not as expected 38 0 0. Input geotif has band datatype byte. When i convert geotiff to int16 or uint everything works as expected.
I include zip file contains ok and wrong map file points to different geotiff 
[report.zip](https://github.com/MapServer/MapServer/files/6280241/report.zip)

rouault:
- In 8 bit mode, when there was a single STYLE with DATARANGE/COLORRANGE,
  it was used even for values outside of DATARANGE, which seems
  inappropriate
- In 16 bit mode, there was no check on the DATARANGE. So if several
  STYLE were defined in the CLASS, only the first one was taken into
  account.


jmckenna:
Thanks for this change @rouault Do you think we should also include a test for this case?

rouault:
> Do you think we should also include a test for this case?

I'll adjust existing test(s)

rouault:
> I'll adjust existing test(s)

commit amended with updated tests demonstrating the fixed behaviour

jmckenna:
thanks @rouault !

PR #6352, opened 2021-06-22
WMS: make attribute color binding work with FILTER vendor parameter (fixes #6200)

### LINKED ISSUE #6200: Vendor WMS parameter FILTER not work with attribute parameters in STYLE
Hi, using [attribute] for example in COLOR or OUTLINECOLOR, and then a FILTER vendor parameter is applied from a client, it does not return any geometry

```
LAYER
  ...
  STYLE
    ...
    COLOR [color]
```

On the contrary, if the FILTER is applied and the layer has a color defined in hexadecimal or rgb if it works

rouault:
@sdlime The root cause of the issue was due to the ``!MS_DRAW_QUERY(drawmode))`` test in https://github.com/MapServer/MapServer/blob/18d3a67d22fd596920d51fc1f827a4ffa4be126b/maputil.c#L157 that prevented the attribute binding to be resolved, since WMS FILTER relies on querymap rendering. I'm not completely sure why this test was there, but I suspect it only applies to the STYLE hilite mode describe in https://mapserver.org/mapfile/querymap.html . Hence my proposed fix which adds the MS_DRAWMODE_QUERY flag only in hilite mode, and not in the selected mode used by WMS FILTER.


jmckenna:
Thanks @rouault !

I believe this fix can be backported.  Let's see...

PR #6358, opened 2021-07-05
Update Appveyor builds to Visual Studio 2019

geographika:
Updates the Appveyor Windows CI to use Visual Studio 2019, and the new MSVC 2019 release-1928-x64-dev buildkit from https://www.gisinternals.com/sdk.php

- update to MSSQL 2019
- updated to use PROJ7
- test JSON results now all updated to include `"crs": { "type": "name", "properties": { "name": "urn:ogc:def:crs:EPSG::3857" } },` - likey due to updates to the GDAL dependencies in the buildkit

One strange issue is that for JSON responses, there is now an additional `\r` in the break between the header output and the rest of the JSON response body. This broke the Python code to DEMIME a response in `mapserver\msautotest\pymod\mstestlib.py`
A fix was added to the Python code to get this working again, but an upstream fix would be better. 

image

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,7 +2,7 @@ branches:
   except:
     - /(cherry-pick-)?backport-\d+-to-/
 
-image: Visual Studio 2017
+image: Visual Studio 2019
 
 cache:
   - '%LOCALAPPDATA%\pip\Cache'
@@ -18,7 +18,7 @@ environment:
       secure: mHoJHeXdXbBNoDf7MA4ZEg==
 
   # VS 2017
-  VS_VERSION: Visual Studio 15 2017
+  VS_VERSION: Visual Studio 16 2019
   matrix:
   - platform: x86
     Python_ROOT_DIR: c:/python27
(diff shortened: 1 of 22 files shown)


geographika:
I think the extra `CR` seem to be due to an update to the cURL dependency rather than any changes in the MapServer codebase.

geographika:
Unless @szekerest has any comments/objections I'll merge this in the next few days.

rouault:
> I think the extra `CR` seem to be due to an update to the cURL dependency rather than any changes in the MapServer codebase.

That's weird. I don't see where curl is triggered by those tests.

geographika:
> 
> 
> > I think the extra `CR` seem to be due to an update to the cURL dependency rather than any changes in the MapServer codebase.
> 
> That's weird. I don't see where curl is triggered by those tests.

On furhter thought probably nothing to do with cURL. There were no MapServer code changes as part of this pull request,  the "only" changes were  to the Appveyor image and the dependencies, so it could be one of:

- Visual Studio 2017 > 2019
- Windows Server 2016 > Windows Server 2019
- Updates to a GDAL or Proj function used in MapServer?

Issue #5096, opened 2015-05-08
Symbol is not displayed  in legend in Mapserver 7 without keyimage


wirkus:
Hello,

I'm testing Mapserver7. Everything is OK, but I have problems with symbols in the legend.
The mapserver-symbols and png-symnbols are displayed in the Map but not in the legend, see below.
IF I use a KEYIMAGE, the legend is displayed.

SIZEUNITS meters
CLASS
NAME "Eingriff"
STYLE
SYMBOL 'circle'
COLOR 255 20 0
SIZE 12
MINSIZE 8
MAXSIZE 16
OUTLINECOLOR 0 0 0
WIDTH 1
MINWIDTH 1
MAXWIDTH 1
END
LABEL
TYPE TRUETYPE
#COLOR 0 255 55
COLOR 255 0 0
OUTLINECOLOR 254 254 254
FONT "Arial"
POSITION uc
PARTIALS FALSE
FORCE TRUE
MINDISTANCE -1
SIZE 9
MINSIZE 8
MAXSIZE 12
END
END

Do I have to define a keyimage?

Best regards

tbonfort:
I suspect this has something to do with using SIZEUNITS. What happens if you remove that?

astroidex:
Yes, @tbonfort. It has to do with using SIZEUNITS. @Schpidi and I were testing the issue at FOSS4G code sprint. @Schpidi changed the mapfile https://github.com/mapserver/msautotest/blob/master/renderers/legend.map in every layer we said SIZEUNITS meters. The effect was:

* labeling in the legend is much smaller -> depends on the SIZE of label in CLASS
* some symbols are not displayed at all (all point symols)
* polygon symbols - outline is not displayed at all

Schpidi:
Original:
![legend](https://cloud.githubusercontent.com/assets/482265/8761506/f8c80b5c-2d54-11e5-90d7-fa8bd023a812.png)

SIZEUNTES METERS:
![legend](https://cloud.githubusercontent.com/assets/482265/8761507/fe1c46b8-2d54-11e5-9907-4b1e80de5471.png)

wirkus:
Hello,
we tested point symbols in the legend again:
 * SIZEUNITS pixel is OK
 * SIZEUNITS meters: umn symbol: no image in the legend but text
 * SIZEUNITS meters: png symbol: no image in the legend but text
[...]

PR #6359, opened 2021-07-05
Change legend symbol scale factor when SIZEUNITS is set to METERS

PatrikSylve:
When SIZEUNITS METERS is set in LAYER, the symbol won't show in the legend. This is due to that the scale factor will get the value INFINTY.

Related issue: 
https://github.com/mapserver/mapserver/issues/5096


sdlime:
Would a better test be when layer size units is not given in pixels?

jmckenna:
@PatrikSylve if these tests pass we can include this in today's 7.6.4 release.

PatrikSylve:
> @PatrikSylve if these tests pass we can include this in today's 7.6.4 release.

Seems like the PDF image is still not passing the test here. 

I had some issues with running the tests on my machine so I used the following command to generate the expected images:
> ./mapserv -nh "QUERY_STRING=map=/home/patriksylve/code/MapServer/msautotest/renderers/legend_sizeunits_meters.map&mode=legend&map.imagetype=pdf" >> img/legend_sizeunits_meters.pdf 

The results look fine to me. Could it maybe be caused by differing dependency versions or something like that?

jmckenna:
@PatrikSylve I understand your frustration - I've now setup your branch locally on Windows here and 3 tests fail here for `legend_sizeunits_meters.map` (the non-cairo PNG, and the PDF & SVG.  (I can see one-digit precision differences in the SVG numbers)  Hmmmmmmm.....

PatrikSylve:
@jmckenna Hm ok. That is strange.  

I managed to get the tests running on my linux now, and it now says that all 4 tests passed. I tried compiling mapserver on another linux as well and got the same result. 

However, when i run local tests on legend.map the PDF-format fails, but the resulting legend.pdf has the same hash as the legend_sizeunits_meters.pdf. So I will try to commit the legend.pdf but renamed as legend_sixeuntis_meters.map. Hopefully the test will pass now.

PatrikSylve:
Seems like it was different versions of cairo that caused the problem: 
pdf_diff

jmckenna:
Thanks for all of this effort @PatrikSylve !  (you're likely right, caused by Cairo version differences)

jmckenna:
This has been backported to branch-7-6 manually

PR #6368, opened 2021-07-09
PostGIS backend: do not try fetching OID with PostgreSQL >= 12

rouault:


jmckenna:
thanks so much @rouault great work!

PR #6370, opened 2021-07-10
Fix python module installation with alternative root/prefix

landryb:
Trying to package/build mapserver head before 8.0, it failed for me during install/fake, as it tried writing to the system dirs in usr/local while CMAKE_INSTALL_PREFIX and DESTDIR were properly defined - and ignored.

i think this was an oversight in the large commit done in dd310d211. Not 100% confident about all cases, but this fixes the 'install to alternate dir' problem for me. Cmake expert feedback welcome :)

### DIFF
--- a/mapscript/python/CMakeLists.txt
+++ b/mapscript/python/CMakeLists.txt
@@ -123,7 +123,7 @@ install(
     endif()
 
     execute_process(
-      COMMAND ${Python_EXECUTABLE} setup.py install \${Python_ROOT} \${Python_PREFIX}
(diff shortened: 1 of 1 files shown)


landryb:
the other fix would be to set `Python_ROOT` and `Python_PREFIX` in the lines above instead of setting their uppercase variants.

geographika:
Looks good to me. The [CMake Python variables](https://cmake.org/cmake/help/latest/module/FindPython.html#hints) are all named similar to `Python_ROOT_DIR`, whereas `PYTHON_ROOT` and `PYTHON_PREFIX` are custom variables that are set in the script and are case-sensitive.

geographika (approved):

Issue #6376, opened 2021-07-21
The "composite" is not described in the mapfile.xsl


gisspace:
![image](https://user-images.githubusercontent.com/6889060/126450926-22839f3d-c00f-4826-b81f-f61892c7f43e.png)
The "composite" is not described in the mapfile.xsl

jmckenna:
Great catch @gisspace.  Can you create Pull Request for this change, against the main branch?

gisspace:
OK

jmckenna:
Fixed by @gisspace through https://github.com/MapServer/MapServer/pull/6377

PR #6377, opened 2021-07-28
The "composite" is not described in the mapfile.xsl

gisspace:
#6376 fix The "composite" is not described in the mapfile.xsl


jmckenna:
thanks @gisspace !

PR #6388, opened 2021-08-20
WMS capabilities output: increase precision of long/lat in bounding boxes to 6 decimal places (fixes #6387)

### LINKED ISSUE #6387: Precision of Layer Extents in GetCapabilities
I specify Layer Extents in my MapFile like so:

    EXTENT 15.380859375 25.224609375 21.181640625 32.87109375

Then I issue a GetCapabilities request, like so:

    http://localhost/mapserv?map=/mapfiles/map.map&service=wms&request=getcapabilities

and the reported extent have four digits of precision:

```xml
<Layer queryable="0" opaque="0" cascaded="0">
     ...
    <CRS>EPSG:4326</CRS>
    <EX_GeographicBoundingBox>
        <westBoundLongitude>15.3809</westBoundLongitude>
        <eastBoundLongitude>21.1816</eastBoundLongitude>
        <southBoundLatitude>25.2246</southBoundLatitude>
        <northBoundLatitude>32.8711</northBoundLatitude>
    </EX_GeographicBoundingBox>
  <BoundingBox CRS="EPSG:4326" minx="25.2246" miny="15.3809" maxx="32.8711" maxy="21.1816"/>
  ...
  </Layer>
```

The extents are written here: 
https://github.com/MapServer/MapServer/blob/5dbbf8a19abb2f30b4852b72c8661faac14ae4c9/mapmetadata.c#L313-L317

and _msMetadataGetDecimal prints the float with `%f` and no digits of precision https://github.com/MapServer/MapServer/blob/5dbbf8a19abb2f30b4852b72c8661faac14ae4c9/mapmetadata.c#L216

Some context as to why this matters: WMS layers in QGIS limit WMS requests to the extents returned by the `GetCapabilities` request. This can effectively truncate valid imagery our users need to see. [While four digits are certainly good enough for some applications](https://observablehq.com/@mourner/latitude-and-longitude-precision), others may require more.

Related https://github.com/MapServer/MapServer/issues/6208: WFS support recently got a new `gml_msGeometry_precision`. Not sure if this fix should do one of the following:
* Use the same metadata keyword
* Add a new keyword
* Hard-code more digits  in [_msMetadataGetDecimal](https://github.com/MapServer/MapServer/blob/5dbbf8a19abb2f30b4852b72c8661faac14ae4c9/mapmetadata.c#L211-L228)

rouault:


jmckenna:
@pedros007 can you confirm that this pull request fixes your issue?

pedros007:
Yes, will do. Judging by the diff it looks good, but we'll load things into QGIS just to be sure. I'll write back when we have confirmed.

rouault:
Last commit fixes the mystery of why the msautotest results had only 5 decimal figures instead of the expected 6 !

PR #6392, opened 2021-08-27
update SECURITY.md for supported versions

jmckenna:
- according to the mapserver-dev thread (https://lists.osgeo.org/pipermail/mapserver-dev/2021-July/016572.html) and the recent discussion in the PSC meeting on 2021-08-26 (https://github.com/MapServer/MapServer/wiki/PSC-Meeting-2021-08-26)
- this was the suggestion by Daniel, Even, Jeff (all with the same idea, yet different wording).
- the wording that seems to be the most clear was provided by Daniel, and is used in this pull request

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -10,12 +10,16 @@ submissions, when describing the vulnerability (see https://mapserver.org/develo
 
 ## Supported Versions
 
-The MapServer PSC (Project Steering Committee) will release patches for security vulnerabilities for the following versions:
+The MapServer PSC (Project Steering Committee) will release patches for security vulnerabilities 
+for the last release branch of the two most recent major releases.  For example, 
+once 8.0 is released, then only 8.0.x and 7.6.x will be patched.
+
+Currently, the following versions are supported:
 
 | Version | Supported          |
 | ------- | ------------------ |
 | 7.6.x   | :white_check_mark: |
 | 7.4.x   | :white_check_mark: |
-| 7.2.x   | :white_check_mark: |
-| 7.0.x   | :white_check_mark: |
+| 7.2.x   | :x:                |
+| 7.0.x   | :x:                |
 | < 7.0   | :x:                |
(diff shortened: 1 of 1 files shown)


sdlime:
We did not patch latest 6.x with the last security release because of age - part of my rationale for setting some sort of a time limit. So, do we need to go back and patch that version or are we essentially considering it unsupported at this point and the SECURITY.md is forward thinking. I don't consider 6.x supported at this point.

We should also have backport workflows for the previous 2 branches - so a backport 7 and a backport 8 here soon.

--Steve

jmckenna:
Will setup backport workflows for 8 and 7, during release day./

dmorissette:
I found that a 1 year time limit was a bit short, but when I wrote my suggestion I didn't realize that 6.4.0 was released in 2013 - 8 years ago - which is definitely quite a stretch.  How about a 3 or 5 years time limit on patching the older release?  

(The rationale for 3 to 5 years is to try to follow what distros such as Ubuntu do even if it's not a direct match. In the case of Ubuntu 20.04, it uses MapServer 7.4 and is supported until 2025)

jmckenna:
I personally think we don't have the funding to focus on years and math each time - since we are all volunteers here (and only a few care to take the time to make these patch releases), and this just adds confusion for the users, who never know what year a release was made. I again suggest to only support the previous 2 release branches (so when 8.0 is released, we support 8.0 and 7.6 branch, and when 8.2 is released we only support the 8.2 and 8.0 branches).  This means at the current status we would only support the 7.6 and 7.4 branches.

sdlime:
I'd opt for as short as folks feel is reasonable - I guess 3 would be ok but that's a long time. Note under this policy 7.4 will not get patches, only the latest in the 7.x release. How do we get the packagers on board - perhaps just stating things in a policy is good enough?

sdlime:
> I personally think we don't have the funding to focus on years and math each time - since we are all volunteers here (and only a few care to take the time to make these patch releases), and this just adds confusion for the users, who never know what year a release was made. I again suggest to only support the previous 2 release branches (so when 8.0 is released, we support 8.0 and 7.6 branch, and when 8.2 is released we only support the 8.2 and 8.0 branches). This means at the current status we would only support the 7.6 and 7.4 branches.

Actually we'd only support latest branch in the 7.x series - so 7.6.x.

PR #6400, opened 2021-09-01
WIP: OGCAPI Tweaks/Fixes

sdlime:
Series of small changes/fixes/tweaks necessary to complete initial OGC API features support.


sdlime:
@jmckenna, @tomkralidis, @rouault  or anyone else - feedback is welcome. The main items addressed here thus far are:

- fixing the API page rendering via Swagger (minor issue with openapi document)
- getting paging controls working for collection/items
- integrating Leaflet maps into collection/items and collection/items/{id}

One thing that probably needs to happen is figuring out how to integrate the value for oga_max_limit into the UI. There no point giving users a control that sets a limit to 10000 when it's configured for a max of 25.

Not sure what else needs addressing - perhaps we could add a task list to this pull request?

--Steve

jmckenna:
@sdlime sorry for the delay in feedback (better late than never, ha).  Looks great!  Below is some testing feedback on MS4W, using the old task list I had created earlier in another document (which tried to collect the ideas/wishlist from everyone at the time) :

### View of (new) API Documentation

(for those who haven't seen it yet...)

ogcapi-api-documentation

### Original issues (and updated status)

* **issue:** cleaning the onlineresource of "?" or "/" ending characters, as it causes problems in HTML landing pages (double-slashes in links, or the "?" will actually break all links)
  * **status**: problem still exists :x:
* **issue:** compile fails without enabling `-DUSE_WFS_SVR` (as well as the obvious `-DUSE_OGCAPI_SVR`)
  * **status**: I cannot reproduce this, maybe someone else reported :question:
* **task:** add ows_contact* information to the landing page (from the associated values set in the mapfile)
  * **status**: it added on the API documentation page, but not the landing page :question:
* **wishlist:** besides the "/ogcapi" API signature (for features, maps..support), also allow "/features" (to allow developers to only request vector collections), for example:
  ```
   http://127.0.0.1/cgi-bin/mapserv.exe/demo-ogcapi/features
  ```
  (Note that is how the GeoServer implementation works)
* **wishlist:** include MapServer logo on landing page :heavy_check_mark:
* **wishlist:** set MapServer favicon for all HTML pages :heavy_check_mark:
* **wishlist:** allow other link types for collection item (set through mimetype/outputformats in mapfile?)
* **wishlist:** instead of text links, use a dropdown for types (JSON, HTML, KML, etc.)
* **wishlist:** for the single layer map, include a dropdown to set the default number of features displayed in the map: 
  ```
    10 [default]
    100
    1000 
  ```
* **wishlist:** for the single layer map, display the current MapServer scale value below the map (as user zooms in, update the scale with the CGI [scale] value)

geographika:
@sdlime - some further feedback from testing.

1. There seem to be issues (at least on IIS) when trying to calculate the root API in `getApiRootUrl`:

https://github.com/MapServer/MapServer/blob/17d8fa9730768c2122cbc78f07598b365d1a6c37/mapogcapi.cpp#L282

A URL such as http://localhost/mapserver/mymap/ogcapi/collections/MyDataSet/items?f=html returns the following server variables in IIS:

+ SERVER_NAME: localhost
+ SERVER_PORT: 80
+ SCRIPT_NAME: /mapserver
+ PATH_INFO: /mymap/ogcapi/collections/MyDataSet/items

Then when constructing URLs such as here: 

https://github.com/MapServer/MapServer/blob/17d8fa9730768c2122cbc78f07598b365d1a6c37/mapogcapi.cpp#L1194

The URL created for the template is http://localhost/mapserver/mymap/ogcapi/collections/MyDataSet/items/collections?f=html
`collections` is added twice and the result link then returns `{"code":"NotFound","description":"Collection items id query failed."}`

The good new is that setting the WEB METADATA option fixes all these links:

`"oga_onlineresource" "http://localhost/mapserver/mymap/ogcapi"`

2. Again when using IIS (could be the same on other servers?), any cookie values sent to the server are seen as parameters. The following code then throws an error when a cookie is included in the request:

https://github.com/MapServer/MapServer/blob/17d8fa9730768c2122cbc78f07598b365d1a6c37/mapogcapi.cpp#L990-L992

{"code":"InvalidParamterValue","description":"Unknown query parameter: ASP.NET_SessionId"}

Could the check simply be removed?
Deleting the cookie allows the site to run, but if the URL is shared with another application then this will cause issues. 

3. As mentioned [in the previous comment](https://github.com/MapServer/MapServer/pull/6180#issuecomment-881901488) every layer in the mapfile requires `"ows_extent"` to be set. 
However navigating to http://localhost/mapserver/mymap/ogcapi/api throws HTTP 500 errors if any layers are missing the following: "ows_title" or missing one of "oga_description" or "ows_abstract" - the strings are then NULL which crashes the JSON templating. Relevant lines:

https://github.com/MapServer/MapServer/blob/17d8fa9730768c2122cbc78f07598b365d1a6c37/mapogcapi.cpp#L1443
https://github.com/MapServer/MapServer/blob/17d8fa9730768c2122cbc78f07598b365d1a6c37/mapogcapi.cpp#L1444

After updating the Mapfile (with 90+ layers) everything worked perfectly - this is such an amazingly powerful addition to MapServer - many thanks!

sdlime:
> @sdlime - some further feedback from testing.

Thanks so much for the testing! I figured we might have an IIS'ism or two to deal with.

> 1. There seem to be issues (at least on IIS) when trying to calculate the root API in `getApiRootUrl`:

Trying to compute is a last resort - it's the same approach taken with the WxS services I believe.  So setting "oga_onlineresource" is definitely going to be recommended and apparently required for IIS. I think this is a documentation issue...

> 2. Again when using IIS (could be the same on other servers?), any cookie values sent to the server are seen as parameters. The following code then throws an error when a cookie is included in the request:

Oh, good catch. This would apply to all servers/platforms. @rouault had added that check when he was fixing up my code. I believe that check is only done with a _collection/items_ request. That behavior is consistent with pygeoapi so I wonder if that is part of the OGC specification, @tomkralidis? I suppose there are 2 fixes:

1. Remove the check and simply ignore any extra parameters.
2. Remove treating cookies as parameters. It's a big hammer but I wonder if it's really necessary functionality. It's definitely very old and I don't believe it was driven by anything other than convenience.

> 3. As mentioned [in the previous comment](https://github.com/MapServer/MapServer/pull/6180#issuecomment-881901488) every layer in the mapfile requires `"ows_extent"` to be set.
>    However navigating to http://localhost/mapserver/mymap/ogcapi/api throws HTTP 500 errors if any layers are missing the following: "ows_title" or missing one of "oga_description" or "ows_abstract" - the strings are then NULL which crashes the JSON templating. 

Definitely a bug...

tomkralidis:
>> Again when using IIS (could be the same on other servers?), any cookie values sent to the server are seen as parameters. The following code then throws an error when a cookie is included in the request:
>
> Oh, good catch. This would apply to all servers/platforms. @rouault had added that check when he was fixing up my code. I believe that check is only done with a collection/items request. That behavior is consistent with pygeoapi so I wonder if that is part of the OGC specification, @tomkralidis?

See [ATS 21](http://docs.ogc.org/is/17-069r3/17-069r3.html#ats_core_query-param-unknown) from OAFeat.  It's expected to throw a 400 when encountering a query parameter that is not part of the API definition.

sdlime:
> > > Again when using IIS (could be the same on other servers?), any cookie values sent to the server are seen as parameters. The following code then throws an error when a cookie is included in the request:
> > 
> > 
> > Oh, good catch. This would apply to all servers/platforms. @rouault had added that check when he was fixing up my code. I believe that check is only done with a collection/items request. That behavior is consistent with pygeoapi so I wonder if that is part of the OGC specification, @tomkralidis?
> 
> See [ATS 21](http://docs.ogc.org/is/17-069r3/17-069r3.html#ats_core_query-param-unknown) from OAFeat. It's expected to throw a 400 when encountering a query parameter that is not part of the API definition.

Hmmmm... Pygeoapi does errors on collection items but not elsewhere. For example, https://demo.pygeoapi.io/master/collections?f=html&foo=bar resolves correctly. I guess it depends on if  a query parameter refers to filtering the data or anything in the QUERY_STRING. I see this also being a limitation when using these services with AJAX clients. For example, jquery tags on a parameter like "_=1631904651203" to make a call unique and that generates an error with pygeoapi and MapServer collection/items requests (JSON or HTML).

tomkralidis:
Perhaps then the OAFeat CITE tests need to updated if ATS 21 interpreted differently?  Have we tried running a demo against CITE?

rouault:
>  Have we tried running a demo against CITE?

If my memories don't lie, I did when I submitted several fixes, and they ended up green. But CITE tests don't necessarily fully and perfectly test a spec

sdlime:
> 1. As mentioned [in the previous comment](https://github.com/MapServer/MapServer/pull/6180#issuecomment-881901488) every layer in the mapfile requires `"ows_extent"` to be set.

@geographika, I don't believe ows_extent is required. None of my test mapfiles have that set so it's being computed.

geographika:
> @geographika, I don't believe ows_extent is required. None of my test mapfiles have that set so it's being computed.

Looking at the code it should end up calling `msLayerGetExtent` if no metadata or LAYER EXTENT is found. I'll check my layers again - there could be a non-spatial layer somewhere in there causing this. So looks like a local issue to my setup.

geographika:
@sdlime - ignore the extent issue - this seems to only occur on layers without geometries (attributes only). It does however cause the whole catalog page to load. 

I've been testing out the OGC Features API setup on the Itasca demo - see https://github.com/MapServer/MapServer-demo/pull/7/files
One issue I ran into was if I don't specify a projection block on a layer then the GeoJSON coordinates are returned in `'init=epsg:26915' ` coords and don't display properly in the preview maps. 
If I add a projection block to the layer then they are returned in epsg:4326 as per the spec and all is fine:

Following works:

```
  LAYER
    NAME airports
    TYPE POINT
    DATA airports
    PROJECTION
        'init=epsg:26915' 
    END
```
According to https://mapserver.org/mapfile/projection.html#important-notes

> If all of your data in the mapfile is in the same projection, you DO NOT have to specify any projection objects. MapServer will assume that all of the data is in the same projection.

Is this an error with the docs or should the OGC Features API be reprojecting these features without each LAYER requiring its PROJECTION block?

geographika:
@sdlime - also another item, maybe just something to document. If a layer doesn't have a template, and is therefore not queryable:

https://github.com/MapServer/MapServer/blob/f9a2df886828cbb8cd090072480dfe6c2483a3c2/mapquery.c#L140

Then the error `{"code":"NotFound","description":"Collection items query failed."}` is returned when accessing and ot the `items` API points e.g. `ogcapi/collections/mcd90py2_anno/items`:

https://github.com/MapServer/MapServer/blob/f9a2df886828cbb8cd090072480dfe6c2483a3c2/mapogcapi.cpp#L1009

Is the whole mechanism of requiring `LAYER TEMPLATE` being set still valid within MapServer? Adding `LAYER TEMPLATE "ttt"` seems to be almost required now to do most things in MapServer even if no templates are ever used.

jratike80:
OAPIF feels the same as WFS and WFS does not require templates, enabling the service with "wfs_enable_request" is enough.

geographika:
> One issue I ran into was if I don't specify a projection block on a layer then the GeoJSON coordinates are returned in `'init=epsg:26915' ` coords and don't display properly in the preview maps.

@sdlime - ignore the above - it looks like it was an issue with CMAKE not picking up PROJ_VERSION_MAJOR correctly, and therefore not copying the projection from the map to layer on initialisation.

geographika:
> 
> 
> OAPIF feels the same as WFS and WFS does not require templates, enabling the service with "wfs_enable_request" is enough.

Does this work currently? I can't see how this is set in the code (maybe a fake template added automatically?) `msIsLayerQueryable` (which checks for a layer or class template) seems to be called for any MS query.

jratike80:
> Does this work currently? I can't see how this is set in the code (maybe a fake template added automatically?) `msIsLayerQueryable` (which checks for a layer or class template) seems to be called for any MS query.

I don't have Mapserver 8 for testing. In my old mapfiles I have also `"gml_include_items" "all"` if that matters.

sdlime:
> @sdlime - also another item, maybe just something to document. If a layer doesn't have a template, and is therefore not queryable:
> 
> https://github.com/MapServer/MapServer/blob/f9a2df886828cbb8cd090072480dfe6c2483a3c2/mapquery.c#L140
> 
> Then the error `{"code":"NotFound","description":"Collection items query failed."}` is returned when accessing and ot the `items` API points e.g. `ogcapi/collections/mcd90py2_anno/items`:
> 
> https://github.com/MapServer/MapServer/blob/f9a2df886828cbb8cd090072480dfe6c2483a3c2/mapogcapi.cpp#L1009
> 
> Is the whole mechanism of requiring `LAYER TEMPLATE` being set still valid within MapServer? Adding `LAYER TEMPLATE "ttt"` seems to be almost required now to do most things in MapServer even if no templates are ever used.

Seems  to me like the layer shouldn't even be in the list of collections in the first place if it's not queryable.

geographika:
@sdlime - I've been debugging the issue around layers without a projection block not projecting to WGS84 when returning results (although this may be a feature as much as an issue - I'm unsure if the OGC Features Spec allows for other CRSs?). 

I've updated the Itasca demo to be accessible all layers via the OGC Features API (at https://github.com/MapServer/MapServer-demo/pull/7 - there are a few values that need to be changed, and I'll switch them to Linux/OSGeo paths before merging). Just wondering if any of the layers without a projection block show in EPSG:26915 for you (e.g. mcd90py2_anno)?

https://github.com/MapServer/MapServer-demo/pull/7

`msProjectionsDiffer` always returns false for me here:

https://github.com/MapServer/MapServer/blob/1ed91b8e580f37434ca365a053e10351ad685eec/mapogcapi.cpp#L1104-L1106

As the LAYER PROJECTION is an unitialized object it has `numargs` count of 0 and so is shown as not differing here:

https://github.com/MapServer/MapServer/blob/1ed91b8e580f37434ca365a053e10351ad685eec/mapproject.c#L2151-L2156

sdlime:
> @sdlime - I've been debugging the issue around layers without a projection block not projecting to WGS84 when returning results (although this may be a feature as much as an issue - I'm unsure if the OGC Features Spec allows for other CRSs?).

There is a CRS standard (https://docs.ogc.org/is/18-058/18-058.html) but we've not done any implementation around that so this is just the Features Core (http://docs.ogc.org/is/17-069r3/17-069r3.html) implementation.

Regarding this issue it looks to me like the code is incomplete in that area. If you look at the code starting at: 
https://github.com/MapServer/MapServer/blob/1c14821bc4c6fc10e966fd05230ad71c05817c7d/mapogcapi.cpp#L1104-L1106

it only looks at the layer projection relative to WGS84. I think the logic should be:

1. if layer has projection and that projection isn't WGS84, set the reprojector
2. else if the map has a projection and that projection isn't WGS84, set the reprojector
3. else throw a configuration exception

Cases 2 and 3 aren't handled. Does that make sense to you?

sdlime:
@geographika, I just updated things to consider both the layer and map projections when transforming the layer bbox and geometries (items). Right now it throws an exception if both the layer and map projections are not defined. It works if the layer does not have a projection defined - defaults to the map projection. Please give it a try. --Steve

geographika:
> @geographika, I just updated things to consider both the layer and map projections when transforming the layer bbox and geometries (items).

Tested and works perfectly with the Itasca demo. The exception makes sense, I'd assumed that the MAP PROJECTION was already copied to the LAYER PROJECTION as part of loading a Mapfile if none was explicitally set on a LAYER, but I guess each service (WMS, WFS etc) must be implementing this logic on its own code path?

geographika:
@sdlime - from my testing I think there are only 2 issues left to handle 

1. Dealing with parameters set in cookies (just seen your mail now). I tried a very hacky fix by adding the following to mapserv.c but it didn't work as expected:

```
  // ignore any cookie values which will be treated as parameters when using the ogcapi
  if (request->path_info != NULL) {
      if (strstr("/ogcapi/", request->path_info) != NULL) {
          putenv("HTTP_COOKIE", NULL);
      }
  }
```

2. (nice to have as the slash is required on Windows) adding a slash if missing to the end of `OGCAPI_HTML_TEMPLATE_DIRECTORY`. See comment at https://github.com/MapServer/MapServer/blob/1c14821bc4c6fc10e966fd05230ad71c05817c7d/mapogcapi.cpp#L241

I tried to implement this, but thought there might already be a function somewhere in MapServer that could be reused. 
If not there is a C implementation in the Git codebase https://github.com/git/git/blob/e5a14ddd2d93da4d951fd63d4f78fe2410debe68/builtin/mv.c#L63

I also tried a hybrid C/C++ approach, but it throws memory errors:

```
  if (directory != NULL) {
      std::string s = directory;
      if (*s.rbegin() != '/'){
          // add a trailing slash if missing
          s += "/";
          return  s.c_str();
      }
  }
```

rouault:
> if (*s.rbegin() != '/'){
slightly less convoluted:
if( !s.empty() && s.back() != '/' )

> return  s.c_str();
That's the memory error. std::string s is a temporary object. a c_str() on it is invalidated as soon as the std::string is destroyed, that is when the scope where it is declared is quitted

A fix would be to change getTemplateDirectory() to return a std::string (be careful of changing the return NULL to return std::string()), and adapt its callers

sdlime:
Should we merge this and deal with additional tweaks in subsequent pull requests? --Steve

jmckenna:
> Should we merge this and deal with additional tweaks in subsequent pull requests? --Steve

yes I think we should merge.

geographika:
> 
> 
> Should we merge this and deal with additional tweaks in subsequent pull requests? --Steve

+1 on merging, it will make it easier to test with the other new features on the main branch. 
@sdlime - any thoughts on handling the backslash - C++ approach (as in the approach suggested at https://github.com/MapServer/MapServer/pull/6400#issuecomment-927094508 or a new C function)?

I think the cookie issue will need to be addressed before a 8.0 RC. Thread for reference: https://lists.osgeo.org/pipermail/mapserver-dev/2021-September/016620.html

tomkralidis:
+1 to merge.  Great work here all.

sdlime:
> > Should we merge this and deal with additional tweaks in subsequent pull requests? --Steve
> 
> +1 on merging, it will make it easier to test with the other new features on the main branch. @sdlime - any thoughts on handling the backslash - C++ approach (as in the approach suggested at [#6400 (comment)](https://github.com/MapServer/MapServer/pull/6400#issuecomment-927094508) or a new C function)?

I guess I'd opt for local changes to getTemplateDirectory() as @rouault suggests. I'm not sure how/if we'd need to support Window's paths that can use either forward or back slashes. That was part of the reason I thought it might just be easier to stick with an exception.

> I think the cookie issue will need to be addressed before a 8.0 RC. Thread for reference: https://lists.osgeo.org/pipermail/mapserver-dev/2021-September/016620.html

I think the issue is bigger than just cookies since arbitrary query parameters are quite common when using services. I guess I don't understand the requirement from the OGC perspective and it seems impractical to implement to me. Plus, we're only enforcing it with the one API path which doesn't make sense to me. pygeoapi behaves the same way. For example:

  https://demo.pygeoapi.io/master/collections?f=json&foo=bar

doesn't throw an error on the _foo_. Looking back at the thread and the responses it seems like the cookies functionality is used so removing it isn't an option, not that that really fixes anything. I do think it would be helpful to track the source of parameter values. Perhaps we should create a request parameters object where the members would be name, value and source. That would be easier to share and source values could be get, post or cookie.

--Steve

PR #6410, opened 2021-09-17
Define rect for a dwithin filter condition

gerdos82:
We observed that our WFS (MapServer + GeoPackage utilizing the OGR backend) is a lot slower when the DWithin spatial operator is used in a filter condition. Although the spatial index in the GeoPackage could be used to accelerate this specific operator it was not.

This PR adds a bounding box to the request passed down to OGR when the DWithin spatial operator is used. For the bounding box we use the envelope of the geometry with the distance used as the buffer. Below is an example of the filter:

`<fes:Filter>
<fes:DWithin><fes:ValueReference>begrenzingPerceel</fes:ValueReference>
<gml:Point gml:id="geo.1" srsName="EPSG:28992"><gml:pos dimension="2">257050.19722083333 470137.85737916664</gml:pos></gml:Point><fes:Distance units="m">12.287004477416515</fes:Distance></fes:DWithin>
</fes:Filter>`

### DIFF
--- a/mapogr.cpp
+++ b/mapogr.cpp
@@ -3434,12 +3434,13 @@ static int  msOGRExtractTopSpatialFilter( msOGRFileInfo *info,
                                           pSpatialFilterNode);
   }
 
-  if( (expr->m_nToken == MS_TOKEN_COMPARISON_INTERSECTS ||
+  if( (((expr->m_nToken == MS_TOKEN_COMPARISON_INTERSECTS ||
       expr->m_nToken == MS_TOKEN_COMPARISON_OVERLAPS ||
       expr->m_nToken == MS_TOKEN_COMPARISON_CROSSES ||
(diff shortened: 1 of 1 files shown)


rouault:
Looks good to me. A new test case in msautotest/wxs/wfs_ogr.map  would be ideal

gerdos82:
>  A new test case in msautotest/wxs/wfs_ogr.map would be ideal

I'm going to try to add a test case.

gerdos82:
> A new test case in msautotest/wxs/wfs_ogr.map would be ideal

I see there already is a test case which covers the DWITHIN filter for ogr. It's located in msautotest/wxs/wfs_filter_ogr.map. Sinds this PR is only a performance fix with no impact on functionality, is it still necessary to add a test case?

jmckenna:
@gerdos82 great point, thanks for examining.  You're right.  thanks again for this improvement!

PR #6416, opened 2021-09-23
Invalid geometry from buildmbr when a point is used a base for the rect

gerdos82:
When `wfs_use_default_extent_for_getfeature` is set to `off` on a layer, the geometry in an intersect filter is used to calculate the `rect`. This `rect` is the input for the `BuildMbr` spatialite function. When the geometry in the intersect filter is a POINT, the POLYGON resulting from `BuildMbr` is invalid, because all vertices in the POLYGON are the same. This POLYGON is used in the `Intersect` spatialite function. It seems to depend on the version of spatialite whether or not this is an issue. Some versions of spatialite seem to handle an intersect with an invalid POLYGON fine and some never result in a match. This PR tries to make this more robust and checks if the `rect` is a POINT and in this case uses a POINT as the input for the intersect, in all other cases the `BuildMbr` is used to create a POLYGON.

### DIFF
--- a/msautotest/wxs/wfs_ogr_gpkg.map
+++ b/msautotest/wxs/wfs_ogr_gpkg.map
@@ -11,6 +11,9 @@
 #
 # RUN_PARMS: wfs_ogr_gpkg_issue_6325.xml [MAPSERV] QUERY_STRING="map=[MAPFILE]&SERVICE=WFS&VERSION=2.0.0&REQUEST=GetFeature&TYPENAMES=test_6325&BBOX=0.75,0.75,9,9&COUNT=1" > [RESULT_DEVERSION]
 #
+# Filter using intersect with point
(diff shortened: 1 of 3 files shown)


geographika (approved):

msautotest/wxs/wfs_ogr_gpkg.map
```diff
@@ -95,6 +98,7 @@ LAYER
     "wfs_featureid"     "id"
     "gml_include_items" "all"
     "gml_types"         "auto"
+    "wfs_use_default_extent_for_getfeature" "no"
```
rouault, line 101: is it really needed to reproduce the bug ? I'd anticipate that, be it set or not, the geometry passed to Intersects() would always be degenerate to the point

gerdos82: When I don't set this, the extent is used to calculate the rect instead of the geometry in the intersect filter.

rouault: I would expect them to be intersected at https://github.com/MapServer/MapServer/blob/main/mapogr.cpp#L2245

gerdos82: My knowledge of MapServer is not extensive enough to completely understand the inner workings, but as I understand it, the rect is only set by the geometry from the filter when `msOGRExtractTopSpatialFilter` is called. This method sets `rect_is_defined` to true when calculating a `rect` was possible based on the filter. It looks like `msOGRExtractTopSpatialFilter` when `wfs_use_default_extent_for_getfeature` is `on`.

rouault: hum, ok, anyway your fix is very reasonable.

PR #6417, opened 2021-10-05
Centerline GEOMTRANSFORM Support

sdlime:
Rework of the original  https://github.com/MapServer/MapServer/pull/5854 pull request. Cleaned up a few pieces of the implementation and renamed _skeletonize_ to _centerline_ - it's a much more accurate name.


LarsSchy:
I am missing mapgraph.c and mapgraph.h when I tried to build.

LarsSchy:
I am very happy with my results now.  

(smoothsia(centerline([shape]), 3, 1, 'angle'))

test_250K_TT32

jratike80:
The repeated labels next. Not drawing the first label at the bottom of a bay is a good idea, but what if the requested image bbox cuts the lake at the middle? How could you define that the first label on the centerline would be OK then?

LarsSchy:
I did some tests with REPEATDISTANCE during the weekend.  If the line is long enough for at least two or more labels, then I would like to avoid to first 20 % of the line and the last 20% of the line for labeling to avoid having labels in the bays.   I will do more experiments with different extents and cutting the lakes.

jmckenna:
Centerline well here on Windows.  (I only briefly got caught during compile because this branch requires USE_OGCAPI_SVR )

sdlime:
I'm working on adding some test cases and am seeing some weird behavior on a clipped polygon. The computed centerline is identical regardless of the viewport because the raw, untransformed shape is used as input.  However, drawing the centerline when the whole polygon isn't visible results in an artifact. A couple of images below illustrate the problem. I'm curious if anyone else is seeing this behavior.

The resulting centerline is of the correct shape type and there is no segment connecting the first and last points. I can't believe this is a geometry clipping issue or we'd have seen it ages ago.

_Zoomed in, note the straight line running from the UR corner of the large, clipped polygon back to the starting point._
ttt

_All features completely visible._
ttt-1

sdlime:
Ugh, never mind. Found the issue about 10 minutes after that last post. Existing GEOMTRANSFORM functions preserved the shape type while the centerline transformation goes from polygon to line. Now I get output like so:

ttt-2

Note this is from the msautotest/misc/world_testpoly.shp shapefile.

sdlime:
Still needs work to clean up memory leaks. Working on it... --Steve

sdlime:
> Am I configuring the layer incorrectly?

Configuration looks ok at first glance. That error message occurs when there are no edges added to the directed graph computed from the Voronoi diagram. I've only seen that happen in cases with super simple geometries like a square or triangle with vertices only at the corners. The algorithm removes Densifying solves that problem. What do you get if you just do?

```
GEOMTRANSFORM (centerline([shape]))
```

Any chance you could share your test case?

--Steve

LarsSchy:
My experiments are working fine now with: (smoothsia(centerline([shape]))) !

I looked at the natural earth data from Seth.   When I looked at the clipped data in QGIS, I can see several very small triangles.  Some are in the data set from the beginning and some are due to the clipping. 

ogr2ogr natural_earth_vector_lakes10m_clip.gpkg -clipsrc -68.7202 53.075 -67.7202 54.075 natural_earth_vector.gpkg ne_10m_lakes

sdlime:
> Maybe these errors should be logged and ignored so that centerlines can still be generated for other features in the layer?

I think that's going to be difficult based on the way the parser works since there may be downstream processing (e.g. SmoothSIA) so there's no way to bail mid-expression evaluation with a "never mind" response. I don't know if that's possible or not.

PR #6319, opened 2021-05-04
Fixes for several security vulnerabilities found with libFuzzer

MaxKellermann:

PR #6418, opened 2021-10-05
Fix various security vulnerabilites found by libFuzzer, part 2

MaxKellermann:
This PR starts with a few build breakage fixes which occur when almost all optional features are disabled at compile time.
The last 3 commits fix security vulnerabilities I found with libFuzzer, one of which qualifies for remote code execution, the worst kind of vulnerability.
A crash report from Google Play of an open source app using a copy of MapServer code (https://github.com/XCSoar/XCSoar/) contained a crash due to one of these vulnerabilities, so I investigated and turned on libFuzzer again, which quickly found more vulnerabilities.
I used this program (specific to my app's internal APIs wrapping our my copy of the MapServer code) to run libFuzzer: https://github.com/XCSoar/XCSoar/blob/master/fuzzer/src/FuzzTopographyFile.cpp

This is similar to my previous PR earlier this year: https://github.com/MapServer/MapServer/pull/6319
Back then, @rouault asked me to change the PR to the "main" branch; do you want me to do the same here? (My way of managing my open source apps is that I always merge bug fixes to the "stable" branch first, and then merge all "stable" bug fixes into the "main" branch.)


rouault:
> do you want me to do the same here?

at least, issue also a corresponding PR against master. Note also that 7.6 branch will receive little maintenance at that point

jmckenna:
wow thanks @MaxKellermann.  This will spawn us to have another 7.6..x release with this security fix.  Great find.

mapserv.c
```diff
@@ -323,7 +323,7 @@ int main(int argc, char *argv[])
   }
   msCleanup();
 
-#ifdef _WIN32
+#ifdef __WIN32
```
rouault, line 326:
```suggestion
#ifdef _WIN32
```

rouault:
> This will spawn us to have another 7.6..x release with this security fix. Great find.

as far as I can tell, the vulnerability is if you read from corrupted / hostile shapefiles. It isn't something that is triggered by remote requests, so the severity is much less important

MaxKellermann:
> as far as I can tell, the vulnerability is if you read from corrupted / hostile shapefiles.

Correct.

jmckenna:
Unfortunately shapefiles are still widely used by MapServer users, it's still an important fix and I think should warrant a security release.

MaxKellermann:
> it's still an important fix and I think should warrant a security release.

Agree. These are severe vulnerabilities.
They may not be directly exploitable through the MapServer API, but once you find a way to convince a MapServer to open your shapefile, you own the server.

Just because we can't currently imagine a way to exploit a server remotely doesn't mean that somebody with more creativity can't find a way. Better be too careful than sorry.

The coding style in this project is very fragile towards such vulnerabilities. For example, the code checking buffer sizes looks fragile and obscure; stuff like:
https://github.com/MapServer/MapServer/blob/1c14821bc4c6fc10e966fd05230ad71c05817c7d/mapshape.c#L770-L773
So many magic numbers! Where do they come from? Near that code fragment, I don't see any buffer size verification. Maybe there is verification somewhere else, but it is not obvious at all that this code is correct; quite contrary, you made it extremely difficult for a human to verify this piece of code.

Much of the code deals with 32 bit signed integers for sizes. That's why you always need an additional `<0` check. Just use unsigned integers! And watch out for integer truncation when you cast `size_t` to `int`.

What about integer overflows?
https://github.com/MapServer/MapServer/blob/1c14821bc4c6fc10e966fd05230ad71c05817c7d/mapshape.c#L617-L620
This will easily overflow, leading to a `SfRealloc()` call with a negative integer. Oh yes, `SfRealloc()` takes a signed integer, WTF. What happens with this negative integer when it's casted back to `size_t` for `realloc()`?

Maybe you could argue that this and that limit in the code prevents this integer overflow. Maybe, maybe somebody finds a way around this. My point is: the MapServer code is horribly fragile, it's the anti-style for making secure code. Everywhere I look, I see potential for vulnerabilities.

Of course, it's plain C, and getting error code paths right in plain C is utterly difficult. Just look at the many memory leaks, or worse, the double-free bugs I found (leaks are just DoS, but double-free often leads to remote code execution). C is simply the wrong tool for the job. You could easily improve the code quality by switching all sources to a C++ compiler and use RAII to manage allocations. (Some argue that C++ isn't safe enough and everybody should be using Rust, but I'm a C++ fanboy, and C++ is certainly a great improvement over plain C, and switching is mostly trivial.)

rouault (approved):

sdlime:
@MaxKellermann, your candor is appreciated. Can you elaborate on "mostly trivial"?

MaxKellermann:
> @MaxKellermann, your candor is appreciated. Can you elaborate on "mostly trivial"?

More like commit 098eb92f86a9b3bfe3a8c21e4ba812c9b6fb1664 (by @rouault). That is: rename `*.c` to `*.cpp`, which just switches from the C compiler to the C++ compiler. This usually needs a few adjustments; C++ is mostly a superset of C, but there are small differences.

The whole `int` -> `bool` transition was good, but has nothing to do with the C++ switch (`bool` has been a plain C feature for 22 years already, and using `int` as boolean has been obsolete ever since); this just makes the transition look more complicated than it needs to be. Most of the other changes in 098eb92f86a9b3bfe3a8c21e4ba812c9b6fb1664 were unnecessary, too. I would really just touch the few things that are really different in C/C++ in the very first step.

Second step (optional): small syntactic stuff like `NULL` -> `nullptr`.

Third step (the interesting part): switch to safe C++ data structures, use RAII to free resources automatically, wrap legacy code / thirdparty C libraries in safe C++ wrappers. This eliminates all of the ugly cleanup code like:
https://github.com/MapServer/MapServer/blob/1c14821bc4c6fc10e966fd05230ad71c05817c7d/mapshape.c#L1354-L1364
... which is very fragile, because it's easy to forget something, and if you add a new struct field, you have to adjust **all** error code paths.

Fourth step: migrate the error handling to C++ exceptions.

MaxKellermann:
Added a backport of 8dc41040c3 and a fix for https://github.com/MapServer/MapServer/pull/6421#issuecomment-934520235

PR #6419, opened 2021-10-05
Fix various security vulnerabilites found by libFuzzer, part 2b

MaxKellermann:
This is https://github.com/MapServer/MapServer/pull/6418 but for the "main" branch.
The buffer overflow fix https://github.com/MapServer/MapServer/pull/6418/commits/86b364591ae95ae8b14b4ce0f1829fa42432a365 was meanwhile mitigated a bit by fcf13d849cb, but that commit still left a hole open.
(@rouault)


MaxKellermann:
The Travis failure is not related to this PR:
> error: can't find Rust compiler

geographika:
Thanks @MaxKellermann! I've hopefully added a fix for the (unrelated) rust/cryptography error.

mapogcfiltercommon.cpp
```diff
@@ -530,6 +529,7 @@ static std::string FLTGetFeatureIdCommonExpression(FilterEncodingNode *psFilterN
           if( pszDot )
             pszId = pszDot + 1;
 
+          int bString=0;
```
rouault, line 532:
```suggestion
          bool bString = false;
```

mapserv.c
```diff
@@ -314,7 +314,7 @@ int main(int argc, char *argv[])
   }
   msCleanup();
 
-#ifdef _WIN32
+#ifdef __WIN32
```
rouault, line 317:
```suggestion
#ifdef _WIN32
```

MaxKellermann:
Fixed the two merge mistakes, thanks @rouault

rouault (approved):

MaxKellermann:
Added a fix for this bug: https://github.com/MapServer/MapServer/pull/6421#issuecomment-934520235

Issue #6415, opened 2021-09-23
msautotests failing on WMS client tests


geographika:
See some of the last few pull requests e.g. https://github.com/MapServer/MapServer/runs/3680893365#step:3:6119
This was a change to the Windows build so shouldn't have affected the Linux testing. 

Could this be related to requiring a config file to allow certain map paths when connecting to https://demo.mapserver.org/cgi-bin/wms?

PR #6420, opened 2021-10-05
Pin cryptography version

geographika:
The previous fix `ENV CRYPTOGRAPHY_DONT_BUILD_RUST=1` only works up to 3.4 so pin the version to avoid `error: can't find Rust compiler` (e.g. [here](https://app.travis-ci.com/github/MapServer/MapServer/jobs/541345228#L1801)

See https://stackoverflow.com/questions/66118337/how-to-get-rid-of-cryptography-build-error and also https://github.com/pyca/cryptography/issues/5771

### DIFF
--- a/.github/workflows/start.sh
+++ b/.github/workflows/start.sh
@@ -29,7 +29,6 @@ eval "$(pyenv virtualenv-init -)"
 ln -s /usr/bin/python3 /usr/bin/python
 ln -s /usr/bin/pip3 /usr/bin/pip
 
-export CRYPTOGRAPHY_DONT_BUILD_RUST=1 # to avoid issue when building Cryptography python module
 pip install --upgrade pip
 
 # Install recent cmake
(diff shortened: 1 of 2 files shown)


geographika:
I believe this fixes the rust issue, but the WMS client tests are still flakey (see https://github.com/MapServer/MapServer/issues/6415).

rouault:
> but the WMS client tests are still flakey (see #6415).

what's weird is that it doesn't fail on the Build github action workflow based config. could you try to add the "ca-certificates" package at the end of line 12 of ci/travis/before_install.sh . I've had an issue recently with GDAL on Travis-CI when hitting OSGeo based servers due to an outdated ca-certifcates package (there have been changes related to Let's encrypt recently in OSGeo infrastructure)

geographika:
Thanks @rouault for the suggestion. That seems to have fixed it (although I thought I'd seen Travis pass since I opened #6415). 
Merging this now so the CI passes again.

PR #6424, opened 2021-10-15
Make sure densify tolerance is greater than zero.

sdlime:
Make sure densify tolerance is greater than zero and a few misc improvements to functions supporting the inner and outer GEOMTRANSFORMs. Mostly checking input values are valid...

### DIFF
--- a/mapprimitive.c
+++ b/mapprimitive.c
@@ -516,10 +516,10 @@ void msClipPolylineRect(shapeObj *shape, rectObj rect)
   double x1, x2, y1, y2;
   shapeObj tmp;
 
-  memset( &tmp, 0, sizeof(shapeObj) );
-
-  if(shape->numlines == 0) /* nothing to clip */
+  if(!shape || shape->numlines == 0) /* nothing to clip */
     return;
+
+  memset( &tmp, 0, sizeof(shapeObj) );
 
   /*
   ** Don't do any clip processing of shapes completely within the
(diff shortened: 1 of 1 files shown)


sdlime:
Should be a quick review. I just wanted to make sure the densify tolerance was > 0. I touched a couple of other functions related to inner/outer rings to check args and make sure they were reasonable. Many of the functions in mapprimitive.c don't check args - think it's worth going through and hardening?

sdlime:
> > Many of the functions in mapprimitive.c don't check args - think it's worth going through and hardening?
> 
> It really depends on situations, and whether they are in time critical paths, and the cost of the checks versus the cost of what the function does. For example for a simple getter that might be called in a loop, doing argument verification can become expensive. In the C or C++ standard libraries, there are (documented) assumptions on the arguments, that aren't enforced by them. For example strlen(NULL) is documented as unspecified behavior, or v[idx] on a std::vector if you pass an invalid index.

There are a few functions like that, for example, msAddPoint(), msAddLine(), msAddLineDirectly() that would be called many times in loops as geometries are manipulated. Still others like geometry clipping or area computation are used relatively infrequently and do a lot of work so simply checking to make sure the input shape isn't NULL seems insignificant.

When you mention performance I wonder if there are approaches to regression tests that could catch changes that significantly affect certain performance benchmarks. I suppose there's probably too much variation in the environments where tests are executed for that to be possible.

PR #6426, opened 2021-10-21
WIP: Require users to opt-in for URL-based update of select properties

sdlime:
This limits URL-based update of select properties by adding a new "urlupdate" property to objects that support that sort of thing currently. Users will have to opt-in and have fine-grained control of which properties to allow update. This support has been added to the scalebar object to serve as a pattern for review.

- [ ] objects to update: map, layer, class, style, label, legend, query map, reference map, ~~scalebar~~ and web.
- [ ] remove support for properties that support runtime substitutions. Those properties require a regex which can be tough to write for complex strings so using substitutions is *much* easier (and safer) for users.
- [ ] consider removing properties that can be set via other mechanisms (e.g. map size or map extent).
- [ ] remove DATAPATTERN and TEMPLATEPATTERN properties from the web object.

--Steve


sdlime:
The more time I spend with this topic the more I think we should consider a different path. For example, I don't think a new parameter hung off a bunch of objects with a cryptic syntax helps anyone. I mean, why go through all this just to change units on a freaking scalebar - just create a second mapfile if that's what is really needed and be done with it.

When @tbonfort added default value support for runtime substitutions I think it rendered the need for inline mapfile snippets largely mute. I think we'd be **much** better off simplifying things by removing this support completely and dealing with specific use cases in other ways moving forward. For example, for a couple of high-value parameters like image format or resolution could be added separately without all the overhead of the current approach and we could expand runsubs to a few other parameters such as layer masks. It would simplify mapfile processing, it would simplify the lexer and would improve security overall. 

--Steve

geographika:
Dev mailing list discussion for reference: https://lists.osgeo.org/pipermail/mapserver-dev/2021-October/016647.html

@sdlime - so would the plan be to drop support for https://mapserver.org/cgi/controls.html#changing-map-file-parameters-via-a-form-or-a-url in the v8.0 release?

I think I've used this syntax once, and then decided it was easier to generate multiple mapfiles as you described. Might be worth a post to the user list to see if it is still widely used?

sdlime:
@geographika, yes on completely dropping for 8.0 - kinda now or never IMHO. I can send a quick email to mapserver-users on the topic.

jmckenna:
I've add this (removing ability to change mapfile params through url) to the 8.0 release plan: https://github.com/MapServer/MapServer/wiki/MapServer-8.0-Release-Plan

sdlime:
Going a different direction, closing... --Steve

PR #6431, opened 2021-11-07
msBuildWFSLayerGetURL: added WFS 1.1.0 in error message

richie21:
Added version WFS 1.1.0 supported by MapServer as a client (in the error message).


rouault:
unrelated test failures

jmckenna:
thanks @richie21 great catch.

geographika:
Doesn't `strncmp(pszVersion, "1.1", 3) != 0)` mean `1.1.X` would work as well as `1.1.0`. Were there other versions of WFS 1.1?

rouault:
> Were there other versions of WFS 1.1?

the corrigendum http://docs.opengeospatial.org/is/04-094r1/04-094r1.html is advertized as 1.1.3, and this can be used in requests

PR #6434, opened 2021-11-12
Properly handle datetime fields

copierrj:
We encountered incorrect date/time values in WFS GetFeature responses of our (GeoPackage backed) MapServer instances. Only the year was rendered correctly. All other parts where zeroed out. 

While investigating the issue it was discovered that the code that handles date/time values (`msParseTime` in `maptime.c`, called by `msGMLWriteItem` in `mapgml.c`) expects the provided date/time strings to be formatted in ISO 8601. Unfortunately the OGR backend (`mapogr.cpp`) provides date/time strings in a different format (yyyy/mm/dd hh:mm:ss) and `msParseTime` only succeeds in parsing the year part of the date/time string and silently ignores the rest. Resulting in malformed GML output.

It was discovered that the `msOGRGetValues` function treats every field as a string field, leaving it up to OGR how to convert any non-string field into a string. After changing the code to use `OGR_F_GetFieldAsDateTime` for date/time fields and properly format it as an ISO date/time string the problem was resolved.

I don't know if this is actually the right approach to fix this issue. But if I understand the OGR API documentation correctly, `OGR_F_GetFieldAsString` is only supposed to be called for real, integer and string fields.

### DIFF
--- a/mapogr.cpp
+++ b/mapogr.cpp
@@ -576,10 +576,38 @@ static char **msOGRGetValues(layerObj *layer, OGRFeatureH hFeature)
 
   int *itemindexes = (int*)layer->iteminfo;
 
+  int nYear;
+  int nMonth;
+  int nDay;
+  int nHour;
+  int nMinute;
+  int nSecond;
+  int nTZFlag;
+
   for(i=0; i<layer->numitems; i++) {
     if (itemindexes[i] >= 0) {
       // Extract regular attributes
-      values[i] = msStrdup(OGR_F_GetFieldAsString( hFeature, itemindexes[i]));
+      OGRFieldDefnH hFieldDefnRef = OGR_F_GetFieldDefnRef(hFeature, itemindexes[i]);
+      switch(OGR_Fld_GetType(hFieldDefnRef)) {
+      case OFTTime:
+      case OFTDate:
+      case OFTDateTime:
+          OGR_F_GetFieldAsDateTime(hFeature, itemindexes[i], &nYear, &nMonth, &nDay, &nHour, &nMinute, &nSecond, &nTZFlag);
+          switch(nTZFlag) {
+          case 0:
+          case 1:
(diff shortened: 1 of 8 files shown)


jmckenna:
@copierrj can you also: 
 - update this pull request to latest main
 - add a test for your case, into msautotest (see https://mapserver.org/development/tests/autotest.html)
   - there is an OGR-GPKG-GetFeature test in `/msautotest/wxs/wfs_ogr_gpkg.map` that might be useful

copierrj:
@jmckenna I'll put it on my TODO list

jmckenna:
@copierrj thanks

mapogr.cpp
```diff
@@ -576,10 +576,38 @@ static char **msOGRGetValues(layerObj *layer, OGRFeatureH hFeature)
...
+          switch(nTZFlag) {
+          case 0:
+          case 1:
+            values[i] = msStrdup(CPLSPrintf("%04d-%02d-%02dT%02d:%02d:%02d", nYear, nMonth, nDay, nHour, nMinute, nSecond));
+            break;
+          case 100:
+            values[i] = msStrdup(CPLSPrintf("%04d-%02d-%02dT%02d:%02d:%02dZ", nYear, nMonth, nDay, nHour, nMinute, nSecond));
+            break;
+          default:
+            values[i] = msStrdup(CPLSPrintf("%04d-%02d-%02dT%02d:%02d:%02d%+02d:00", nYear, nMonth, nDay, nHour, nMinute, nSecond, nTZFlag));
```
rouault, line 605: the conversion of TZFlag to a time zone is incorrect. You can borrow logic/code from https://github.com/OSGeo/gdal/blob/master/ogr/ogrutils.cpp#L1358

rouault:
@copierrj I've a followup commit that you should merge from https://github.com/PDOK/mapserver/pull/7

copierrj:
@rouault thanks!

copierrj:
I've retested the changes for our particular use case and everything seems to work just fine. Too bad I overlooked the incorrect null handling first time around.

@jmckenna are these changes to the PR sufficient?

PR #6440, opened 2021-11-21
msautotest: rename test file to wfsogr10_shapezip.zip.with_headers

rouault:


rouault:
Travis ASAN job failure unrelated and addressed per https://github.com/MapServer/MapServer/pull/6442

PR #6445, opened 2021-12-02
6434 fix datetime

darkblue-b:
just code, per evenR from gdal/ogrutils.cpp


rouault:
Closing that one since https://github.com/MapServer/MapServer/pull/6434 is in a better shape

PR #6446, opened 2021-12-14
Extending PropertyIsNull with GPKG support

WouterVisscher:
We are migrating several of our backend from PostGIS to GeoPackage (why have large DB's for serving just a couple MB's of data). 
With that migration we noticed that WFS requests with a FILTER containing `<PropertyIsNull>` stopt working.
What I can gather from pull https://github.com/MapServer/MapServer/pull/5354 is that this is a "thing" 😄 

But I also noticed that this works for not only PostGIS, but also SpatiaLite. So my think was SpatiaLite is a extention for SQLite, GeoPackage is based on SQLite so,... could it be this simple... 

Well I do get working PropertyIsNull Filters with a GeoPackage backend now... so I guess it is?
If I missing something please let me know.


geographika:
Thanks @WouterVisscher - were you able to compile and confirm it was fixed?
It would be good to add a GPKG test similar to the [SQLite one](https://github.com/MapServer/MapServer/pull/5354/files#diff-22034a5df52bd85f399c57bc4344af5cca7b8fdc84fa38db9cadf64232962770)

The failing Appveyor build is due to Python 3.10 being picked up - I'll look at this, it is unrelated to your change.

WouterVisscher:
>  were you able to compile and confirm it was fixed?
Yes, I've got it working like intended

> add a GPKG test similar
Will do

rouault (changes requested):
Please rename msOGRIsSpatialite to msOGRSupportsIsNull

jmckenna:
thanks @WouterVisscher, also for adding then GPKG test.

PR #6452, opened 2021-12-28
msProjectRect(): fix reprojecting from polar stereographic to another projected system

rouault:


jmckenna:
Excellent change, thanks @rouault !

PR #6456, opened 2022-01-21
netCDF output related improvements for WCS 

rouault:
*   WCS: support providing metadata items to GDAL output
    
    If a Mapfile defines a outputformat that is enabled in a WCS
    layer, then it is possible to define a per-layer value for GDAL metadata items,
    both at the dataset level with a syntax like:
    
        "wcs_outputformat_{format_name}_mdi_{domain}_{key}" "value"
    
    where:
    - format_name is substituted with the name of the output format
    - domain is substituted with the GDAL metadata domain into which to
      write the item. The special value "default" can be used for the
      default domain
    - key is the key of the metadata item.
    
    For band level metadata items, the syntax is:
    
        "wcs_outputformat_{format_name}_mdi_BAND_{band_number}_{domain}_{key}" "value"
    
    where band_number is a number between 1 and the number of bands of the
    dataset of the layer.
    In case of band selection in WCS output, the code will automatically renumber
    the band number.
    
    netCDF output is a typical use case. For dataset level metadata, the key
    must be prefix with NC_GLOBAL#
    
    e.g.
    
      "wcs_outputformat_netCDF_mdi_default_NC_GLOBAL#title" "A succinct description of what is in the dataset"
    
    For Band metadata, e.g.
    
      "wcs_outputformat_netCDF_mdi_BAND_1_default_NETCDF_VARNAME" "wind"
      "wcs_outputformat_netCDF_mdi_BAND_1_default_long_name"      "wind from direction"
      "wcs_outputformat_netCDF_mdi_BAND_1_default_standard_name"  "wind_from_direction"
      "wcs_outputformat_netCDF_mdi_BAND_1_default_units"          "degree"
    
    NETCDF_VARNAME is a special key name recognized by the GDAL netCDF
    driver to determine the variable name. All other keys are mapped to a
    netCDF attribute belonging to the corresponding variable.

*  WCS and netCDF input/output: automatically set GDAL/netCDF output metadata itemps from netCDF  input dataset metadata (if wcs_outputformat_netCDF_mdi_ not defined)


jmckenna:
thanks @rouault nice enhancement!

PR #6458, opened 2022-01-24
Improved numerical validation.

sdlime:
Implements a common approach to validating numerical values read from the mapfile at parse time. Meant to complement runtime validation done throughout the code. Review and additions most welcome.


mapfile.c
```diff
@@ -5414,8 +5449,10 @@ void initQueryMap(queryMapObj *querymap)
   MS_INIT_COLOR(querymap->color, 255,255,0,255); /* yellow */
 }
 
-int loadQueryMap(queryMapObj *querymap)
+int loadQueryMap(queryMapObj *querymap, mapObj *map)
 {
+  querymap->map = (mapObj *)map;
```
geographika, line 5454: Could there be any issues with adding the map as a property to the querymap? Would it be safer to pass in the `maxsize` and `minsize` values directly to this function?

mapfile.c
```diff
@@ -6528,7 +6561,7 @@ int msUpdateMapFromURL(mapObj *map, char *variable, char *string)
           msyystring = string;
           msyylex();
 
-          if(getDouble(&(rotation_angle)) == -1) break;
+          if(getDouble(&(rotation_angle), MS_NUM_CHECK_RANGE, -360, 360) == -1) break;
```
geographika, line 6564: As above, I was unaware you could use negative angles (to presumably set anti-clockwise rotation?)

mapserver.h
```diff
@@ -937,6 +939,13 @@ The :ref:`QUERYMAP <querymap>` object.
 Instances of querymapObj are always are always embedded inside the :class:`mapObj`.
 */
   typedef struct {
+#ifdef SWIG
+    %immutable;
+#endif /* SWIG */
+    struct mapObj *map; ///< Reference to parent :class:`mapObj`
```
geographika, line 945: May not be required if only used to get the min and max size when loading the map (see comment above).

mapserver.h
```diff
@@ -1487,6 +1496,17 @@ The :ref:`REFERENCE <reference>` object
   /*                             scalebarObj                              */
   /************************************************************************/
 
+  #define MS_SCALEBAR_INTERVALS_MIN 1
+  #define MS_SCALEBAR_INTERVALS_MAX 100
+
+  #define MS_SCALEBAR_WIDTH_MIN 5
```
geographika, line 1502: Docs to udpate at https://mapserver.org/mapfile/scalebar.html#mapfile-scalebar-size

mapserver.h
```diff
@@ -1517,6 +1537,11 @@ The :ref:`REFERENCE <reference>` object
   /*                              legendObj                               */
   /************************************************************************/
 
+  #define MS_LEGEND_KEYSIZE_MIN 5
```
geographika, line 1540: Docs to update at https://mapserver.org/mapfile/legend.html

geographika:
@sdlime - looks good to me, and any validation that can be done at the Mapfile parsing stage will save a lot of pain later on. 

I ran on a couple of large Mapfiles I have here, and hit the error message `loadStyle(): General error message. Invalid WIDTH, must be greater than 0`.  On further investigation this was correctly thrown as I had a `STYLE` with a `WIDTH 0` which serves no purpose. 

```
CLASS
..
            STYLE
                COLOR 0 0 0
                WIDTH 0
            END
            STYLE
                COLOR 255 192 203
                WIDTH 4
            END
        END
END
```

Do you expect parsing speeds to be impacted in any way? I can try and test that here to see. 

Finally I can look at making a docs pull request with the min/max values for various keywords if that helps.

mapfile.c
```diff
@@ -1733,21 +1765,19 @@ static int loadLabel(labelObj *label)
         if(loadColor(&(label->shadowcolor), NULL) != MS_SUCCESS) return(-1);
         break;
       case(SHADOWSIZE):
-        /* if(getInteger(&(label->shadowsizex)) == -1) return(-1); */
         if((symbol = getSymbol(2, MS_NUMBER,MS_BINDING)) == -1) return(-1);
         if(symbol == MS_NUMBER) {
-          label->shadowsizex = (int) msyynumber;
+          label->shadowsizex = (int) msyynumber; // TODO > 0
```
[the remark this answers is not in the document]
sdlime: No, this is an oversight on my part - forgot to resolve the "TODO". Will update...

mapfile.c
```diff
@@ -4751,31 +4785,32 @@ int loadReferenceMap(referenceMapObj *ref, mapObj *map)
         if(loadColor(&(ref->outlinecolor), NULL) != MS_SUCCESS) return(-1);
         break;
       case(SIZE):
-        if(getInteger(&(ref->width)) == -1) return(-1);
-        if(getInteger(&(ref->height)) == -1) return(-1);
+        if(getInteger(&(ref->width), MS_NUM_CHECK_RANGE, 5, ref->map->maxsize) == -1) return(-1); // is 5 reasonable?
```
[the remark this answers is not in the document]
sdlime: Yes. Once there is agreement that that's the appropriate value. It seemed reasonable to me...

mapfile.c
```diff
@@ -6176,7 +6213,7 @@ static int loadMapInternal(mapObj *map)
       break;
       case(ANGLE): {
         double rotation_angle;
-        if(getDouble(&(rotation_angle)) == -1) return MS_FAILURE;
+        if(getDouble(&(rotation_angle), MS_NUM_CHECK_RANGE, -360, 360) == -1) return MS_FAILURE;
```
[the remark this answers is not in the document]
sdlime: I believe so - I mean it makes sense to me. @tbonfort, what do you think?

mapfile.c
sdlime: We use the same approach with other objects (e.g. layer, scalebar) where there is an opaque pointer back to the parent so this should be ok. I was trying to avoid changing function signatures too much.

mapfile.c
sdlime: Would be good to get confirmation from @tbonfort but this should be ok and should be easy to test.

mapserver.h
sdlime: Yes, once values are confirmed.

mapserver.h
sdlime: Yes, once values are confirmed.

sdlime:
@geographika, I'm not anticipating any significant impact on parsing speed. If you have the opportunity to validate then that would be great. In the big scheme of things additions like the config file will have a more significant affect - especially for straight CGI execution.

In terms of things like _WIDTH 0_, I wrestled with handling default values when I worked through this. I did do some testing when writing mapfiles via MapScript to make sure that default values weren't being written routinely. I mean, we wouldn't want MapServer to barf reading a mapfile that was written by MapServer. That doesn't seem to be the case here and we can recommend commenting things out rather than using a value like 0.

That said, we could conditionally allow default values in a few cases if something like _WIDTH 0_ is pretty commonplace - recommended in the docs or whatever.

--Steve

mapfile.c
```diff
@@ -424,9 +452,13 @@ int loadColor(colorObj *color, attributeBindingObj *binding)
 
   color->alpha=255;
   if(symbol == MS_NUMBER) {
-    color->red = (int) msyynumber;
-    if(getInteger(&(color->green)) == -1) return MS_FAILURE;
-    if(getInteger(&(color->blue)) == -1) return MS_FAILURE;
+    if(msyynumber >= -255 && msyynumber <= 255) {
```
rouault, line 455: why do we accept negative values for colors ?

sdlime: It's used to trigger alternate behavior in at least one case. https://mapserver.org/mapfile/reference.html says that a negative color component is used to draw an outline and no fill. We could add a flag to getColor() to allow negative color components and only set that to true for the reference color. We do initialize colors to -1,-1,-1 in a number of spots.

rouault: putting the above as a comment in the code would be helpful :-)

sdlime: You mean part of my explanation?

rouault: yes, it is not obvious when reading the code why a negative value for a color is accepted

sdlime: added comment...

sdlime:
I think I got things resolved (except for docs)... Please let me know about merging. --Steve

rouault (approved):

sdlime:
This will be a candidate for backport to 7.6 but I think we should wait until some 8.0 testing happens.

geographika:
@sdlime - do you want me to update the docs? I'm also going to update mappyfile to include the min/max values so can do both at the same time.

sdlime:
> @sdlime - do you want me to update the docs? I'm also going to update mappyfile to include the min/max values so can do both at the same time.

That would be great!

jmckenna:
re: backport: I'd also prefer getting feedback through main branch first (will do tests with MS4W today)

re: docs: I'd like to see us getting back to filing an associated ticket in the documentation repository, for each new feature: the benefit is that the docs team will test each new feature, while adding the docs.  This also is a way to keep track of what is missing in the docs.

ejn:
As a user I'd say that this change (much as I find it a good thing (TM)) shouldn't be backported - it's a change that could break existing installations. Yes, those are installations with "invalid" mapfiles, but if they weren't sufficiently invalid that they couldn't be used then it would be rude to introduce what is effectively a breaking change in a minor (or even bugfix) release.

jmckenna:
I agree with you @ejn

jmckenna:
@sdlime something bizarre is happening locally for me, sorry if this sounds like a crazy question: could the output of map2img's `-map_debug` switch be affected by these changes?   `-all_debug` works fine.  `-map_debug` generates the map image but does not output any debug messages (just a blank line).   This works with a build from a month ago.  Something is different, but I don't know what yet.  (either in my build, or a change in the code)

sdlime:
No, the various debug settings are unaffected. This is actually an oversight since if debug is given as a number then it should be checked as well but the code in mapfile.c uses getSymbol() so I missed it. I can patch that...

Anyway, map2img is setting map->debug directly, apart from what's in the mapfile so these changes have no bearing on that.

jmckenna:
Ok thanks, makes sense.  I'll take a deeper look and see if it was something on my end (likely) first.

jmckenna:
@sdlime solved, it was on my end.  Will continue testing this...

jmckenna:
@steve can something be done to include a line number in the error message?  For example, a mapfile with 30 entries of `WIDTH 0` returns only the following with map2img:
```
loadStyle(): General error message. Invalid WIDTH, must be greater than 0
loadStyle(): General error message. Invalid WIDTH, must be greater than 0 <br>
```
Or is this a huge can of worms, for another day?

sdlime:
> @Steve can something be done to include a line number in the error message? For example, a mapfile with 30 entries of `WIDTH 0` returns only the following with map2img:
> 
> ```
> loadStyle(): General error message. Invalid WIDTH, must be greater than 0
> loadStyle(): General error message. Invalid WIDTH, must be greater than 0 <br>
> ```
> 
> Or is this a huge can of worms, for another day?

No, this can be addressed in the error handling. That said, we could also test for >= 0 if width=0 is a common setting (the default width is 1). @geographika has a ticket #6463 that is documenting some exceptions we'll need to address. Perhaps you could add this there so it's all in one spot.

mapserver.h
```diff
@@ -1517,6 +1537,11 @@ The :ref:`REFERENCE <reference>` object
...
+  #define MS_LEGEND_KEYSIZE_MAX 200
```
gu-LGB, line 1541: I'm currently migrating from MS 7.6 to 8.0 and I have a problem with this limitation: So far I could provide legends consisting of a PNG like this:

```
LEGEND
   KEYSIZE 216 216
   KEYSPACING 7 7
   STATUS ON
END

CLASS
  KEYIMAGE "./inspire_lbgr/symbol/legende-sorption_matrix.png"
  NAME " "
END
```
Is there anything against raising the limit to 1000?

PR #6460, opened 2022-01-24
Fix compilation warning in shptreevis

geographika:
Use correct size to avoid the following:

` warning C4267: '=': conversion from 'size_t' to 'int', possible loss of data`


jmckenna:
thanks @geographika !

PR #6462, opened 2022-01-26
Remove unused LABEL>MINLENGTH property

geographika:
As mentioned in #6217 the `LABEL>MINLENGTH` property was added in #2383, but is unused (and undocumented).
I propose this is removed from the codebase (it can be added back in if required).


sdlime:
Will need to update maplexer.l and regenerate maplexer.c.

sdlime:
IMHO it's ok to merge... --Steve

Issue #6463, opened 2022-01-26
New Mapfile Validation Rules


geographika:
Following on from #6458 and the docs pull request [here](https://github.com/MapServer/MapServer-documentation/pull/663), there are a few validation rules to check/confirm:

1. Double check that the `ANGLE` keywords can have negative values. For example the [STYLE ANGLE](https://mapserver.org/mapfile/style.html#mapfile-style-angleitem) docs currently state clockwise angles should be calculated as positive values:

>     Angle, given in degrees, to rotate the symbol (counter clockwise).
>     Default is 0 (no rotation).  If you have an attribute that
>     specifies angles in a clockwise direction (compass direction), you
>     have to adjust the angle attribute values before they reach
>     MapServer  (360-ANGLE), as it is not possible to use a mathematical
>     expression for `ANGLE`.

2. There are several cases where the default values for objects would not pass validation e.g. STYLE [MINSIZE](https://mapserver.org/mapfile/style.html#mapfile-style-minsize), MINWIDTH, OUTLINEWIDTH. Docs would read as *Default is 0. Must be greater than 0.*
Also CLUSTER BUFFER (defaults to 0 but must be > 0). Same for LAYER TOLERANCE. 

3. Perhaps more problematic are when default values are written out to Mapfiles by MapServer, which would then produce invalid maps. E.g in `writeLayer` all the following default to -1, but then validation is set to > 0. 

```c
  writeNumber(stream, indent, "MINGEOWIDTH", -1, layer->mingeowidth);
  writeNumber(stream, indent, "MINSCALEDENOM", -1, layer->minscaledenom);
  writeNumber(stream, indent, "MINFEATURESIZE", -1, layer->minfeaturesize);
```

4. [STYLE SYMBOL](https://mapserver.org/mapfile/style.html#mapfile-style-symbol) docs state the index starts at 1, but validation allows 0. 

5. (nitpick) Minor inconsistency in validation for same property names - MAP MAXSIZE is >= 1 and LABEL MAXSIZE is > 0 (results in same validation as these are both integer values.

sdlime:
For 3... 

```
writeNumber(stream, indent, "MINGEOWIDTH", -1, layer->mingeowidth);
``` 

means that if layer->mingeowidth == -1 then don't write that property so that default value should never be output. We do that to avoid bloated output filled with useless settings. Are you seeing something different?

sdlime:
[...]

PR #6466, opened 2022-01-31
Addressed issues identified in issue #6463.

sdlime:
Addressed issues identified by @geographika and @jmckenna. Note I left style->angle check at 0 to 360.

### DIFF
--- a/mapdraw.c
+++ b/mapdraw.c
@@ -282,8 +282,8 @@ imageObj *msDrawMap(mapObj *map, int querymap)
   if(map->debug >= MS_DEBUGLEVEL_TUNING) msGettimeofday(&mapstarttime, NULL);
 
   if(querymap) { /* use queryMapObj image dimensions */
-    if(map->querymap.width != -1) map->width = map->querymap.width;
-    if(map->querymap.height != -1) map->height = map->querymap.height;
+    if(map->querymap.width > 0 && map->querymap.width <= map->maxsize) map->width = map->querymap.width;
+    if(map->querymap.height > 0 && map->querymap.height <= map->maxsize) map->height = map->querymap.height;
   }
 
   msApplyMapConfigOptions(map);
--- a/mapfile.c
+++ b/mapfile.c
@@ -1669,20 +1669,20 @@ static int loadLabel(labelObj *label)
             label->force = MS_LABEL_FORCE_GROUP;
             break;
           default:
-            msSetError(MS_MISCERR, "Invalid FORCE, must be ON,OFF,or GROUP" , "loadLabel()");
+            msSetError(MS_MISCERR, "Invalid FORCE, must be ON,OFF,or GROUP (line %d)" , "loadLabel()", msyylineno);
             return(-1);
         }
         break;
       case(LABEL):
         break; /* for string loads */
       case(LEADER):
-        msSetError(MS_MISCERR, "LABEL LEADER not implemented. LEADER goes at the CLASS level." , "loadLabel()");
+        msSetError(MS_MISCERR, "LABEL LEADER not implemented. LEADER goes at the CLASS level (line %d)" , "loadLabel()", msyylineno);
         return(-1);
       case(MAXSIZE):
         if(getInteger(&(label->maxsize), MS_NUM_CHECK_GT, 0, -1) == -1) return(-1);
         break;
       case(MAXSCALEDENOM):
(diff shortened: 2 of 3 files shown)


geographika (approved):
Looks good to me, and fine to leave the ANGLE validation as-is. I'll update https://github.com/MapServer/MapServer-documentation/pull/663 with these changes.

sdlime:
Thanks @geographika. Will give @jmckenna a little while to test before merging...

jmckenna:
Thanks @sdlime I am testing now...

jmckenna (commented):
@sdlime I tested the issues that I reported earlier, here's the feedback:

6. line number in error messages --> works well.  For example (not sure why map2img repeats it twice though, but that's very minor) :
```
loadStyle(): General error message. Invalid WIDTH, must be greater than or equal to 0 (line 171)
loadStyle(): General error message. Invalid WIDTH, must be greater than or equal to 0 (line 171) <br>
```
7. test for WIDTH >=0     <-- works well

8. MapScript's save() method for mapObj --> now saves the mapfile with a valid QUERYMAP, such as:
```
  QUERYMAP
    STATUS OFF
    STYLE HILITE
  END # QUERYMAP
```
*new*

9. a valid `MINSCALEDENOM 0` in a LAYER fails, such as:
```
   LAYER
     ...
     MINSCALEDENOM 0
     MAXSCALEDENOM 2539
     ...
   END # layer
```
with error message:
```
getDouble(): Symbol definition error. Parsing error near (0):(line 9281)
getDouble(): Symbol definition error. Parsing error near (0):(line 9281) <br>
```

jmckenna (approved):

sdlime:
@jmckenna, why should MINSCALEDENOM 0 be considered valid?

jmckenna:
> @jmckenna, why should MINSCALEDENOM 0 be considered valid?

This is the case for all mapfiles generated by the "Basemaps" sub-project of MapServer.

sdlime:
> > @jmckenna, why should MINSCALEDENOM 0 be considered valid?
> 
> This is the case for all mapfiles generated by the "Basemaps" sub-project of MapServer (which are ~10,000 lines long), and, with a user hat on, this makes sense to me. Also, I'm for less-breaking of existing mapfiles, especially huge ones like this that we manage under our umbrella.

Ok, I get it. The min/maxscaledenom values are used to limit processing and aren't used in computations so allowing GTE 0 is ok. Should that be the consistent for web, layer, class, style and label objects? @geographika, what do you think? Easy change on my end and now's the time.

sdlime:
Re: 8. @jmckenna, can you check a QUERYMAP with SIZE -1 -1 - so written by an older version? That should be ok as well based on my testing.

jmckenna:
> 
> Ok, I get it. The min/maxscaledenom values are used to limit processing and aren't used in computations so allowing GTE 0 is ok. Should that be the consistent for web, layer, class, style and label objects? @geographika, what do you think? Easy change on my end and now's the time.

Yes I believe this should be consistent for the other scaledenoms in objects.

jmckenna:
> Re: 8. @jmckenna, can you check a QUERYMAP with SIZE -1 -1 - so written by an older version? That should be ok as well based on my testing.

True, thanks, confirmed here that QUERYMAP SIZE -1 -1 is now accepted.

sdlime:
> > Ok, I get it. The min/maxscaledenom values are used to limit processing and aren't used in computations so allowing GTE 0 is ok. Should that be the consistent for web, layer, class, style and label objects? @geographika, what do you think? Easy change on my end and now's the time.
> 
> Yes I believe this should be consistent for the other scaledenoms in objects.

Changing to allow 0 or higher for all minx/maxscaledenom values then.

jmckenna:
> Changing to allow 0 or higher for all minx/maxscaledenom values then.

Yes I believe so.   I think it makes sense to a user.  (staying away from technical debates though ha)

geographika:
> Ok, I get it. The min/maxscaledenom values are used to limit processing and aren't used in computations so allowing GTE 0 is ok. Should that be the consistent for web, layer, class, style and label objects? @geographika, what do you think? Easy change on my end and now's the time

It looks like it is used in [a few Mapfiles in "the wild"](https://github.com/search?q=%22MINSCALEDENOM+0%22++extension%3Amap&type=Code&ref=advsearch&l=&l=) so seems a good idea to allow it. Consistency across the Mapfile objects sounds good.

geographika:
@sdlime - docs updated in https://github.com/MapServer/MapServer-documentation/pull/663 to reflect the changes here

PR #6468, opened 2022-02-01
Add negative ANGLE tests

geographika:
This msautotest file is identical to [symbolrot.map](https://github.com/MapServer/MapServer/blob/main/msautotest/renderers/symbolrot.map) but uses negative values. The images produced look identical (need to confirm there are no floating point issues that may require a new reference image). 

This seems to confirm that a `STYLE ANGLE` can use negative values as discussed in #6463. 

Tests will currently fail at the moment with the error `loadStyle(): General error message. Invalid ANGLE, must be between 0 and 360 (line 77) <br>`

@sdlime - should the validation for `STYLE ANGLE` allow the range -360 to 360 as with the other angles? Or is further testing required?

### DIFF
--- a/msautotest/renderers/symbolrot_negative.map
+++ b/msautotest/renderers/symbolrot_negative.map
@@ -1,0 +1,151 @@
+# RUN_PARMS: symbolrot.png [MAP2IMG] -m [MAPFILE] -i png -o [RESULT]
+
+#
+# Test support for rotating symbols counter-clockwise with their map.
+# This should produce the same image as when the ANGLE values are clockwise
+#
+# REQUIRES: OUTPUT=PNG SUPPORTS=PROJ
+#
+MAP
+    NAME "test"
+
+    EXTENT 520000 140000 540000 160000
+    UNITS METERS
+
+    IMAGECOLOR 255 255 255
+    STATUS ON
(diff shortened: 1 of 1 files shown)


sdlime:
> @sdlime - should the validation for STYLE ANGLE allow the range -360 to 360 as with the other angles? Or is further testing required?

Seems safe enough and parity would be a good thing.

geographika:
The first Travis test run failed for unrelated reasons. Second passed - so using 0 to 360 clockwise and 0 to -360 anticlockwise produce identical outputs for `ANGLE` in `MAP`, `CLASS`, and `STYLE` objects.

sdlime:
I suppose we should try negative angles by themselves for labels and styles. I know I tested labels independently but I don't think I tried a style.

geographika:
This test checks `STYLE  ANGLE -315`, and the `labelrot.map` tests `LABEL ANGLE -45` so I think they're covered?

sdlime:
> This test checks `STYLE ANGLE -315`, and the `labelrot.map` tests `LABEL ANGLE -45` so I think they're covered?

But with an explicit MAP ANGLE. What happens if you remove the map angle?

PR #6472, opened 2022-02-03
Reproject layer extent when changing layer projection

PatrikSylve:
Hi, I encountered a case when requesting a layer with `CONNECTIONTYPE WMS` would be ignored. 

In my case, I requested the data in EPSG:3857, and the requested layer's default projection was EPSG:4326. When mapserver was building the WMS url, it would change the projection of the layer object to match the requested SRS 3857 but leave the EXTENT unchanged.  When the image response had been fetched, it was ignored in the final response as it didn't pass the extent check in  `msLayerIsVisible`. 

This change will reproject the layer objects extent to match the layers assigned projection.


geographika:
@PatrikSylve - the change appears to make sense. Do you have an example of the layer configuration / Mapfile and WMS request parameters that show the issue?

Ideally a test could be added to show this is failing and then fixed with the test. [This test](https://github.com/MapServer/MapServer/blob/f2e8fe1db6df4ad08138f519dee09bc35fbfc48c/msautotest/gdal/wmsclient.map) could be adapted?

jmckenna:
> Added a test case now that will output an empty image without the fix.

Nice test, and glad to see demo.mapserver.org services used here.  Thanks!

jmckenna:
same here.  Let's give it a go...   Great catch @PatrikSylve !

PR #6477, opened 2022-02-12
EnablePaging methods: check return of Open() method to avoid assertions

rouault:


sdlime:
@rouault, should this be backported to 7.6?

jmckenna:
@sdlime we may need a 7.6.5 release, if my hunch is correct of major PROJ >=6 issues being recently discovered, so I think we should backport.

PR #6484, opened 2022-02-25
Accept ESRI:XXX (or other autorities) as value for PROJECTION blocks and wxs_srs metadata

rouault:


jmckenna:
thanks @rouault !

PR #6491, opened 2022-03-14
Python MapScript updates to work with Python 3.10

geographika:
This update fixes issues when building MapScript for Python 3,10 with (as yet unreleased) SWIG 4.1.0. 

- Calls to `PyObject_CallMethod` (used by MapScript for writing images) need to use `Py_ssize_t` rather than `int` for specifying size (to allow output > 2GB). See https://bugs.python.org/issue40943
- SWIG methods all now have `__SWIG` appended to them which broke `intarray___setitem__(order, i, self->layerorder[i]);` - I'm unsure why the method name rather than array access was used in the first place.

### DIFF
--- a/mapscript/python/pyextend.i
+++ b/mapscript/python/pyextend.i
@@ -13,6 +13,10 @@
  *
  *****************************************************************************/
 
+// required for Python 3.10+ - see https://bugs.python.org/issue40943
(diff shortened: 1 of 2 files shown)


jmckenna:
thanks @geographika !

Issue #6493, opened 2022-03-14
Labelling lines with zero lengths may cause the application to crash


szekerest:
To reproduce this a LINE layer should be configured with TRANSFORM FALSE and the shapefile should contain a line with zero length (all points are at the same location).

The crash location is at:

```
                mapserver!msLineLabelPoint+0x209 [mapprimitive.c @ 1717]
		mapserver!msPolylineLabelPoint+0x141 [mapprimitive.c @ 1648]
		mapserver!lineLayerDrawShape+0x41c [mapdraw.c @ 1854]
		mapserver!msDrawShape+0x80d [mapdraw.c @ 2169]
		mapserver!msDrawVectorLayer+0x7be [mapdraw.c @ 1164]
		mapserver!msDrawLayer+0x3a3 [mapdraw.c @ 813]
		mapserver!msDrawMap+0x415 [mapdraw.c @ 403]
		mapscript!CSharp_mapObj_draw+0xd
```

Note: Not all of the data providers support returning such invalid geometries, the shapefile driver does.

szekerest:
yes

PR #6494, opened 2022-03-14
Labeling lines with zero lengths may cause the application to crash (#6493)

szekerest:


sdlime:
It's not clear to me what these changes prevent downstream. What am I missing?

szekerest:
@sdlime These changes causes to initialize the values (longest_segment_index, longest_segment_line_index, longest_segment_point_index, longest_line_index) even if the line length is zero.

sdlime:
I saw that but I couldn't tell why that fixes anything.

szekerest:
Because, the following code in (msLineLabelPoint) will use uninitialized longest_segment_index as soon as the line length is zero.

```
        j = ll->longest_segment_index;
        lar->label_points[lar->num_label_points].x = (p->point[j-1].x + p->point[j].x)/2.0;
        lar->label_points[lar->num_label_points].y = (p->point[j-1].y + p->point[j].y)/2.0;
```

sdlime:
> Because, the following code in (msLineLabelPoint) will use uninitialized longest_segment_index as soon as the line length is zero.
> 
> ```
>         j = ll->longest_segment_index;
>         lar->label_points[lar->num_label_points].x = (p->point[j-1].x + p->point[j].x)/2.0;
>         lar->label_points[lar->num_label_points].y = (p->point[j-1].y + p->point[j].y)/2.0;
> ```

Ok, so I think what the changes do. Ok with me to move forward... --Steve

PR #6430, opened 2021-10-28
WIP: remove PHP-native MapScript code for 8.0 release

jmckenna:
- part of the MapServer 8.0 release plan (https://github.com/MapServer/MapServer/wiki/MapServer-8.0-Release-Plan)
- as PHP-native MapScript is unmaintained and now causing constant security concerns & efforts
- as PHP 7 & 8 are now well supported by SWIG
  - see issue #6201 for PHP 8.0 progress with SWIG MapScript
- this pull request:
  - removes old native PHP MapScript
  - updates TravisCI and GitHub build action to use Ubuntu Focal
  - builds PHPNG MapScript for PHP 7.4.6, 8.0.19, 8.1.6
  - builds Python MapScript for Python 3.7.7, 3.8.3, 3.9.0
  - updates many msautotest tests
  - updates all PHP unit tests

PR #6496, opened 2022-03-15
Add Python MapScript annotations using SWIG 4.1

geographika:
This pull request allows functions and properties in the Python MapScript bindings to automatically include type annotations. These are useful for when generating the MapScript API docs (see https://github.com/MapServer/MapServer-documentation/pull/544 and [RFC 132](https://mapserver.org/development/rfc/ms-rfc-132.html)). 

This is made possible by recent changes in SWIG - see https://github.com/swig/swig/pull/1951#issuecomment-1057325471
and is merged into the SWIG master branch, so will be a part of the 4.1 release. The Appveyor script makes use of a SWIG version compiled against this branch. 

Annotations are off by default and can be set using `WITH_PYMAPSCRIPT_ANNOTATIONS=1` in CMake. 

The `WITH_MAPSCRIPT_PROPERTYANNOTATIONS` has been renamed `WITH_PYMAPSCRIPT_ANNOTATIONS` for clarity. 

Python 3.5 and earlier do not support property annotations, so if enabled only methods will be annotated. 
Python 2.7 does not support annotations at all, so enabling annotations is ignored. 

Note the SWIG command switch is `-features python:annotations=c`. Types are currently C types, but there are plans in SWIG to switch these to Python types in the future.


jmckenna:
Thanks @geographika !

(for the records, the MapServer TravisCI builds were recently also updated to pull SWIG-master, through https://github.com/MapServer/MapServer/pull/6430 )

PR #6497, opened 2022-03-21
Fix CMake package name warnings

geographika:
This fixes several of the following types of CMake warnings (in version 3.22) in the build outputs, e.g.

```
  The package name passed to `find_package_handle_standard_args` (PkgConfig)
  does not match the name of the calling package (HarfBuzz).  This can lead
  to problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
```

One of these still remains where it is searching for `SVG` in a `FindSVGCairo` package. I'm unsure how to resolve this one. 

https://github.com/MapServer/MapServer/blob/a43a21de9a1973178f4405f597c765d6a075a014/cmake/FindSVGCairo.cmake#L47


geographika:
No issues with these changes on the older CMakes used by the CI, or on newer CMakes (where I'm getting the warnings), so hopefully good to merge.

PR #6500, opened 2022-03-23
More CMake package name warning fixes

geographika:
Following on from #6497 a couple more fixes. 
I'm not 100% sure about the change from `include(FindPerl)` to `find_package(Perl)` - will see what happens with the CI. 

@sdlime - I've not been building with Perl and Postgres so these got missed in the first pull request.


geographika:
Travis is passing which tests both `WITH_PERL=1` and `WITH_POSTGIS=1`. @sdlime if you get a chance to confirm locally that the warnings are gone for you, then I'll mark this as ready for review.

sdlime:
> 

Yup, warnings are gone locally.

PR #6509, opened 2022-04-09
find PG 14 through CMake

jmckenna:
- modifies the embedded FindPostgreSQL.cmake for older CMake versions


rouault:
Actually even the latest CMake didn't include PG 14. I've just submitted that to upstream per https://gitlab.kitware.com/cmake/cmake/-/merge_requests/7158

Issue #6521, opened 2022-05-02
Shapefile driver fails to load layer with null shapes


szekerest:
The indexed access of the shapes fails as of this commit

https://github.com/MapServer/MapServer/commit/67fb0e85d614df4ebad510cda6fd27d9d7ac9675

PR #6522, opened 2022-05-02
Handle null shapes in msShapefileWhichShapes (#6521)

szekerest:


jmckenna:
great catch @szekerest

PR #6527, opened 2022-05-14
CMake: fix recent issue with mapscript python

rouault:
where 'make' no longer copied files needed to run 'python setup.py bdist'

follow-up of 2f52ef4b6613c7bc9b1cad327a2aff179978cd9d

CC @szekerest


szekerest:
Thanks @rouault

PR #6534, opened 2022-06-15
Expose reprojectionObj to MapScript

sdlime:
Legacy MapScript projection operations are very slow when used in batch context - e.g. looping over a series of geometries and reprojecting. This pull request exposes the reprojectionObj to MapScript and adds overloaded project methods to point, line, rect and shape classes. Use looks like so (in Perl):
```
my $proj_26915 = new mapscript::projectionObj('epsg:26915');
my $proj_4326 = new mapscript::projectionObj('epsg:4326');
my $reprojector =  new mapscript::reprojectionObj( $proj_26915,  $proj_4329 );

while (my $shape = $layer->nextShape()) {
    my $point = $shape->getCentroid();
    $point->project($reprojector);
    # do something with $point
}
```
--Steve

### DIFF
--- a/mapproject.h
+++ b/mapproject.h
@@ -87,9 +87,22 @@ but are not directly exposed by the mapscript module
     int wellknownprojection; ///< One of ``wkp_none 0``, ``wkp_lonlat 1``, or ``wkp_gmerc 2``
   } projectionObj;
 
+  typedef struct {
+#ifdef SWIG
+    %immutable;
+#endif
+    projectionObj* in;
+    projectionObj* out;
+    PJ* pj;
+    int should_do_line_cutting;
+    shapeObj splitShape;
+    int bFreePJ;
+#ifdef SWIG
+    %mutable;
+#endif
+  } reprojectionObj;
 #ifndef SWIG
 
-  typedef struct reprojectionObj reprojectionObj;
   MS_DLL_EXPORT reprojectionObj* msProjectCreateReprojector(projectionObj* in, projectionObj* out);
   MS_DLL_EXPORT void msProjectDestroyReprojector(reprojectionObj* reprojector);
 
@@ -104,10 +117,10 @@ but are not directly exposed by the mapscript module
   MS_DLL_EXPORT int msProjectShapeEx(reprojectionObj* reprojector, shapeObj *shape);
   MS_DLL_EXPORT int msProjectLine(projectionObj *in, projectionObj *out, lineObj *line); /* legacy interface */
   MS_DLL_EXPORT int msProjectLineEx(reprojectionObj* reprojector, lineObj *line);
-  MS_DLL_EXPORT int msProjectRect(projectionObj *in, projectionObj *out, rectObj *rect);
+  MS_DLL_EXPORT int msProjectRect(projectionObj *in, projectionObj *out, rectObj *rect); /* legacy interface */
+  MS_DLL_EXPORT int msProjectRectAsPolygon(reprojectionObj* reprojector, rectObj *rect);
   MS_DLL_EXPORT int msProjectionsDiffer(projectionObj *, projectionObj *);
-  MS_DLL_EXPORT int msOGCWKT2ProjectionObj( const char *pszWKT, projectionObj *proj, int
-      debug_flag );
+  MS_DLL_EXPORT int msOGCWKT2ProjectionObj( const char *pszWKT, projectionObj *proj, int debug_flag );
   MS_DLL_EXPORT char *msProjectionObj2OGCWKT( projectionObj *proj );
 
   MS_DLL_EXPORT void msFreeProjection(projectionObj *p);
--- a/mapproject.h
+++ b/mapproject.h
@@ -88,19 +88,26 @@ but are not directly exposed by the mapscript module
   } projectionObj;
 
   typedef struct {
-#ifdef SWIG
-    %immutable;
-#endif
+#if PROJ_VERSION_MAJOR >= 6
+#ifndef SWIG
     projectionObj* in;
     projectionObj* out;
     PJ* pj;
     int should_do_line_cutting;
     shapeObj splitShape;
     int bFreePJ;
-#ifdef SWIG
-    %mutable;
+#endif
+#else
+#ifndef SWIG
+    projectionObj* in;
+    projectionObj* out;
+    int should_do_line_cutting;
+    shapeObj splitShape;
+    int no_op;
+#endif
 #endif
   } reprojectionObj;
+
 #ifndef SWIG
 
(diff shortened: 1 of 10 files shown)


mapproject.h
```diff
@@ -87,9 +87,22 @@ but are not directly exposed by the mapscript module
     int wellknownprojection; ///< One of ``wkp_none 0``, ``wkp_lonlat 1``, or ``wkp_gmerc 2``
   } projectionObj;
 
+  typedef struct {
```
rouault, line 90: We could likely just use a forward declaration here "typedef struct reprojectionObj reprojectionObj;" here, and let the definition in mapproject.c, since nobody needs to inspect the content of the fields.

mapscript/swiginc/reprojection.i
```diff
@@ -0,0 +1,45 @@
...
+   ===========================================================================
+*/
+
+
+%extend reprojectionObj
+{
+    /// Create new instance of reprojectionObj. Input parameters are two projectionsObj's - in and out.
+    reprojectionObj(projectionObj *in, projectionObj *out) 
+    {
+	if(!in || !out) return NULL;
```
rouault, line 36: indentation issue.
and msProjectCreateReprojector() supports one of in or out to be NULL, and that means that it is the geographic CRS corresponding to the non-NULL argument. So the test should rather use a &&

rouault:
a test in msautotest/python would also be welcome

mapscript/swiginc/reprojection.i
sdlime: Made that change, thanks!

sdlime:
> a test in msautotest/python would also be welcome

Agreed, just gotta brush up on my Python.

mapproject.h
sdlime: That's what was there originally right? Swig choked on it.

mapscript/swiginc/reprojection.i
```diff
@@ -0,0 +1,44 @@
...
+   ===========================================================================
+*/
+
+
+%extend reprojectionObj
+{
+    /// Create new instance of reprojectionObj. Input parameters are two projectionsObj's - in and out.
+    reprojectionObj(projectionObj *in, projectionObj *out) 
+    {
+	if(!in && !out) return NULL;
```
geographika, line 36: Should be indented?

geographika (approved):
It would be good to add in test cases for these for the various objects e.g. in https://github.com/MapServer/MapServer/blob/main/mapscript/python/tests/cases/line_test.py
These could be added in subsequent pull requests.

mapproject.h
rouault: hum, there's maybe a subtelty. In GDAL, we used forward declarations and SWIG, but I see that the type mapped to SWIG is actually a pointer to a struct, not the struct itself.
```
/** Opaque type for C++ GDALMDArray */
typedef struct GDALMDArrayHS* GDALMDArrayH;
```
(The issue of exposing the structure content is that it is not the same depending on PROJ >= 6 or PROJ < 6.)

So maybe have in mapproject.h
typedef struct reprojectorObj* reprojectObjHandle;

and in swiginc/reprojection.i

```
class reprojectorObj {
private:
  reprojectorObj();
public:
%extend {
....
}
```

sdlime: Maybe we need to give up on proj < 6 altogether?

sdlime:
Build check failure is related to PHP and not this pull request - as near as I can tell anyway.

jmckenna:
> Build check failure is related to PHP and not this pull request - as near as I can tell anyway.

Thanks Steve, diving in....

jmckenna:
@sdlime please merge from `main` again into this branch, thanks.

sdlime:
> @sdlime please merge from `main` again into this branch, thanks.

I re-ran the build and Travis checks - build completed fine, so that's good.

mapproject.h
rouault: > we need to give up on proj < 6 altogether?

with that commit, they will certainly break due 2 incompatible definitions of struct reprojectionObj (the PROJ 6 one in mapproject.h and the remaining PROJ4 one in mapproject.c)

sdlime: I fiddled with the suggestion:

`typedef struct reprojectorObj* reprojectObjHandle;`

but ran into (what look to be) complicated Swig issues. 

I hadn't noticed the second reprojectionObj definition in mapproject.c. Could both the Proj 6 and Proj 4 definitions both be moved to mapproject.h?

rouault: > Could both the Proj 6 and Proj 4 definitions both be moved to mapproject.h?

yes, that's doable.
But I'm wondering if the plan couldn't be to have in mapproject.h just
```
#ifdef SWIG
typedef struct {
   %immutable;
   int opaque; // to have a non empty structure
} reprojectionObj;
#endif
```

that would only be used by SWIG,and keep the actual definitions in mapproject.c
But that's still unsatisfactory. GDAL SWIG binding clearly shows there's a way to have opaque structures mapped to SWIG, although I know fighting SWIG is not fun :-)

sdlime: The error I received with the opaque structure mapping (and definitions in mapproject.c) was "Warning 303: %extend defined for an undeclared class reprojectionObj.". Could the form of the structure declarations in mapproject.c have anything to do with it:

```
struct reprojectionObj
{
    projectionObj* in;
    projectionObj* out;
    PJ* pj;
    int should_do_line_cutting;
    shapeObj splitShape;
    int bFreePJ;
};
```
versus:
```
typedef struct
{
    projectionObj* in;
    projectionObj* out;
    PJ* pj;
    int should_do_line_cutting;
    shapeObj splitShape;
    int bFreePJ;
} reprojectionObj;
```

jmckenna:
@sdlime I've just pushed a PHP MapScript unit test, into this branch.  It tests the new reprojectionObj instance, and also the project() method on ~8,000 points (and outputs the speed).

Of note is that by default (as this same way is used throughout msautotest) I tried `new projectionObj("proj=latlong")` as the source projection, and it fails silently on Windows, but of course using `new projectionObj("EPSG:4326")` works as expected.

jmckenna (approved):

sdlime:
> @sdlime I've just pushed a PHP MapScript unit test, into this branch. It tests the new `reprojectionObj` instance, and also the `project()` method on ~3800 points (and outputs the speed).

Cool! Maybe we can crib that test for Python as well. Out of curiosity, what's the speed like without the reprojectionObj? That is calling project with the 2 projectionObj's.

Note I pushed this back to draft status until handling the reprojectionObj structure definitions is sorted out.

mapproject.h
sdlime: @rouault, does the latest change look ok to you? Looks like @jmckenna's PHP tests ran fine.

sdlime:
Ok, I think we're good to go with this.

PR #6540, opened 2022-06-26
Also install coshp utility.

sebastic:
#6518 adds the utility but doesn't actually install it.


jmckenna:
thanks @sebastic   (this will be part of beta2, as beta1 was tagged minutes before this)

Issue #6543, opened 2022-06-27
8.0.0beta1 FastCGI only replies to one request in error, then is stuck


landryb:
on OpenBSD testing the 8 beta, using spawn-fcgi (without chrooting) eg `spawn-fcgi -s /var/www/run/mapserv.sock -u www -- /var/www/cgi-bin/mapserv` or supervisord, i have a strange behaviour of mapserver which only replies to the first request with an error message coming from https://github.com/MapServer/MapServer/blob/main/mapfile.c#L6288

outputs `msLoadMap(): General error message. getcwd() returned a too long path`

my test query is `http://serverip/cgi-bin/mapserv?map=/tmp/mapserver/ign.map`

`/etc/mapserver.conf` is super basic:
```
CONFIG
  ENV
    MS_MAP_PATTERN "^/tmp/mapserver"
    MS_DEBUGLEVEL "5"
    MS_ERRORFILE "/tmp/mapserver.log"
  END
   MAPS
  END
END
```

then the process ends up in a stuck state, eg doesnt replies anymore to requests, nginx returns a 502, the process still runs but needs to be killed.

i dont understand how/why this error codepath is taken.

landryb:
at some point i thought this was related to params conflicting, eg i had `fastcgi_param MS_MAPFILE /tmp/mapserver/ign.map;` in my nginx config, but with that commented out it doesnt help.

landryb:
after instrumenting the error message returned by getcwd() by printing errno, the failure is EACCES.

after fiddling with spawn-fcgi options, adding `-d /tmp` (which forces a chdir() to `/tmp`, otherwise i guess it chdir() to somewhere else ?) doesn't result in more errors from `getcwd()`, and i have ms8 working ... the error message was definitely not helpful.

landryb:
thanks a lot @rouault ! #6545 looks much better

landryb:
now that i dig deeper in this issue, i found out that the behaviour of 'only replies to one request with an error then is stuck' looks like a regression from adding ogc api support in #6180.

to summarize, with 7.6.4 configured as fastcgi spawned by spawn-fcgi server behind nginx, i can load http://localhost/cgi-bin/mapserv without any argument, i get `No query information to decode. QUERY_STRING is set, but empty. ` from https://github.com/MapServer/MapServer/blob/main/cgiutil.c#L237 which is expected, and if i repeat the same request i still get the same error message.

[...]

PR #6545, opened 2022-06-27
Improve error message in case of getcwd() failure (fixes #6543)

rouault:


jmckenna:
thanks @rouault and @landryb !

Issue #6550, opened 2022-06-29
native flatgeobuf triggers endless loop


landryb:
testing ogcapi with 8.0.0b1 on OpenBSD, i managed to trigger an endless loop in native flatgeobuf:
```
(gdb) r QUERY_STRING="" PATH_INFO=/ign/ogcapi/collections/points/items/
Starting program: /usr/local/bin/mapserv QUERY_STRING="" PATH_INFO=/ign/ogcapi/collections/points/items/
headersize=1628
^C
Program received signal SIGINT, Interrupt.
realloc (ptr=0x6c3a4b47220, size=0) at /usr/src/lib/libc/stdlib/malloc.c:1793
1793    done:
(gdb) bt
#0  realloc (ptr=0x6c3a4b47220, size=0) at /usr/src/lib/libc/stdlib/malloc.c:1793
#1  0x000006c3d123e076 in flatgeobuf_ensure_buf (ctx=0x6c3a4b21b00, size=1628) at /usr/obj/ports/mapserver-8.0.0-beta1/mapserver-8.0.0-beta1/flatgeobuf/flatgeobuf_c.cpp:84
#2  flatgeobuf_decode_header (ctx=0x6c3a4b21b00) at /usr/obj/ports/mapserver-8.0.0-beta1/mapserver-8.0.0-beta1/flatgeobuf/flatgeobuf_c.cpp:261
#3  0x000006c3d121bd9e in msFlatGeobufLayerOpen (layer=0x6c3a4b31000) at /usr/obj/ports/mapserver-8.0.0-beta1/mapserver-8.0.0-beta1/mapflatgeobuf.c:150
#4  0x000006c3d11612c2 in msQueryByRect (map=0x6c34a194000) at /usr/obj/ports/mapserver-8.0.0-beta1/mapserver-8.0.0-beta1/mapquery.c:1115
#5  0x000006c3d1160501 in msExecuteQuery (map=0x6c34a194000) at /usr/obj/ports/mapserver-8.0.0-beta1/mapserver-8.0.0-beta1/mapquery.c:113
#6  0x000006c3d107f2ef in processCollectionItemsRequest (map=<optimized out>, request=0x6c365ba3680, collectionId=<optimized out>, featureId=0x0, format=HTML)
    at /usr/obj/ports/mapserver-8.0.0-beta1/mapserver-8.0.0-beta1/mapogcapi.cpp:1031
[...]

PR #6552, opened 2022-06-29
fix native flatgeobuf

landryb:
cf #6550 - still work in progress.


bjornharrtell:
This is possibly an indication of issues with unindexed flatgeobuf files... which I've not really taken into consideration (I think I want to disallow those entirely and throw a meaningful error message). But the file you generated *should* have an index so I'm still puzzled currently.

bjornharrtell:
flatgeobuf_ensure_buf was indeed dubious.. however I have fixed that in another way with safeguards in #6553.

And flatgeobuf_read_feature_offset was also incorrect, also have an alternative fix for that in #6553 but I need to test it.

bjornharrtell:
This was not enough to get basic WFS to work, I found multiple problems.. This is replaced by https://github.com/MapServer/MapServer/pull/6553 which makes basic WFS work for me.

bjornharrtell:
Closing as replaced by #6553.

PR #6555, opened 2022-07-04
Coverity Scan fixes

rouault:


jmckenna:
thanks @rouault !

sdlime:
> thanks @rouault !

Yes, thanks @rouault! (just back from vacation)

PR #6557, opened 2022-07-04
msProjectShapeLine(): fix reprojection of lines crossing the antimeridian when source CRS is +proj=longlat +lon_wrap=0

rouault:


jmckenna:
thanks @rouault !

PR #6559, opened 2022-07-04
SLD: unset layer classgroup …

rouault:
…unless SLD content has a named class in UserStyle.IsDefault, in which case we use that named class (soften changes d89ac1fadbd34bed5e991c834c601d3e9026091c)


jmckenna:
thanks @rouault !

PR #6565, opened 2022-07-07
Config file parsing: use msSetPROJ_LIB() when ENV PROJ_LIB is set

rouault:


geographika:
@rouault - I tested with this pull request and encountered no issues. 

There is still the following scenario (with FastCGI) that is confusing:

- Set PROJ_LIB in the CONFIG file, and request a map with no `CONFIG PROJ_LIB` setting. The setting in CONFIG is used
- Now request another Mapfile that has its own `CONFIG PROJ_LIB` setting - this is ignored and the setting in the CONFIG is still used

I'm not sure if this is expected behaviour?

For reference thread from mailing list where this was discussed: https://lists.osgeo.org/pipermail/mapserver-dev/2022-July/016818.html

rouault:
> * Now request another Mapfile that has its own `CONFIG PROJ_LIB` setting - this is ignored and the setting in the CONFIG is still used

On Linux with Apache and FCGI, I don't confirm this:  CONFIG "PROJ_LIB" of the mapfile overrides ENV PROJ_LIB of the config file. However there are subtle effect: if after a successful request with that setup, I comment out CONFIG "PROJ_LIB" in the mapfile or set it to another value (wrong path typically), it is not taken into account right away. I need to restart Apache. There must be some caching.

jmckenna:
It sounds like this is ok to merge.  Will merge in the next hour, unless someone screams...

geographika:
> On Linux with Apache and FCGI, I don't confirm this: CONFIG "PROJ_LIB" of the mapfile overrides ENV PROJ_LIB of the config file. However there are subtle effect: if after a successful request with that setup, I comment out CONFIG "PROJ_LIB" in the mapfile or set it to another value (wrong path typically), it is not taken into account right away. I need to restart Apache. There must be some caching.

Very similar in IIS - likely when a new FastCGI process takes over (after 3 mins) then whichever setting is first picked up seems to be cached. 

@sdlime - it was probably always this way, but we should probably update the docs to recommend using the `CONFIG` file for this setting as different Mapfiles could use conflicting paths for `PROJ_LIB` and this could cause inconsistent responses from the server.

PR #6573, opened 2022-07-12
Apply special processings of PROJ_LIB to PROJ_DATA as well

rouault:
PROJ 9.1 recognizes the PROJ_DATA environment variable, as a replacement
for PROJ_LIB (both are still recognized by PROJ >= 9.1. PROJ_LIB will be
retired in PROJ 10, whose release is unplanned for now)

Cf https://github.com/OSGeo/PROJ/pull/3253


jmckenna:
thanks @rouault !

PR #6574, opened 2022-07-13
Skia renderer

bjornharrtell:


sdlime:
Well this is an interesting addition. Are there advantages over AGG?

bjornharrtell:
Did some early benchmarking on 3 million + polygon squares. AGG is a lot faster on that case... somewhere 50% - 80% faster. I believe I've compiled Skia with the optimal settings so currently I can't explain this other than that Skia isn't as fast as I had assumed/thought.

pelord:
https://github.com/mapbox/mapbox-gl-js/issues/1

jratike80:
@pelord , could you clarify what you want to say?

jmckenna:
@bjornharrtell just letting you know that the next release plan is coming together, for 8.4.0, in case you have time available before the release to tackle this again.  See [here ](https://github.com/MapServer/MapServer/wiki/MapServer-8.4-Release-Plan) for the release plan dates.

(judging by the earlier responses you got about this, I'd say that you have something very interesting, potentially, to others ;)

oshawa-connection:
> Did some early benchmarking on 3+ million polygon squares. AGG is a lot faster on that case... somewhere 50% - 80% faster. I believe I've compiled Skia with the optimal settings so currently I can't explain this other than that Skia isn't as fast as I had assumed/thought.

Just to add my $0.02. I’ve tried something similar with blend2d graphics library, and I get similar results to you with skia: it’s slower than agg despite all benchmarks saying it’ll be faster. 

Consider that most benchmarks are run in programs where the context only has to be setup once as part of a continuously running program, generating several frames as part of some animation.

My theory is that for fast draws of e.g. simple polygons, the overhead of setup of the graphics library context makes the overall render slower than agg. and because CGI that affects all renders. Don’t know if that matches up with what you’ve seen?

bjornharrtell:
@oshawa-connection yeah it could be that the architecture of skia that allows för gpu accel is a disadvantage. That said, I believe they have optimized the software pipeline to great lengths regardless. But I've not dived deep enough to know anything really.

PR #6577, opened 2022-08-03
reset isreported to MS_FALSE in msResetErrorList() (#6543)

landryb:


landryb:
fixes #6543 for me but more dragons might be lurking.

jmckenna:
@landryb please also add a test for this in msautotest (follow guidelines at https://mapserver.org/development/tests/autotest.html ) and include it in each of your pull requests.  Thanks!

geographika:
I'm not sure this can be tested using msautotest, as it is the FastCGI that would need to be tested.
Even added an Apache test for this in https://github.com/MapServer/MapServer/pull/6294
Maybe similar could be done for `spawn_fcgi`?

sdlime:
> Good catch @landryb! Resetting all fields makes sense to me.
> 
> I'm not sure this can be tested using msautotest, as it is the FastCGI that would need to be tested. Even added an Apache test for this in #6294 Maybe similar could be done for `spawn_fcgi`?

I can confirm this change fixes the issue on my installation. So this wasn't just an 8.0 issue... Nice @landryb!

jmckenna:
@landryb if you now fetch the latest commits from `main` into this PR branch, the build should run fine.

landryb:
> @landryb please also add a test for this in msautotest (follow guidelines at https://mapserver.org/development/tests/autotest.html ) and include a test in each of your pull requests. Thanks! (the more time you spend creating tests in msautotest, the less dragons faced in the future, ha)

e8ee9b8 should test that but with the existing apache setup. I dont have a linux setup around right now to go through all the nginx/spawn-fcgi setup madness :)

landryb:
well afaict from the build job log the two tests are run:
```
2022-08-04T05:15:16.3933693Z Running CGI query
2022-08-04T05:15:16.4849104Z Running FastCGI query
2022-08-04T05:15:16.5642633Z Running FastCGI query again
2022-08-04T05:15:16.5756039Z Check that we return an error if given no arguments
2022-08-04T05:15:16.5850926Z Check again to make sure further errors are reported (#6543)
```

jmckenna:
> [e8ee9b8](https://github.com/MapServer/MapServer/commit/e8ee9b83b5bc270c065ef3372b38ea3622597405) should test that but with the existing apache setup. I dont have a linux setup around right now to go through all the nginx/spawn-fcgi setup madness :) Oh, and completely untested..

thanks for adding this, much appreciated.

PR #6584, opened 2022-08-04
reset layer filteritem to its old value when the query extent doesnt overlap the layer extent

### LINKED ISSUE #6580: double-free on query->filteritem with ogcapi (and africa.fgb)
extracted from https://github.com/MapServer/MapServer/issues/6550#issuecomment-1170493133

im chasing a double-free fully reproducible using OGCAPI, using a query trying to access a random value which would be 'outside' the datafile, eg

`PATH_INFO=/ign/ogcapi/collections/africa/items/1244`

the mapfile has 4 layers, the africa layer is the 3rd reuses the fgb present in the repo for msautotest:
```
    LAYER
        NAME "africa"
        TYPE polygon
        METADATA
        "ows_title" "africa"
          "oga_include_items" "pop_est,name_en"
          "oga_featureid" "id"
        END
                EXTENT -17.62 -34.81 51.139 37.4
        CONNECTIONTYPE flatgeobuf
      DATA "/home/landry/src/mapserver/msautotest/misc/data/africa.fgb"
        CONNECTIONOPTIONS "VERIFY_BUFFERS" "NO" END
        TEMPLATE VOID
        END
```

at runtime on OpenBSD, i get this double-free message coming from our own malloc that does memory corruption detection:
```
$./mapserv PATH_INFO=/ign/ogcapi/collections/africa/items/1244                                                               
Content-Type: application/json
Status: 404

{"code":"NotFound","description":"Collection items id query failed."}
mapserv(53919) in free(): chunk is already free 0xc2646c44ff0
Abort trap (core dumped) 
```
strangely, trying to access all items in the layer also fails:
```
./mapserv PATH_INFO=/ign/ogcapi/collections/africa/items 
Content-Type: application/json
Status: 500

{"code":"ConfigError","description":"Template rendering error. [inja.exception.render_error] (at 52:22) variable 'response.features.0.properties' not found (collection-items.html)."}

```
i have no issue accessing another fgb datafile (`points` layer) with the same kind of queries, be it all items or a single item by its id.

looking at the code, it seems `filteritem` pointers are at some point shared by the `query` object and the `layer` object:
```
$git grep filterit|grep -i free
mapfile.c:  msFree(layer->filteritem);
maplayer.c:    msFree(layer->filteritem);
maplayer.c:    if(lp->filteritem) free(lp->filteritem);
maplayer.c:      free(lp->filteritem);
mapquery.c:  if(query->filteritem) free(query->filteritem);
```
so maybe one of those is freeing too much ? apparently in the code using filteritem here and there, sometimes `query-filteritem` pointer is assigned to the layer object, which might already free it ?

crash with double-free

`./mapserv PATH_INFO=/ign/ogcapi/collections/africa/items/1244`
## Attach simple test case (drag/drop it here)

in mapserver.conf i have this section for ogcapi mapping to mapfile:
```
  MAPS
    "ign" "/home/landry/gis/mapfile/ign.map"
  END
```

zip with mapfile and two working fgb files below:
[data.zip](https://github.com/MapServer/MapServer/files/9253152/data.zip)

## Operating system

OpenBSD - note that we have stronger memory constraints than on other operating systems, so would explain why the bug cant be reproduced on other like linux - but clang/llvm asan on valgrind should show the double-free too.

master

landryb:
fixes #6580 - after all we're looping on all map layers so we cant just `goto restore_old_filter`, so i just copy/pasted a similar idiom found in other places in mapquery.c

more eyes welcome on the logic, but fixes the double-free for me in the case where the layer extent doesn't overlap with the map/query extent.

### DIFF
--- a/mapquery.c
+++ b/mapquery.c
@@ -905,6 +905,9 @@ int msQueryByFilter(mapObj *map)
 
     status = msLayerWhichShapes(lp, search_rect, MS_TRUE);
     if(status == MS_DONE) { /* no overlap */
+      lp->filteritem = old_filteritem; /* point back to original value */
+      msCopyExpression(&lp->filter, &old_filter); /* restore old filter */
+      msFreeExpression(&old_filter);
       msLayerClose(lp);
       continue;
     } else if(status != MS_SUCCESS) goto restore_old_filter;


sdlime:
I approve... @landryb thanks for digging in on this (and other) issue(s)!

jmckenna:
thanks again @landryb !

PR #6594, opened 2022-08-14
Set Standerd for Oss-fuzz initialization.

0x34d:
Multiple changes for fuzzing.

1. fixing library linker for fuzzer.
2. add fuzzer for map files.
3. add how-to for fuzzing.
4. Set Standard for [Oss-fuzz](https://github.com/google/oss-fuzz) initialization.


sdlime:
Hi Arjun: Thanks for the contribution. Is this something you've been working on already - fuzzing MapServer I mean? Need to get a couple of other folks to review. --Steve

rouault:
@MaxKellermann any comment on this as you initiated the fuzzers directory ?

MaxKellermann:
I don't understand the commits in this PR, and the commit messages don't explain anything.

0x34d:
Hello @MaxKellermann,
Here is the Explanation.
- [change to standard fuzzer](https://github.com/MapServer/MapServer/pull/6594/commits/8471968d8a5bb0b5657de9ca3077a1458ad3c701)
    An ideal stage is to compile different ```fuzzers```, from ```one code base```. 
    ```ex-fuzzer: AFL,honggfuzz,libfuzzer;```
    ```ex-sanitizer:ASan,USan,MSan.```
    This can't be done with predefined flags.

- [add map fuzzer & switch fuzzer library linker to static](https://github.com/MapServer/MapServer/pull/6594/commits/62b047ba8091a68c0cd6b382618ea2f434228f59)
   This update has two things.
  1. Adding ``map file`` fuzzer.
      adding fuzzer support for ``` map files ```, for fuzzing map files.
  2. FixLikner error:
     While compiling the fuzzer, I found out that the fuzzer linked to ```shared maplib library```.
     This change is hardcore, it says fuzzer should always link to a ```static maplib library```.
    ```Note-```:-fuzzing engine cannot get coverage report from ```shared library```.

- [add FUZZER.md file for how-to](https://github.com/MapServer/MapServer/pull/6594/commits/9c46c7fd4ecdbaff32e650242ddcd9b554f0fea2)
it's a how-to guide to compiling fuzzer from now on.

0x34d:
Hi @sdlime,
My End goal is to integration [MapServer](https://github.com/MapServer/MapServer) into [oss-fuzz](https://github.com/google/oss-fuzz)
These changes are made for oss-fuzz infrastructure heavily.

MaxKellermann:
Sorry, @0x34d your explanation reads very cryptic to me, I still don't understand a thing (and if you had any explanation, please amend it to be part of the commit message; discussions on GitHub are lost after merging, only commit messages remain, and good commit messages are important).

> An ideal stage is to compile different fuzzers, from one code base.
> ex-fuzzer: AFL,honggfuzz,libfuzzer;
> ex-sanitizer:ASan,USan,MSan.
> This can't be done with predefined flags.

What does "an ideal stage is to ..." mean? What does "ex-fuzzers ... ex-sanitizers" mean?

>  While compiling the fuzzer, I found out that the fuzzer linked to shared maplib library.

And the problem with that is ... ? You wrote about "FixLikner error" - do you encounter linker errors? Why? Is that a bug in your toolchain, or is using the shared fuzzer library unsupported in your toolchain? If that's so, why is it your toolchain's default? This doesn't make sense for me.

> This change is hardcore, it says fuzzer should always link to a static maplib library.

What does "change is hardcore" mean? "It says" - who? And why?

geographika:
@0x34d - thanks for the interest shown in setting this up for the MapServer project. I don't know enough about the topic of fuzzers to be able to add much to a review. 

From the [oss-fuzz new project guide](https://google.github.io/oss-fuzz/getting-started/new-project-guide/#prerequisites) it states:

> Note: We prefer that you keep and maintain [fuzz targets](https://google.github.io/oss-fuzz/reference/glossary/#fuzz-target) in > your own source code repository. If this isn’t possible, you can store them inside the OSS-Fuzz project directory you created.

Maybe setting it up there so we can see how it works, and what issues are found would be a good place to start, and then move to the MapServer repository later?

0x34d:
> What does "an ideal stage is to ..." mean? What does "ex-fuzzers ... ex-sanitizers" mean?

an ideal stage for proper fuzzing.
There are three major fuzzers for fuzzing: [AFLplusplus](https://github.com/AFLplusplus/AFLplusplus)  [LibFuzzer](https://llvm.org/docs/LibFuzzer.html) [honggfuzz](https://github.com/google/honggfuzz)
There are three major sanitizers for fuzzing, to find interesting bugs: [AddressSanitizer](https://github.com/google/sanitizers/wiki/AddressSanitizer) [UndefinedBehaviorSanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) [MemorySanitizer](https://github.com/google/sanitizers/wiki/MemorySanitizer)


> And the problem with that is ... ? You wrote about "FixLikner error" - do you encounter linker errors? Why? Is that a bug in your toolchain, or is using the shared fuzzer library unsupported in your toolchain? If that's so, why is it your toolchain's default? This doesn't make sense for me.

Well, it's simple.
Target under test, The inbuild library for the target under-test needs to be static, not shared.
If the inbuilt library is shared, then the fuzzing engine can not get any coverage report for fuzzing.
If the fuzzing engine does not get a good coverage report,
then there is no good malformed input, and not any interesting bug can be found.

> What does "change is hardcore" mean? "It says" - who? And why?
It's regarding static library linking.

0x34d:
> Maybe setting it up there so we can see how it works, and what issues are found would be a good place to start, and then move to the MapServer repository later?


I can Do that.

I'm Going to use these mail for oss-fuzz. These mail IDs will receive the bug report 

- mapserver-security@osgeo.org
- max.kellermann@gmail.com
- sethg@geographika.co.uk

jmckenna:
> I'm Going to use these mail for oss-fuzz. These mail IDs will receive the bug report

please also include: jmckenna at gatewaygeomatics.com dot com

MaxKellermann:
> > What does "an ideal stage is to ..." mean? What does "ex-fuzzers ... ex-sanitizers" mean?
> 
> an ideal stage for proper fuzzing.

I still don't understand this. Are you using automatic translation to translate from your native language into English?

> There are three major fuzzers for fuzzing: [AFLplusplus](https://github.com/AFLplusplus/AFLplusplus) [LibFuzzer](https://llvm.org/docs/LibFuzzer.html) [honggfuzz](https://github.com/google/honggfuzz) There are three major sanitizers for fuzzing, to find interesting bugs: [AddressSanitizer](https://github.com/google/sanitizers/wiki/AddressSanitizer) [UndefinedBehaviorSanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) [MemorySanitizer](https://github.com/google/sanitizers/wiki/MemorySanitizer)

Do you mean to say that you want MapServer to support different fuzzers and different sanitizers?
If yes, how is a user supposed to select a fuzzer or sanitizer? Just removing all the (hard-coded) flags without documenting how one should do that selection doesn't look like a good idea.

Your proposed changes may or may not be good, I don't know because I don't understand your goal and I don't understand your explanations.

> Well, it's simple. Target under test, The inbuild library for the target under-test needs to be static, not shared. If the inbuilt library is shared, then the fuzzing engine can not get any coverage report for fuzzing. If the fuzzing engine does not get a good coverage report, then there is no good malformed input, and not any interesting bug can be found.

So this is very specific to the environment where you want to run the fuzzers, but it's not a generic problem, it doesn't solve anybody's problems but only yours. I don't think MapServer's build files should hard-code the flags for your environment. Instead, you should document how to pass the needed flags to cmake.

And really, you should write better commit messages; and only do one thing in one commit. If you have a commit message like "add map fuzzer & switch fuzzer library linker to static" then it should be obvious that this should be 2 commits, not one. And the commit "switch fuzzer library linker to static" should be properly documented, the commit message should explain why.

rouault:
I believe I understand the root of the misunerstanding between @0x34d and @MaxKellermann .
If I understand well, the original work by @MaxKellermann resulted in a standalone fuzzer binary using libFuzzer integrated with CLang, wheras this pull request modifies the build so that it uses OSS-Fuzz infrastructure.

My take (and experience with ossfuzz integration in other projects like GDAL or PROJ) is that building the fuzzer binaries for ossfuzz is best done in a dedicated build.sh script (like the one in https://github.com/google/oss-fuzz/pull/8300), be it hosted in mapserver or the ossfuzz repository.
For now, regarding Mapserver own's CMake scripts, I'd say we could just keep the approach of @MaxKellermann for people wanting to have fuzzer outside of ossfuzz infra.

Static linking to MapServer makes it slightly easier with ossfuzz integration. Cf https://google.github.io/oss-fuzz/getting-started/new-project-guide/#static-and-dynamic-linking-of-libraries. I believe it can also works with shared lib, but libmapserver.so.* would need to be explicitly copied to the $OUT directory

MaxKellermann:
It is a good idea to make the build scripts flexible enough so they can be built with different fuzzers. Just this PR's approach isn't good - it just removes the flags, requiring the user to figure everything out from scratch.

Same for switching to static libraries - if ossfuzz requires static linking, just pass the required flags to cmake from the ossfuzz build script.

I think the ossfuzz build script should live in the MapServer repository, because MapServer may add new build options eventually, and having the build script in this repository allows keeping the build script in sync with MapServer's build options.

rouault:
superseded per https://github.com/MapServer/MapServer/pull/6636

Issue #6601, opened 2022-08-26
Variable !BOX! not being recognized.


PostholerCom:
I've installed 8rc1 and I've run into an issue. I didn't see anything in the change log, so I'll put it here.

The !BOX! variable that has been historically available doesn't seem to be supported with CONNECTIONTYPE POSTGIS any longer. For instance:

```
   DATA "geom from (
      select
         geom
      from sometable
      where st_intersects(geom, !BOX!)
   ) as sq using unique geom using srid=4269"
```

Even if I comment the line out I get an error, such as:
--where st_intersects(geom, !BOX!)

If it exists any where in the DATA block, it fails.

Here's the ms error log:

[Thu Aug 25 13:46:55 2022].451955 msPostGISBuildSQL(): General error message. Failed to build SQL 'where'.
[Thu Aug 25 13:46:55 2022].451980 msPostGISLayerWhichShapes(): Query error. Failed to build query SQL.
[Thu Aug 25 13:46:55 2022].452010 msDrawMap(): Image handling error. Failed to draw layer named 'wptByTrailType'.

...http response:

<?xml version='1.0' encoding="UTF-8" standalone="no" ?>
<!DOCTYPE ServiceExceptionReport SYSTEM ["http://schemas.opengis.net/wms/1.1.1/exception_1_1_1.dtd"](http://schemas.opengis.net/wms/1.1.1/exception_1_1_1.dtd)>
<ServiceExceptionReport version="1.1.1">
<ServiceException>
msDrawMap(): Image handling error. Failed to draw layer named &#39;wptsAll&#39;.
msPostGISLayerWhichShapes(): Query error. Failed to build query SQL.
msPostGISBuildSQL(): General error message. Failed to build SQL &#39;where&#39;.
</ServiceException>
</ServiceExceptionReport>

If I remove the !BOX! it will work.

Thanks!
Scott

jmckenna:
@PostholerCom the trick for debugging PG issues is always getting the full query sent from MapServer to PostGIS (once you get that full query, then you can test it with the `psql` commandline.  You can follow the precise steps for this at https://mapserver.org/optimization/vector.html#debugging-speed-issues-with-postgis (I had tried to slowly explain my steps there)

Or, you can also just use the `map2img` command with the `-all_debug 5` switch, to get the full msPostGISLayerWhichShapes query.  (this is what Steve was asking for in the email thread)

PostholerCom:
[...]

PR #6602, opened 2022-08-29
Fix for #6601 - ensure a valid WHERE clause is created in all cases

geographika:
Return a non-empty string, so in cases where `!BOX!` is used a valid WHERE clause is still created. See #6601. 

The logic currently relies on returning empty strings when invalid WHERE clauses are created to throw an error, so the least intrusive fix is to create a "fake" where clause to return all records and ensure the check for an empty string is skipped.


sdlime:
Thanks for taking this on @geographika! Fix seems ok to me at first glance...

PostholerCom:
Thank you for looking into this!

jmckenna:
I've added an msautotest for the boxtoken for PostGIS, let's see how it goes...

jmckenna:
darn, it's likely failing because of libpng version differences (for the msautotest)...will have to tackle tomorrow...

geographika:
> darn, it's likely failing because of libpng version differences (for the msautotest)...will have to tackle tomorrow...

I was wondering how you generated msautotest images on Windows (or Mac?). A WFS test could be easier?

jmckenna (approved):

jmckenna:
thanks @geographika & @PostholerCom !

PR #6603, opened 2022-08-29
Check if a LAYER has a NAME before accessing the property

geographika:
See discussions at https://lists.osgeo.org/pipermail/mapserver-dev/2022-August/016847.html

> It appears that as of 8.0 a layer must have a NAME. In the past it has been
possible for a layer not to have a NAME if it had a  GROUP.

> with a WMS request. Apache throws a 500 error "End of script
output before headers". I define a LAYER with a GROUP but no NAME

This update  ensures the same behaviour as 7.6 and avoids crashing the server.


jmckenna:
thanks @geographika !

PR #6606, opened 2022-09-01
update tests for PHP 8.1.10 release

jmckenna:
- this will fail today, until binaries are uploaded...


jmckenna:
/me yikes, I always seem to get the 666 tickets in every repository! 😈

PR #6615, opened 2022-09-14
update security policy to mention 8.0 release

jmckenna:


sdlime:
Looks good to me. I wish we could reduce the support window for 7.6 to just 2 years (or even 1) but that seemed rather unpopular... --Steve

sdlime:
Oops, didn't mean to close.

Issue #6613, opened 2022-09-14
pagination with ogcapi doesnt work in demo server


landryb:
https://demo.mapserver.org/cgi-bin/mapserv/localdemo/ogcapi/collections/countries/items?limit=10 and https://demo.mapserver.org/cgi-bin/mapserv/localdemo/ogcapi/collections/countries/items?f=html&limit=10&offset=10 give the exact same features..
## Expected behavior and actual behavior.

the features returned should properly paginate on the complete dataset..
and the same problem seems to affect json output ?

eg https://demo.mapserver.org/cgi-bin/mapserv/localdemo/ogcapi/collections/countries/items?f=json&limit=10 & https://demo.mapserver.org/cgi-bin/mapserv/localdemo/ogcapi/collections/countries/items?f=json&limit=10&offset=10 give the same result.

i don't know what dataset type is used for countries, but it smells fishy. And the problem is the same for other collections on the demo service.

@sdlime @jmckenna ?

## MapServer version and installation method

well ms4w head i guess..

sdlime:
Jeff, were can the demo service files be found?

jmckenna:
I likely use old templates (I assumed this problem might happen).  I'll update the templates, from the latest ones in the `main` branch.  @landryb does this problem occur for your own local service?

sdlime:
It looks like the offset param has no affect - you can just try with JSON (so independent of the templates) and you always get the same set of results back. It certainly worked at some point. @jmckenna or @landryb, can you confirm what I'm seeing?

geographika:
@sdlime - I can confirm on my side - same results no matter what the offset parameter.

jmckenna:
I also confirm locally here.  (it seems to not be just a demo issue) @landryb can you confirm this issue locally?

sdlime:
@landryb did in the original ticket. I missed that at first.

jmckenna:
ok, so to summarize: @sdlime @geographika and myself can duplicate this locally with main, using our own individual service, but @landryb cannot reproduce this with `main`, just with demo, correct?

landryb:
sorry havent had time to try reproducing locally with 8.0.0, i've seen it on demo but i havent checked yet other instances. I'm also pretty sure the offset parameter worked at some point. Might be dependent on dataset type being used ?

landryb:
[...]

PR #6617, opened 2022-09-15
Set layer startindex for paging in OGC Features API

geographika:
Possible fix for #6613. 

The database drivers use `lp->startindex` for generating queries with `rownnum` for paging. E.g.

https://github.com/MapServer/MapServer/blob/6c3486b3577d7efc6988d4e319df45807419d59f/mapmssql2008.c#L1577

This doesn't appear to be set when using the OGC Features API, only `map->query.startindex` is set. 

Paging in WFS works by setting `lpQueried->startindex`. 

https://github.com/MapServer/MapServer/blob/e789297546c00d60f6aad3fb51e53cc4b2dd8e35/mapwfs.cpp#L2961

Testing locally with MSSQL this correctly sets the `rownum` based on the offset. I'm unsure of any possible side effects of this change though. 
An `offset=20` produces a SQL query with `[rownum] >= 21 and [rownum] < 31` - is this the correct? Or should the `1+ offset` be changed to simply `offset`?

### DIFF
--- a/mapogcapi.cpp
+++ b/mapogcapi.cpp
@@ -1058,9 +1058,10 @@ static int processCollectionItemsRequest(mapObj *map, cgiRequestObj *request, co
               return MS_SUCCESS;
             }
 
-            // msExecuteQuery() use a 1-based offst convention, whereas the API
+            // msExecuteQuery() use a 1-based offset convention, whereas the API
             // uses a 0-based offset convention.
             map->query.startindex = 1 + offset;
+            layer->startindex = 1 + offset;
         }
 
         if(msExecuteQuery(map) != MS_SUCCESS) {


sdlime:
Is WFS paging still working? I swear this was working previously "as is" - but maybe not.

jmckenna:
My demo uses SpatiaLite db, so Seth is on track.  (will test this locally this morning)

jmckenna:
@sdlime that is exactly my morning plan: test with shp, test with this PR, and come back here with feedback.

jmckenna:
Good news: the change in this pull request seems to solve the offset issue: see 
https://demo.mapserver.org/cgi-bin/mapserv/localdemo/ogcapi/collections/countries/items?f=html&limit=10&offset=10 

versus
https://demo.mapserver.org/cgi-bin/mapserv/localdemo/ogcapi/collections/countries/items?limit=10

(my delay was me rebuilding and testing on several servers)

jmckenna:
thanks @geographika !

PR #6640, opened 2022-09-30
Add a github action CI-Fuzz workflow

rouault:
CI-Fuzz triggers fast OSS-Fuzz checks on pull requests. See https://google.github.io/oss-fuzz/getting-started/continuous-integration/ for more details

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,24 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


rouault:
CC @sdlime  This (adding ci-fuzz as a github action workflow) should be useful for non-regression of fuzzer targets, as well as Mapserver operational code itself

https://github.com/MapServer/MapServer/blob/main/fuzzers/README.md contains general instructions regarding fuzzing

sdlime:
That's really pretty cool. So it's not fuzzing the pull request per se - at least at this point w/o code coverage - but making sure you haven't gone backwards. +1 on integration from me.

PR #6644, opened 2022-10-02
mapshape.c: avoid integer overflow in msSHPOpenVirtualFile()

rouault:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=51532

### DIFF
--- a/mapshape.c
+++ b/mapshape.c
@@ -229,10 +229,22 @@ SHPHandle msSHPOpenVirtualFile( VSILFILE * fpSHP, VSILFILE * fpSHX )
     return( NULL );
   }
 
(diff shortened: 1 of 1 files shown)


sdlime:
Another 7.6 backport?

rouault:
> Actually most of these fuzz fixes should be I'd think.

Adding the 7.6 backport label, but I just want to point there's always a trade-off of the benefit vs the risk of regression (I can remember having sometimes introduced regressions among the hundred of ossfuzz related bug fixes I did in GDAL). I'm not sure there's a real security issue here, but the fix should be relative safe

sdlime:
Understood, good point. The ones tagged as security are probably the only ones to consider then... --Steve

Issue #6626, opened 2022-09-22
MapScript msLoadMapFromString doesn't accept a configObj


geographika:
Attempting to load a Mapfile from a string which references plugins will fail, as since MapServer 8.0 plugins are referenced by keys rather than paths.

Errors such as the following will be logged: `msGetSymbol(): General error message. Can't load requested dynamic library: mssql`

I think allowing a configObj to be passed to the function and then adding: `map->config = config; // create a read-only reference`  in https://github.com/MapServer/MapServer/blob/35a52a48519e770da4b2a8f4b1af6fbdba03d950/mapfile.c#L6151 should fix this.

geographika:
Fixed with #6646

PR #6646, opened 2022-10-03
Allow a CONFIG to be used when loading a Mapfile from a string

geographika:
Update to address #6626. 

[Python ](https://github.com/MapServer/MapServer/blob/545b44a4924505c97d2192b042091cc38972fc5f/mapscript/python/pyextend.i#L24)and C# MapScript both allow loading a Mapfile from a string. 

MapServer 8.0 now uses a `CONFIG` file that allows setting keys for plugins and other settings.  This update allows a config file to be used when loading the Mapfile from a sting:

`mapscript.fromstring("MAP NAME TEST END", r"C:/MapfileResources/", config_file)` 

I've not tested if the C# approach works - the following is reported in the SWIG output, although I'm unsure if this is relevant:

```
 [  6%] Swig compile ../mapscript.i for csharp
/home/runner/work/mapserver/mapserver/mapscript/v8/../swiginc/map.i:62: Warning 516: Overloaded method mapObj::mapObj(char *,int,configObj *) ignored,
/home/runner/work/mapserver/mapserver/mapscript/v8/../swiginc/map.i:57: Warning 516: using mapObj::mapObj(char *,int) instead.
```

### DIFF
--- a/mapfile.c
+++ b/mapfile.c
@@ -6148,7 +6148,7 @@ static bool msGetCWD(char* szBuffer, size_t nBufferSize, const char* pszFunction
 /*
 ** Sets up string-based mapfile loading and calls loadMapInternal to do the work.
 */
-mapObj *msLoadMapFromString(char *buffer, char *new_mappath)
+mapObj *msLoadMapFromString(char *buffer, char *new_mappath, const configObj *config)
 {
   mapObj *map;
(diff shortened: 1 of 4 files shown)


sdlime:
I think this look right - matches the load from file workflow.

geographika:
The last update should allow the C# MapScript bindings to also set a `CONFIG` file.

PR #6684, opened 2022-10-06
maplexer.l: avoid non-null terminated msyystring_buffer that can cause read heap-buffer-overflow

rouault:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=52175

### DIFF
--- a/maplexer.c
+++ b/maplexer.c
@@ -1,6 +1,6 @@
-#line 2 "/home/bjorn/code/MapServer/maplexer.c"
+#line 2 "/home/even/mapserver/mapserver/maplexer.c"
 
-#line 4 "/home/bjorn/code/MapServer/maplexer.c"
+#line 4 "/home/even/mapserver/mapserver/maplexer.c"
 
 #define  YY_INT_ALIGNED short int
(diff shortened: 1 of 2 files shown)


sdlime:
Won't we have the same problem then with getDouble or anywhere else we use "Parsing error near (%s)"?

rouault:
@sdlime On closer investigation, the issue was in maplexer.l where msyystring_buffer was not always null-terminated

sdlime:
Ah... Might still be good to consider trimming msyystring_buffer when using outputting the location where the error was encountered. Either the whole string if the buffer is less than 20 characters or the first 17 characters followed by '...'. Otherwise an unterminated quoted string mistake could cough up more text than necessary.  What do you think?

rouault:
> What do you think?

potential improvement, but not in my todo list currently :-)

sdlime:
> not in my todo list currently :-)

Absolutely!

rouault:
manually backported to 7.6 per https://github.com/MapServer/MapServer/pull/6688

PR #6692, opened 2022-10-08
loadFeaturePoints(): fix memleak in error code path

rouault:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=52222


jmckenna:
thanks @rouault !

PR #6693, opened 2022-10-08
msOGRShapeFromWKT(): fix memleak in error code path

rouault:
Likely fix for https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=52223 (couldn't reproduce the leak with the reproducer and current main, but this commit does fix a leak)


jmckenna:
thanks @rouault !

PR #6747, opened 2022-11-02
freeLabelLeader(): fix memleak (including in nominal code path)

rouault:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=52947


jmckenna:
thanks @rouault !

PR #6753, opened 2022-11-12
Bit manipulation: avoid undefined-shift behavior when left-shifting of 31 bits

rouault:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=53339


geographika:
@rouault - looks like Appveyor has updated to Python 3.11 which means some dependencies are missing. I'll take a look.

geographika:
@rouault #6754 should hopefully resolve the Appveyor issue.

PR #6754, opened 2022-11-12
Ignore Python 3.11 on Appveyor

geographika:
When searching for Python versions 3.10 and 3.11 are picked up instead of 2.7 on Appveyor. Quick fix for now is to ignore these two Python versions until they are supported by MapScript and/or 2.7 is deprecated.


rouault:
manually backport to branch-8-0 per aa45522847418edb1bdf10636cea989a6cce38bf

Issue #6758, opened 2022-11-13
Add WIN32 flag to SWIG csharp command line


szekerest:
As of the cmake build, the WIN32 flag for the SWIG command line hasn't been added, though it was present in the nmake based builds.
This results in the GEOS exceptions to cause app crashes like this:

```
	Exception #:  0XE06D7363
	Stack:        
		KERNELBASE!RaiseException+0x68
		VCRUNTIME140!CxxThrowException+0xad
		geos!geos::operation::buffer::BufferBuilder::bufferLineSingleSided+0xe9a
		geos_c!GEOSOffsetCurve_r+0x10e
		mapserver!msGEOSOffsetCurve+0x70 [e:\builds-t1\src\mapserver-intramaps\mapgeos.c @ 807]
		mapserver!msOffsetCurve+0x24 [e:\builds-t1\src\mapserver-intramaps\maputil.c @ 1764]
		mapserver!msDrawLineSymbol+0x18d [e:\builds-t1\src\mapserver-intramaps\maprendering.c @ 601]
		mapserver!msDrawShadeSymbol+0xc5 [e:\builds-t1\src\mapserver-intramaps\maprendering.c @ 705]
		mapserver!polygonLayerDrawShape+0xdd [e:\builds-t1\src\mapserver-intramaps\mapdraw.c @ 1928]
		mapserver!msDrawShape+0x7b5 [e:\builds-t1\src\mapserver-intramaps\mapdraw.c @ 2173]
		mapserver!msDrawVectorLayer+0x804 [e:\builds-t1\src\mapserver-intramaps\mapdraw.c @ 1175]
		mapserver!msDrawLayer+0x3a3 [e:\builds-t1\src\mapserver-intramaps\mapdraw.c @ 813]
		mapserver!msDrawMap+0x415 [e:\builds-t1\src\mapserver-intramaps\mapdraw.c @ 403]
		mapscript!CSharp_mapObj_draw+0xd
```

and this

```
Exception #:  0XE06D7363
	Stack:        
		KERNELBASE!RaiseException+0x68
		VCRUNTIME140!CxxThrowException+0xad
		geos!geos::operation::buffer::BufferBuilder::bufferLineSingleSided+0xe9a
		geos_c!GEOSOffsetCurve_r+0x10e
		mapserver!msGEOSOffsetCurve+0x70 [e:\builds-t1\src\mapserver-intramaps\mapgeos.c @ 807]
[...]

PR #6759, opened 2022-11-13
Add WIN32 flag to SWIG csharp command line, prevent classObj_drawLegendIcon to cause crash (#6758)

szekerest:


mapscript/swiginc/class.i
```diff
@@ -169,8 +169,17 @@
     }
     else
       layer->scalefactor = map->resolution/map->defresolution;
-    
+
+  #if defined(WIN32) && defined(SWIGCSHARP)
+    __try {
+        return msDrawLegendIcon(map, layer, self, width, height, dstImage, dstX, dstY, MS_TRUE, NULL);
```
rouault, line 175: this looks like a workaround than a proper fix. Couldn't the exception be caught upwards?
But if you want to keep this, why not just using the _WIN32 canoncial symbol ? This would avoid defining WIN32 explicitly

szekerest: Similar workaround already exists in map.i that have been working for the client who has contributed that.
Using _WIN32 might also work, I haven't tested that. Also I didn't check who else is relying on WIN32 actually (it has been enabled in the nmake based builds for years).

rouault: > Similar workaround already exists in map.i that have been working for the client who has contributed that.

hum, it would be good to understand what happens. It seems to me that the workaround, if really needed, should be moved to msGEOSOffsetCurve() itself, as I presume it affects the CGI mapserv too. Is this is a bug in GEOS (but the C functions of GEOS are supposed to catch C++ exceptions) ? or some issue with binaries not being compiled with consistent runtimes ?

szekerest: Looks like BufferBuilder::bufferLineSingleSided in geos throws util::IllegalArgumentException. I'm not sure who is responsible to catch this, but it brings down the entire IIS host process if that is not catched.

rouault: GEOSOffsetCurve_r() (https://github.com/libgeos/geos/blob/7a10a9f1857f8ba0231f5d3a9022e5d959b89c50/capi/geos_ts_c.cpp#L1168) calls execute() which catches all C++ exceptions: https://github.com/libgeos/geos/blob/7a10a9f1857f8ba0231f5d3a9022e5d959b89c50/capi/geos_ts_c.cpp#L424
So C++ exceptions shouldn't reach mapserver

szekerest: This looks like the same in my geos version. It is strange why the exceptions are propagated in spite of that.

szekerest: It looks like a SEH (OS) exception is being generated with ex. code = 0XE06D7363 (I'm not quite sure why). But the default /EHsc option makes that it is not being catched by the C++ exception handlers.

jmckenna:
this looks good to me as-is.  Shall we merge this now?  Will give a few days for more feedback.

szekerest:
As far as I see it in [this article](http://web.archive.org/web/20070501045100/http://support.microsoft.com/kb/185294) each C++ exceptions are generated with the code 0xE06D7363 and to properly handle the error, you need to wrap the offending code in a try...catch block. However it is still a question why geos doesn't handle this within the library.

Issue #6751, opened 2022-11-08
GetFeature PropertyIsLike doesn't allow filtering on parentheses


kad-dirc:
## Expected behavior and actual behavior

We have a GetFeature wfs request and want to filter on a string which contains parentheses, e.g. `Pand in gebruik (niet ingemeten)`.
Using the PropertyIsLike filter we expect to filter this feature with `*(niet ingemeten)`. 
But it fails to find the feature. 

PropertyIsLike snippet:
```xml
<PropertyIsLike escapeChar="[" singleChar="." wildCard="*" matchCase="true">
    <ValueReference>status</ValueReference>
    <Literal>* (niet ingemeten)</Literal>
</PropertyIsLike>
```

With debug log enabled we see that Mapserver converts the request to an sql stamenent and escapes these parentheses, e.g.
`... FROM "pand" WHERE ("pand"."status" GLOB '*\(niet ingemeten\)')`

Hence it doesn't find anything.

This makes it impossible to filter on strings containing parentheses. I would consider this as a bug, since the escaped parentheses in the sql doesn't seem to be of any use to me. 
Or am I wrong?

## Steps to reproduce

See zip file for a simple gpkg and mapfile config.

```shell
# Start a mapserver
docker run --rm -v `pwd`:/srv/data -e MS_MAPFILE=/srv/data/config/mapfile/service.map -p 8080:80 --name mapserver pdok/mapserver

# GetFeature request which fails (but should succeed)
curl --location --request POST 'localhost:8080/lv/bag/wfs/v2_0' \
--header 'Content-Type: application/xml' \
--data-raw '<GetFeature xmlns="http://www.opengis.net/wfs/2.0" service="WFS" version="2.0.0" outputFormat="application/json" startIndex="0" count="1000"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/wfs/2.0 http://schemas.opengis.net/wfs/2.0/wfs.xsd">
    <Query typeNames="bag:pand" srsName="EPSG:28992">
        <Filter xmlns="http://www.opengis.net/fes/2.0">
            <PropertyIsLike escapeChar="[" singleChar="." wildCard="*" matchCase="true">
                <ValueReference>status</ValueReference>
                <Literal>* (niet ingemeten)</Literal>
            </PropertyIsLike>
        </Filter>
    </Query>
</GetFeature>'
# Corresponding SQL snippet (from mapserver debug log)
[...]

PR #6763, opened 2022-11-14
MS_TOKEN_COMPARISON_LIKE is never generated by maplexer.l. Remove any uses of it

rouault:
CC @sdlime  I figured this when investigating #6751. I'm not sure if it was used at one point and then removed, or just never got implemented. Anyway it seems good policy to remove this dead code. To preserve the API/ABI, I renamed the enum value to MS_TOKEN_COMPARISON_LIKE_unused for mapserver builds, and keep the original MS_TOKEN_COMPARISON_LIKE name for potential (unlikely) external users


sdlime:
Pretty sure it just never implemented, I had found a wildcard string match implementation at one point (e.g. *foo*) but never got around to it for whatever reason. Your fix seems just fine.

--Steve

sdlime:
Probably not necessary but do you think it's worth backporting to branch-8-0?

rouault:
> Probably not necessary but do you think it's worth backporting to branch-8-0?

no, doesn't seem necessary

PR #6775, opened 2022-12-15
Fix usage of FILE* and VSILFILE*

rouault:
In GDAL 3.7dev, VSILFILE* is no longer an alias for FILE* (see https://github.com/OSGeo/gdal/pull/6883), so we have to adjust our usage to be correct.

This MapServer change is backward compatible with GDAL < 3.7 too

CC @szekerest


rouault:
appveyor build failure due to GDAL not being found in gisinternals package, which is likely caused by the issue we are trying to fix with this fix :-)

PR #6780, opened 2022-12-30
Fix information disclosure and denial of service related to CONTEXT= loading

### LINKED ISSUE #6779: Information disclosure at context parameter
# Steps to reproduce the problem.

see the output of these 2 requests, it gives the information which file is on the system and which is not

https://demo.mapserver.org/cgi-bin/wms?context=/etc/passwd

https://demo.mapserver.org/cgi-bin/wms?context=/etc/passwd2

# Expected behavior

there should be an env paramenter like MS_MAP_PATTERN for context files

rouault:
Fixes #6779

### DIFF
--- a/.github/workflows/start.sh
+++ b/.github/workflows/start.sh
@@ -121,6 +121,7 @@ export PATH=/tmp/install-mapserver/bin:$PATH
 # Demonstrate that mapserv will error out if cannot find config file
 mapserv 2>&1  | grep "msLoadConfig(): Unable to access file" >/dev/null && echo yes
 mapserv QUERY_STRING="MAP=wfs_simple.map&REQUEST=GetCapabilities" 2>&1  | grep "msLoadConfig(): Unable to access file" >/dev/null && echo yes
+mapserv QUERY_STRING="map=ows_context.map&CONTEXT=ows_context.xml&SERVICE=WMS&VERSION=1.1.1&REQUEST=GetCapabilities"  2>&1  | grep "msLoadConfig(): Unable to access file" >/dev/null && echo "Check that we can't load a OWS context file if MS_CONTEXT_PATTERN is not defined: yes"
 
 echo "Check that MS_MAP_NO_PATH works"
 cat <<EOF >/tmp/mapserver.conf
--- a/mapcontext.c
+++ b/mapcontext.c
@@ -664,7 +664,7 @@ int msLoadMapContextLayerDimension(CPLXMLNode *psDimension, layerObj *layer)
 */
 int msLoadMapContextGeneral(mapObj *map, CPLXMLNode *psGeneral,
                             CPLXMLNode *psMapContext, int nVersion,
-                            char *filename)
+                            const char *filename)
 {
 
   char *pszProj=NULL;
@@ -809,7 +809,7 @@ int msLoadMapContextGeneral(mapObj *map, CPLXMLNode *psGeneral,
 ** Load a Layer block from a MapContext document
 */
 int msLoadMapContextLayer(mapObj *map, CPLXMLNode *psLayer, int nVersion,
-                          char *filename, int unique_layer_names)
+                          const char *filename, int unique_layer_names)
 {
   char *pszValue;
   const char *pszHash;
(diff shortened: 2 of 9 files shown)


sdlime:
Could also add support for context file aliases in the config file (for 8.x and up). I was wondering if just normalizing the error message ("invalid context file") for both of these cases might be a quick fix.

rouault:
> Could also add support for context file aliases in the config file (for 8.x and up)

I'm always a bit nervous by this functionality. It allows user to refer to CONTEXT=EVIL_ENVIRONMENT_VARIABLE where they would have managed to define (or use an existing) EVIL_ENVIRONMENT_VARIABLE in the webserver.

> I was wondering if just normalizing the error message ("invalid context file") for both of these cases might be a quick fix.

I don't understand which exact error message you are referring to

sdlime:
> > Could also add support for context file aliases in the config file (for 8.x and up)
> 
> I'm always a bit nervous by this functionality. It allows user to refer to CONTEXT=EVIL_ENVIRONMENT_VARIABLE where they would have managed to define (or use an existing) EVIL_ENVIRONMENT_VARIABLE in the webserver.

The aliases would be stored in the config file, perhaps in a CONTEXT block. No environment variables would be used. Matching the setup for mapfiles more broadly might make sense.

> > I was wondering if just normalizing the error message ("invalid context file") for both of these cases might be a quick fix.
> 
> I don't understand which exact error message you are referring to

The two sample URLs produce:

- msLoadMapContext(): WMS Map Context error. Unable to read /etc/passwd2 msGetMapContextFileText(): Unable to access file. (/etc/passwd2)
- msLoadMapContext(): WMS Map Context error. Not a Map Context file (/etc/passwd)

If those instead both just said "Invalid context file" and logged the more detailed message then that would fix the immediate disclosure issue.

--Steve

rouault:
> The aliases would be stored in the config file, perhaps in a CONTEXT block.

I'll probably let you tackle this if you wish. Not sure how broadly used this functionality is and if it is worth investing too much in it.

> If those instead both just said "Invalid context file" and logged the more detailed message then that would fix the immediate disclosure issue.

I believe this isn't enough: the issue is more severe than just the information disclosure. It is also close to be a denial of service, as users could point to any arbitrary file on the file system and if it is large enough, load up to 2 GB of content in RAM, potentially exhausting resources of the server

rouault:
> I believe this isn't enough:

I've also just found a particular value of CONTEXT= that could stall the request forever

sdlime:
I've never used them personally, not even sure how to enable TBH. It must not be available by default - I've not been able to trigger the problem against just any mapfile. @jmckenna and @tomkralidis might know a bit more?

Could also force a check on the file extension, Map files must end in .map and there's a built-in pattern for that check. Could do the same here with .xml.

rouault:
> It must not be available by default

well, it is. See the example I added in https://github.com/MapServer/MapServer/pull/6780/files#diff-9b8bde2a19d8f17bc9f84c4689d466f7e207eba91973bca7cfe0a02d671769e2R7

sdlime:
> > It must not be available by default
> 
> well, it is. 

Confirmed. It's not a fatal error so when you might see the error is confusing. Marking the mapfile as immutable (e.g. VALIDATION "immutable" "true" END) bypasses processing a context file. So would probably want:

  MS_CONTEXT_PATTERN
  MS_CONTEXT_BAD_PATTERN
  MS_CONTEXTFILE_PATTERN (rarely, if ever, set)
  MS_DEFAULT_CONTEXTFILE_PATTERN (probably something like "\\.[Xx][Mm][Ll]$")

The latter two would mirror what is done in msLoadMap() for mapfiles https://github.com/MapServer/MapServer/blob/02eb9727f644c6fc4564743e655a248128f75903/mapfile.c#L6320. We could add context alias support to config files at a later date (e.g. MS_CONTEXT_NO_PATH).

rouault:
> MS_CONTEXTFILE_PATTERN (rarely, if ever, set) MS_DEFAULT_CONTEXTFILE_PATTERN (probably something like "\.[Xx][Mm][Ll]$")

added in last commit

> MS_CONTEXT_PATTERN
> MS_CONTEXT_BAD_PATTERN

ok, renamed MS_MAP_CONTEXT_* to MS_CONTEXT_* in reworked initial commit

sdlime:
I guess we'll need to backport using getenv()... or will the CPL functions work as is?

rouault:
> or will the CPL functions work as is?

CPLGetConfigOption() fallbacks to getenv(): https://github.com/OSGeo/gdal/blob/57aabd225d4b948433e986c26f2a4fdedca61331/port/cpl_conv.cpp#L1722

tomkralidis (approved):

sdlime:
This makes sense to me. Would make sense to get a few more eyes on this since it will be a bit of a regression. I can't imagine it's used that widely but I may be wrong.

sdlime:
I approve changes + backporting.

sdlime (approved):

PR #6781, opened 2023-01-01
support HTTP_MAPSERVER_CONFIG_FILE for suExec environment

boiledorange73:
For suExec environment, also HTTP_X_MAPSERVER_CONFIG_FILE ,which starts with "HTTP_X_", could be accepted.

### DIFF
--- a/mapserv-config.cpp
+++ b/mapserv-config.cpp
@@ -122,10 +122,10 @@ configObj *msLoadConfig(const char* ms_config_file)
     // get config filename from environment
     ms_config_file = getenv("MAPSERVER_CONFIG_FILE");
     // If MAPSERVER_CONFIG_FILE not set,
-    // Tries to use HTTP_X_MAPSERVER_CONFIG_FILE
-    // (under suExec environment, SetEnv can affect to only HTTP_X_*)
+    // Tries to use HTTP_MAPSERVER_CONFIG_FILE
+    // (under suExec environment, SetEnv passes only HTTP_* or SSL_*)
     if (ms_config_file == NULL) {
-      ms_config_file = getenv("HTTP_X_MAPSERVER_CONFIG_FILE");
+      ms_config_file = getenv("HTTP_MAPSERVER_CONFIG_FILE");
     }
   }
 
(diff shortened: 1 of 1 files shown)


sdlime:
Does a suEXEC environment not support just a plain environment var named MAPSERVER_CONFIG_FILE? Otherwise this seems like a relatively benign add. Would also need a documentation update.

boiledorange73:
I mistook ! Sorry ...

Envs starting with "HTTP_" or "SSL_" passed, NOT only "HTTP_X_".

It is not wrong that suExec filters environment variables.
https://github.com/omnigroup/Apache/blob/master/httpd/support/suexec.c
(or https://groups.google.com/g/comp.infosystems.www.authoring.cgi/c/hL8D-2uug7Q )

jmckenna:
This seems fair (and I volunteer to add a note in the docs for this).  Sound good @sdlime ?

geographika:
Does the CGI wrapper script at https://mapserver.org/cgi/wrapper.html still work well for MapServer? It sounds like that might be a better approach for handling this?

See also "Option 3: Create a wrapper script" at https://help.dreamhost.com/hc/en-us/articles/215364578-suEXEC-overview

geographika:
I'd be in favour of closing this. The `HTTP_` prefix indicates that the variable comes from a request header, which is not the case for
`MAPSERVER_CONFIG_FILE`, and could potentially be sent by a client if there are no restrictions on the web server. 

The workaround seems to be a wrapper script which would set the `MAPSERVER_CONFIG_FILE` environment variable:

```
export MAPSERVER_CONFIG_FILE=/path/to/mapserver.config
/path/to/your/actual/cgi/mapserv
```

geographika:
@boiledorange73 - rather than modifying the MapServer codebase you should be able to use a wrapper script and set `MAPSERVER_CONFIG_FILE` there. 
There is an example (using Apache) of a wrapper script for MapServer at https://github.com/camptocamp/docker-mapserver/blob/master/runtime/usr/local/bin/mapserv_wrapper

PR #6785, opened 2023-01-04
Optimise WFS requests by reducing fields selected

geographika:
Similar to #6131 and #6131 - this change avoids selecting fields from a database which aren't required. 

As discussed in the comments at https://github.com/MapServer/MapServer/pull/6131#issuecomment-675750118

> any fields referenced in a CLASS are selected at the database level even if never used for a WFS request (this was also the case in > my previous pull request) - I presume there is no way to get round this due to buildLayerItemList

Simply setting the number of classes to 0 prior to retrieving features from the database avoids adding in all fields referenced in CLASSes in a LAYER. For example if a layer has a CLASS similar to below then `Myfield` is added to the database `SELECT` query (in [buildLayerItemList](https://github.com/MapServer/MapServer/blob/8d6c9ecdd29c75b3976788a0d287860a922145c3/maplayer.c#L795):

```
        CLASS
            NAME "Example"
            EXPRESSION ( [Myfield] = 1)
            STYLE
            ...
            END
        END
```

Setting `numclasses ` to 0 avoids adding any of these fields in, which aren't required by a WFS query. 

In a sample application for a layer with lots of classes these can greatly reduce the number of fields requested from the database ,which are then discarded, and can show some significant performance improvements. 

I can't see any side-effects of this change, but I may be missing something.

### DIFF
--- a/mapwfs.cpp
+++ b/mapwfs.cpp
@@ -2540,6 +2540,8 @@ this request. Check wfs/ows_enable_request settings.", "msWFSGetFeature()",
         layerObj *lp;
         lp = GET_LAYER(map, j);
         if (lp->status == MS_ON) {
+          // classes don't affect retrieval of features, so set the count to 0
+          // this avoids selecting any fields referenced in a CLASS from the data source
           lp->numclasses = 0;
           int status = msWFSRunBasicGetFeature(map, lp, paramsObj, nWFSVersion);
           if( status != MS_SUCCESS )
--- a/mapwfs.cpp
+++ b/mapwfs.cpp
@@ -2162,9 +2162,19 @@ static int msWFSRunFilter(mapObj* map,
 
     }
 
+    // ignore any classes used by the WFS layer to avoid selecting fields we don't need
+    int numclasses = lp->numclasses;
+    lp->numclasses = 0;
+
     /* run filter.  If no results are found, do not throw exception */
     /* this is a null result */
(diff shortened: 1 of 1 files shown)


mapwfs.cpp
```diff
@@ -2540,6 +2540,9 @@ this request. Check wfs/ows_enable_request settings.", "msWFSGetFeature()",
         layerObj *lp;
         lp = GET_LAYER(map, j);
         if (lp->status == MS_ON) {
+          // classes don't affect retrieval of features, so set the count to 0
+          // this avoids selecting any fields referenced in a CLASS from the data source
+          lp->numclasses = 0;
```
rouault, line 2545: Won't that cause memory leaks when the lp object is cleaned up later ? (do we have msautotest testing that scenario by the way? if not would be appropriate to edit an existing one) It would probably be safer to save lp->numclasses values before setting it to 0 and restoring the original value after the msWFSRunBasicGetFeature() call.

I assume this could also be done before the call to FLTApplyFilterToLayer() done at line 2167.
mapogcapi.cpp could also benefit from that.

geographika: Thanks for the review @rouault - much appreciated. 
There were no memory leaks as the layer cleanup code uses the `layer->maxclasses` value to loop through classes in the layer and free them rather than `lp->numclasses`. However it does seem safer to set `numclasses` back to the original value once the query is run. 

I've also added the same approach to `FLTApplyFilterToLayer` which also calls `buildLayerItemList`.

PR #6786, opened 2023-01-04
Optimise WMS requests by ignoring any fields that won't be used

geographika:
Similar to #6785 but for WMS requests. 

If a WMS request includes specific STYLES e.g. `&SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&FORMAT=image%2Fpng&TRANSPARENT=true&STYLES=MyStyle&TILED=false&LAYERS=MyLayer&` then MapServer sets the `CLASSGROUP` of the `LAYER` to the selected style. In this example `MyStyle`. 

The generic `buildLayerItemList` currently goes through all classes in a layer to find any references to fields (anything in [SquareBrackets]) and adds them to the SELECT clause sent to the datasource. 

This update checks if `CLASSGROUP` has been set, and if it has then ignores any `CLASS` that doesn't have a matching `GROUP` name. Fields in these unused classes aren't required and not selecting them can improve performance. This is most noticeable in layers which have many classes and groups. 

In my real-world test case the database SELECT query goes from selecting 21 fields to selecting 2, and due to the complex nature of the LAYER database view (from 2 seconds per view load to 0.1). 

Will see if any tests are broken by this change. GetFeatureInfo requests are ok as they always select all fields rather than building a list.

### DIFF
--- a/maplayer.c
+++ b/maplayer.c
@@ -809,7 +809,13 @@ static void buildLayerItemList(layerObj *layer)
 
   /* layer classes */
   for(i=0; i<layer->numclasses; i++) {
-    
+
+    // if classgroup has been set we can ignore any fields required
+    // for rendering unused classes
+    if (layer->class[i]->group && layer->classgroup &&
+        strcasecmp(layer->class[i]->group, layer->classgroup) != 0)
+        continue;
+
     if(layer->class[i]->expression.type == MS_EXPRESSION) /* class expression */
       msTokenizeExpression(&(layer->class[i]->expression), layer->items, &(layer->numitems));
 


sdlime:
So this is limited only to draws then? If so it's a nice optimization. Not sure how this would work with query/WFS (#6785) operations though since those are like getFeatureInfo requests.

geographika:
> So this is limited only to draws then? If so it's a nice optimization. Not sure how this would work with query/WFS (#6785) operations though since those are like getFeatureInfo requests.

For WMS `GetFeatureInfo` requests all fields are selected by setting the [get_all](https://github.com/MapServer/MapServer/blob/8d6c9ecdd29c75b3976788a0d287860a922145c3/maplayer.c#L888) parameter. WMS Legends seem fine too. 

I was expecting some msautotests to fail with this update so was surprised when they all passed. I thought there could be issues around the "painter's model" described in https://mapserver.org/development/rfc/ms-rfc-124.html where all classes can be used. Either there are few tests for this, or the CLASSGROUP of a LAYER isn't set. 

For complex layers with lots of different CLASS GROUPS it can make a big difference when using complex database views. 

For WFS queries I don't think any of fields in the CLASS are used, hence #6785, although there could be issues with client-sent SLD files?

I'm going to deploy on a test server to see if anyone notices any unforseen consequences.

geographika:
As discussed yesterday, this can be added to a 8.2 release and have some further testing in beta in case there are possible CGI calls that might be affected. However if `CLASSGROUP` is set or changed then the request shouldn't care about any other classes in the layer. 

Will merge in a couple of days into main if no objections.

rouault:
LGTM. Merging

Issue #6492, opened 2022-03-14
Wrong DescribeFeatureType result on boolean data types when jusing gml_types = auto


frodewj:
## Expected behavior and actual behavior.

When using a Postgre/PostGIS table as data source and "GML_TYPES" "auto" a DescribeFeatureType request should report boolean, not integer, on boolean data types.

Example datasource: 
![image](https://user-images.githubusercontent.com/3868499/158197399-2a614cbb-582a-455b-a3a7-efa8b42e312f.png)

Example mapfile:
![image](https://user-images.githubusercontent.com/3868499/158198953-bb70a211-b9c2-4912-a888-7c5102027739.png)

Example, result of a DescribeFeatureType request:
![image](https://user-images.githubusercontent.com/3868499/158197581-0d336851-cc91-4791-b9b6-8e2ba89822c0.png)

## Steps to reproduce the problem.

Test with a DescribeFeatureType request

## Attach simple test case (drag/drop it here)
Testdata.zip including mapfile, data and commands.
[testdata.zip](https://github.com/MapServer/MapServer/files/8254472/testdata.zip)

## Operating system
Tested on Windows 10 and Windows Server 2019.

## MapServer version and installation method
Mapserver 7.6.4 from Gisinternals (https://www.gisinternals.com/query.html?content=filelist&file=release-1930-x64-gdal-3-4-1-mapserver-7-6-4.zip)

geographika:
@frodewj - a workaround is to override just that field type, as per https://github.com/MapServer/MapServer/pull/6137

```
    METADATA
        "gml_include_items" "all"
        "gml_types" "auto"
        "gml_test_bool_type" "Boolean"
    END
```

Otherwise, the field to type conversion code is at https://github.com/MapServer/MapServer/blob/672de62a4e283a1cb6f21338dffe65bcdac65453/mappostgis.cpp#L2980

frodewj:
@geographika - thanks. 
Looks like the workaround can't be combined with "gml_types" "auto".
Using auto seems to override any stated field types, but we could skip using auto and specify each field type.

geographika:
@frodewj - looks like https://github.com/MapServer/MapServer/pull/6137 (which allows individual field type overrides) was not backported to 7.6.4, but will be available in the upcoming 8.0 release. You should be able to test in the dev releases from GISInternals.

jmckenna:
[...]

PR #6793, opened 2023-01-11
Set Boolean field type with PostGIS driver when using gml_types auto

geographika:
Fix for #6492. When using the PostGIS driver and:
```
    METADATA
        "gml_include_items" "all"
        "gml_types" "auto"
    END
```

Boolean fields are set as Integer. This pull request updates them to Boolean. As this code has been unchanged for ~10 years (#3570) I'm assuming it was an oversight?

The MSSQL driver sets Boolean GML Boolean type for bit fields (the MSSQL equivalent): https://github.com/MapServer/MapServer/blob/2b613f88307c3c71faac3e2438bf4b41b71f6b8c/mapmssql2008.c#L903

https://github.com/postgres/postgres/blob/c30f54ad732ca5c8762bb68bbe0f51de9137dd72/src/interfaces/ecpg/ecpglib/typename.c#L77 also converts `BOOLOID` fields to Boolean ([ecpg](https://www.postgresql.org/docs/current/app-ecpg.html) is the embedded SQL preprocessor for C programs).
GDAL also sets `BOOLOID` to Booleans - https://github.com/OSGeo/gdal/blob/a632e3093d6dd34cdcf979a5b2a20529fe2df943/ogr/ogrsf_frmts/pg/ogrpglayer.cpp#L2179

The pull request also includes tests for all WFS versions, and a new msautotest that tests all `"gml_types" "auto"` field conversions.


geographika:
Although the field storage for a Postgres Boolean is [1 bit](https://dba.stackexchange.com/a/246246/28311), the returned values can be:

> “true”, “false”, and a third state, “unknown”,

https://www.postgresql.org/docs/15/datatype-boolean.html

I've removed the `gml_width = '1';` to avoid setting a width. I believe this can affect Shapefile/DBF outputs.

sdlime:
> As this code has been unchanged for ~10 years (https://github.com/MapServer/MapServer/issues/3570) I'm assuming it was an oversight?

I guess we have to assume so - I doubt @warmerdam will recall...

PR #6794, opened 2023-01-14
Add Proj and GDAL versions to MapServer version output

geographika:
This pull request adds the version numbers of PROJ and GDAL to the version output for MapServer (using `mapserv -v` on the command line, and also present in some error messages). Both are required dependencies (at least since 8.0) so will always have a value. 

As different versions of PROJ and GDAL can have an impact on how MapServer behaves, and what is supported, it is useful to be able to easily access this information, and allow users to report it when having problems. 

The output will now look similar to the following:

```
MapServer version 8.1-dev PROJ version 7.2 GDAL version 3.4 OUTPUT=PNG 
OUTPUT=JPEG OUTPUT=KML SUPPORTS=PROJ SUPPORTS=AGG SUPPORTS=FREETYPE
SUPPORTS=CAIRO SUPPORTS=SVG_SYMBOLS SUPPORTS=SVGCAIRO SUPPORTS=ICONV 
SUPPORTS=FRIBIDI SUPPORTS=WMS_SERVER SUPPORTS=WMS_CLIENT SUPPORTS=WFS_SERVER
SUPPORTS=WFS_CLIENT SUPPORTS=WCS_SERVER SUPPORTS=SOS_SERVER
SUPPORTS=OGCAPI_SERVER SUPPORTS=FASTCGI SUPPORTS=THREADS
SUPPORTS=GEOS SUPPORTS=PBF INPUT=JPEG INPUT=POSTGIS
INPUT=OGR INPUT=GDAL INPUT=SHAPEFILE INPUT=FLATGEOBUF
```

Note the additional text - `PROJ version 7.2 GDAL version 3.4`. 

For production environments it is recommended that the `MS_NO_VERSION` environment variable is set to avoid these details being displayed to client applications:

https://github.com/MapServer/MapServer/blob/9209de5962ffc5287ab2c52f597b58f1751e2e1d/etc/mapserver-sample.conf#L42-L43

### DIFF
--- a/maperror.c
+++ b/maperror.c
@@ -538,9 +538,10 @@ char *msGetVersion()
 
   sprintf(version, "MapServer version %s", MS_VERSION);
 
-  static char projVersion[20];
-  sprintf(projVersion, " Proj version %d.%d", PROJ_VERSION_MAJOR, PROJ_VERSION_MINOR);
-  strcat(version, projVersion);
+  // add versions of required dependencies
+  static char PROJVersion[20];
+  sprintf(PROJVersion, " PROJ version %d.%d", PROJ_VERSION_MAJOR, PROJ_VERSION_MINOR);
+  strcat(version, PROJVersion);
 
   static char GDALVersion[20];
   sprintf(GDALVersion, " GDAL version %d.%d", GDAL_VERSION_MAJOR, GDAL_VERSION_MINOR);
--- a/maperror.c
+++ b/maperror.c
@@ -532,11 +532,19 @@ void msWriteErrorImage(mapObj *map, char *filename, int blank)
 
 char *msGetVersion()
 {
-  static char version[1024];
+  static char version[2048];
 
   if(CPLGetConfigOption("MS_NO_VERSION", NULL) != NULL) return ""; // supressing version information
 
   sprintf(version, "MapServer version %s", MS_VERSION);
+
+  static char projVersion[20];
+  sprintf(projVersion, " Proj version %d.%d", PROJ_VERSION_MAJOR, PROJ_VERSION_MINOR);
+  strcat(version, projVersion);
+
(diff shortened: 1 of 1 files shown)


sdlime:
This information does show up in error messages though... Makes the MS_NO_VERSION config file option more important.

geographika:
@sdlime - `mapserv -v` seems fine to me already?
We could simply remove the version from all client responses (and leave it in the command line apps). There are only 3 places - WFS, SOS, and WMS, and then the general `msCGIWriteError` where it would need to be removed. 
It is also available as a template `[version]` option which I think should be left. 

Alternatively maybe setting `MS_NO_VERSION` to 1 by default would work?

sdlime:
@geographika, I was thinking about the web client responses specifically. Suppressing version from the places you reference would be the best fix IMHO. That would also remove the need for MS_NO_VERSION altogether.

Setting MS_NO_VERSION by default would mean pushing that key into the config before loading the config file. There's currently no mechanism to unset it. Currently the code just checks for the presence of the key and doesn't look at the value. That would need to change, so "1" or "TRUE" would turn it on and anything else would turn if off. I'm sort of thinking why bother and go with your first suggestion. 

Maybe a question for the -dev list?

geographika:
I think we could set the default to off using:

```
if(CPLGetConfigOption("MS_NO_VERSION", 1) == 1) return "";
```

and then a user would have to explicitly set  `MS_NO_VERSION 0` in their config file. 

However, I'd be for removing all version details from error messages. Admins can always login to a server and call `mapserv -v` to get these details.

rouault:
The backport of this to 8.0 broke PROJ < 6 support (cf #6905, fix in #6906), but we should be more prudent on what we backport. This is a new functionality, not a bugfix.

PR #6795, opened 2023-01-15
Allow custom projections to be requested via WMS

geographika:
This pull request allows a custom projection codes, in the format AUTHORITY:XXXX e.g. `epsg2:42304` to be advertised and used to reproject data in MapServer using the custom projection. Any non-EPSG projection codes stored in text files in the `PROJ_LIB` directory can be used. 

This resolves the issue where a non-EPSG code cannot be advertised and used by WMS clients, see for example:

+ https://gis.stackexchange.com/questions/301733/mapserver-using-a-crs-for-which-there-is-no-epsg-code
+ https://lists.osgeo.org/pipermail/mapserver-users/2021-December/082473.html

For example in a Mapfile the `wms_srs` parameter can be set to `epsg2:42304`:

```
WEB
  METADATA
    "wms_srs""EPSG:4326 epsg2:42304"
  END
END
```

Clients can then call `&SERVICE=WMS&VERSION=1.1.0&REQUEST=GetMap&SRS=epsg2:42304&..`
For this to work correctly a file named `epsg2` is present in the `PROJ_LIB` directory with the following contents:

```
<42304> +proj=lcc +lat_1=49 +lat_2=77 +lat_0=49 +lon_0=-95 +x_0=0 +y_0=0 +ellps=GRS80 +datum=NAD83 +units=m +no_defs  no_defs <>
```

This update builds on the work in #6484, which allowed a custom projection to be advertised, but unless it was within the `proj.db` database it could not be used (`proj.db` contains not only EPSG, but also ESRI and IGNF). Adding a projection in a text file and placing in the `PROJ_LIB` folder is easier than trying to hack records into `proj.db`, and allows for more future-proof deployments. See initial discussions on the mapserver-dev mailing list: https://lists.osgeo.org/pipermail/mapserver-dev/2023-January/016906.html

Notes:

+ when advertising projection codes EPSG is in uppercase. For custom codes the authority code needs to match the casing of the file in `PROJ_LIB` (on Linux), or it will not be found. For example the file must be named `epsg2` if the advertised projection is  `epsg2:42304`. 
+ there was a bug in MapServer that would allow custom projection codes to be used even if they were not advertised (e.g. any client could request a WMS in `esri:54009`. This is fixed by this pull request, and only advertised projections can be used by a client. This prevents attempts to try and load files into PROJ by a malicious agent. 
+ msautotests cover using a custom projection. It relies on copying a `epsg2` file into the `PROJ_LIB` directory. This copy command has been added to the Travis/GHA build script, and the Vagrant build script. 
+ msautotest also checks non-advertised projections cannot be used. 
+ custom projections in WFS (not OGC API) are not covered by this pull request (several additional functions would need to be modified and a new msautotest added
+ there are discussions on the mailing list about which authority names can be used to remain compliant with the WMS specs. See https://lists.osgeo.org/pipermail/mapserver-users/2018-November/080845.html and https://lists.osgeo.org/pipermail/mapserver-users/2018-November/080848.html. Key quotes from these are:

> However, CRS, EPSG and AUTO namespaces are defined by OGC. So, if you define a new one:
> - 1/ I am not sure it is compatible with OGC standards
> - 2/ I am not sure clients will understand this new namespace

A client will need to be able to request the custom projection and provide suitable extents, but MapServer will be doing the projection server-side. 

From @jratike80: 

> “Every Layer CRS has an identifier that is a character string. Two types of Layer CRS identifiers are permitted:
> “label” and “URL” identifiers:
> ⎯ Label: The identifier includes a namespace prefix, a colon, a numeric or string code, and in some instances a comma followed > by additional parameters. This International Standard defines three namespaces: CRS, EPSG and AUTO2, as discussed below.
> ⎯ URL: The identifier is a fully-qualified URL that references a publicly-accessible file containing a definition of the CRS that is > compliant with ISO 19111.
> "This international standard does not mandate support for any particular Layer CRS(s).


mapfile.c
```diff
@@ -1351,6 +1351,44 @@ int msLoadProjectionStringEPSG(projectionObj *p, const char *value)
...
+
+  if (!strchr(value, ':')) {
+      return -1;
+  }
+
+  char** papszList = NULL;
+  papszList = msStringSplit(value, ':', &(num_params));
+
+  if (num_params != 2) {
+      return -1;
```
rouault, line 1371:
```suggestion
      msFreeCharArray(papszList, num_params);
      return -1;
```

mapfile.c
```diff
@@ -1351,6 +1351,44 @@ int msLoadProjectionStringEPSG(projectionObj *p, const char *value)
...
+  if (strncasecmp(value, "init=", 5) == 0) {
+      return -1;
+  }
+
+  if (!strchr(value, ':')) {
+      return -1;
+  }
+
+  char** papszList = NULL;
+  papszList = msStringSplit(value, ':', &(num_params));
```
rouault, line 1368:
```suggestion
  char** papszList = msStringSplit(value, ':', &(num_params));
```

mapfile.c
```diff
@@ -1351,6 +1351,44 @@ int msLoadProjectionStringEPSG(projectionObj *p, const char *value)
...
+
+  if (num_params != 2) {
+      return -1;
+  }
+
+  size_t buffer_size = 0;
+  char* init_string = NULL;
+  buffer_size = 5 + strlen(value) + 1;
+  init_string = (char*)msSmallMalloc(buffer_size);
```
rouault, line 1377: combined declaration and definitions is both shorter and less error prone:
```suggestion
  const size_t buffer_size = 5 + strlen(value) + 1;
  char* init_string = (char*)msSmallMalloc(buffer_size);
```

mapfile.c
```diff
@@ -1351,6 +1351,44 @@ int msLoadProjectionStringEPSG(projectionObj *p, const char *value)
...
+
+  /* translate into PROJ format. */
+  msStringToLower(papszList[0]);
```
rouault, line 1380: On Linux, filename lookups are case sensitive. I would drop the msStringToLower(). I would say it is the responsibility of the user to provide its input in a consistent way with the case of the file.
Or otherwise it should be documented that the filenames of the init files should be lower case

mapfile.c
```diff
@@ -1351,6 +1351,44 @@ int msLoadProjectionStringEPSG(projectionObj *p, const char *value)
...
+  snprintf(init_string, buffer_size, "init=%s:%d", papszList[0], atoi(papszList[1]));
+  
+  p->args = (char**)msSmallMalloc(sizeof(char*) * 2);
```
rouault, line 1383: I know you took inspiration from code above, but it doesn't look right to me.
It should be either:
- null terminate the array:
```suggestion
  p->args = (char**)msSmallCalloc(sizeof(char*), 2);
```
- or
```suggestion
  p->args = (char**)msSmallMalloc(sizeof(char*));
```
I would say the later is probably the more logical given that (I believe that) args[] is not assumed to be NULL terminated.

mapfile.c
geographika: Only issue with this is that these codes will likely be read directly from the GetCapabilities for the service, and lots of Mapfiles in the wild (and in msautotest) use capital letters in the metadata: `"wms_srs" "EPSG2:42304 EPSG:42101 EPSG:4269 EPSG:4326"`

These codes would then be passed to the GetMap request as-is which would mean they wouldn't be found on Linux. There are lots of instances of hardcoded `init=epsg:...` in the codebase.

rouault: just use capital filenames then :-)

init=epsg:XXXX is treated as a special case inside PROJ given all the backward compatibility issues. If there is an "epsg" filename in PROJ_DATA/PROJ_LIB, it is used. Otherwise PROJ will lookup EPSG:XXXX in proj.db

geographika:
One issue that I've noticed debugging this is that it is possible to request any non-EPSG projection without being restricted by `MAP` or `LAYER` `METADATA` settings (e.g. `wms_srs`).

For example if a WMS 1.3 request includes `&CRS=epsg:54009` and this is not listed as an available projection then the following error is returned:

```
<ServiceExceptionReport xmlns="http://www.opengis.net/ogc" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.3.0" xsi:schemaLocation="http://www.opengis.net/ogc http://schemas.opengis.net/wms/1.3.0/exceptions_1_3_0.xsd">
<ServiceException code="InvalidSRS"> msWMSLoadGetMapParams(): WMS server error. Invalid CRS given : CRS must be valid for all requested layers. msProcessProjection(): Projection library error. proj error "generic error of unknown origin" for "init=epsg:54009" </ServiceException>
</ServiceExceptionReport>
```

However if a request includes `&CRS=esri:54009` then it bypasses any checks, and will return a response.

Relevant code for reference:

https://github.com/MapServer/MapServer/blob/e60fb9a2e85e16104ccea637bd90e5163c66eb58/mapwms.cpp#L1036-L1045

`epsgbuf` is an empty string as it doesn't match `EPSG`, and the following checks in the map and layer metadata are skipped:

https://github.com/MapServer/MapServer/blob/e60fb9a2e85e16104ccea637bd90e5163c66eb58/mapwms.cpp#L1403-L1409

mapfile.c
geographika: True..these are custom authorities so can be named accordingly. Updated.

rouault:
Regarding bypassing things, I'm just thinking this new +init=filename:code based possibility opens the door to accessing a file anywhere in the file system if you pass things like "SRS=/etc/passwd:1010".  We should validate that the authority name doesn't contain any slash or antislash character, so that the lookup is restricted to the PROJ search paths.
That said, even with that restriction, I'm not completely sure what would happen if you pass a filename that is a large file, like a GeoTIFF grid that might be in PROJ_DATA, let's say "us_nga_egm08_25.tif". I wouldn't exclude that PROJ tries to parse the file. OK, I've just tried that and looked at the output of ``PROJ_DATA=../../PROJ-data/us_nga PROJ_USE_PROJ4_INIT_RULES=YES strace bin/projinfo "+init=us_nga_egm08_25.tif:1000"``. It indeeds starts parsing the file, but gives up soon as there's a check that there must be a newline character at least every 1000 bytes.
But all that probably means that there should likely be a allow list of init filenames in the config or mapfile.

geographika:
We do have an msautotest that relies on path manipulation to get a custom projection file already, so I guess there aren't restrictions for projection blocks in the Mapfile:

https://github.com/MapServer/MapServer/blob/e60fb9a2e85e16104ccea637bd90e5163c66eb58/msautotest/wxs/wms_simple.map#L161-L163

(even though `msProcessProjection` sets the PROJ search paths):

https://github.com/MapServer/MapServer/blob/e60fb9a2e85e16104ccea637bd90e5163c66eb58/mapproject.c#L875

For any projections set by a WMS client using `CRS=` or `SRS=` wouldn't restricting any projections to just those advertised by the service be sufficient (noting this is not the case at the moment)? E.g. any values in "ows_srs" metadata values.

rouault:
> (even though msProcessProjection sets the PROJ search paths):

PROJ search paths are only used to resolve filenames without paths. If you specify a relative or absolute filename, PROJ will use it whatever the defined PROJ search paths are.

> so I guess there aren't restrictions for projection blocks in the Mapfile:

That's under the control of a (hopefully) trusted user, so we shouldn't care about that.

> wouldn't restricting any projections to just those advertised by the service be sufficient (noting this is not the case at the moment)

ah, yes you're right. that's the simplest way to proceed. no need for the complication of a blessed list of init files I suggested above.

geographika:
> ah, yes you're right. that's the simplest way to proceed. no need for the complication of a blessed list of init files I suggested above.

The code has been updated to ensure only advertised projections can be requested (with msautotests).

geographika:
Tests added and code suggestions implemented. Any objections to merging into main?

mapfile.c
```diff
@@ -1351,6 +1351,41 @@ int msLoadProjectionStringEPSG(projectionObj *p, const char *value)
...
+  if (num_params != 2) {
+      msFreeCharArray(papszList, num_params);
+      return -1;
+  }
+
+  const size_t buffer_size = 5 + strlen(value) + 1;
+  char* init_string = (char*)msSmallMalloc(buffer_size);
+
+  /* translate into PROJ format. */
+  snprintf(init_string, buffer_size, "init=%s:%d", papszList[0], atoi(papszList[1]));
```
rouault, line 1378: No reason to restrict that to numeric codes. Some authorities have alphanumeric codes, so:
```suggestion
  snprintf(init_string, buffer_size, "init=%s:%s", papszList[0], papszList[1]);
```

rouault (commented):
cf comment

Issue #6801, opened 2023-01-18
Boolean attribute values not of boolean type in WFS GetFeature GeoJSON response


arbakker:
## Expected behavior and actual behavior.

Boolean attribute values in WFS GetFeature GeoJSON response not correct when using `gml_*_type = Boolean` on the boolean attributes. I tested this for both GeoPackage and Postgis datasources, both datasources give incorrect results. 

Example datasource (where `bool_0` has value `0` for al features and `bool_1` value `1` for all features):

```sh
$ ogrinfo ahn3.gpkg ahn3_bladindex -so
...
FID Column = fid
Geometry Column = geom
puuid: String (0.0)
bladnr: String (24.0)
bool_0: Integer(Boolean) (0.0)
bool_1: Integer(Boolean) (0.0)
```

Example mapfile (see attached [zipfile](https://github.com/MapServer/MapServer/files/10443777/ahn3-boolean-issue.zip)) has two layers configured:

- ahn3_bladindex_gpkg (with a GeoPackage datasource)
- ahn3_bladindex_postgis (with a Postgis datasource)

For both layers the `gml_*_type` config option specified to indicate the boolean attribute type for the boolean attributes: 

```
# "gml_types"                             "auto"
"gml_bool_0_type"                     "Boolean" # has value 0 for all features
"gml_bool_1_type"                     "Boolean" # has value 1 for all features
```

Example [GeoJSON GetFeature](http://localhost:8080/mapserver?request=GetFeature&service=WFS&version=2.0.0&srsname=EPSG:28992&typenames=ahn3:ahn3_bladindex_gpkg&bbox=-234141.2978,153578.31569999998,546883.9822,770313.0357000001,EPSG:28992&count=1&outputFormat=application/json) response GPKG:

```json
{
  "type": "FeatureCollection",
  "name": "ahn3_bladindex_gpkg",
  ...
  "features": [
    {
      "type": "Feature",
      "id": "1",
      "properties": { "bladnr": "02ez1", "bool_0": 0, "bool_1": 1 },
      "geometry": {
       ...
      }
    }
  ]
}

```

[...]

PR #6803, opened 2023-01-18
Set Boolean subtype for OGR Output

geographika:
Currently OGR Output from MapServer ignores field subtypes. See https://gdal.org/development/rfc/rfc50_ogr_field_subtype.html for details on subtypes (from @rouault) - there is no Boolean field type in OGR, but there is an Integer field with a subtype of Boolean. 

This fixes the issue in #6801 (at least for GeoJSON) and the output changes from `"properties": { "id": 1, "mybool": 0 },` to 
`"properties": { "id": 1, "mybool": false },`

However as noted in the RFC not all OUTPUTFORMATS may support this subtype so I'm not sure if this should be merged. 

@arbakker - if you are able to test this change with your data, could you see if it resolves the issue.

Will add to msautotests if this is a change we want to merge.


rouault (approved):

arbakker:
@geographika Will take a look tomorrow, thanks for cranking out a fix that quick and the quick replies.

arbakker:
I just checked, and found that the JSON output for PostGIS datasource is still incorrect (`bool_1` value is outputted as `false`, while should be `true`):

```json
...
"properties": { "bladnr": "02ez1", "bool_0": false, "bool_1": false },
...
```

JSON output for the GeoPackage datasource is correct:
```json
...
"properties": { "bladnr": "02ez1", "bool_0": false, "bool_1": true },
...
```

geographika:
I get the same results. I can see why this is happening, but not sure of the fix..

The Postgres MapServer driver requests values as text e.g. `select id::text, bool_1::text from ahn3_bladindex`
The value associated with the record is then the string `"true"`. 

It appears that when the following call is made: https://github.com/MapServer/MapServer/blob/4de5c84f2236c87e5b81d2c26d857bc7649df4b9/mapogroutput.cpp#L497

`shape->values[i] // "true"`

And this produces the `false` output in the GeoJSON - any non-0 value is considered false:

https://github.com/OSGeo/gdal/blob/a632e3093d6dd34cdcf979a5b2a20529fe2df943/ogr/ogrfeature.cpp#L3339-L3345

```
    if (poFDefn->GetSubType() == OFSTBoolean && nValue != 0 && nValue != 1)
    {
        CPLError(CE_Warning, CPLE_AppDefined,
                 "Only 0 or 1 should be passed for a OFSTBoolean subtype. "
                 "Considering this non-zero value as 1.");
        nValue = 1;
    }
```

If the value passed here is "1" then it correctly outputs `true`. However the only way to do this is to explicitally convert the Boolean to an int in the `DATA` clause e.g. `select id::text, bool_1::int from ahn3_bladindex`, and then also overriding `"gml_types" "auto"` by adding `"gml_bool_1_type" "Boolean"` to the `METADATA`. 

It would be possible to add in a hack to check for `true` or `false` values and then check the field subtype, and switch the value to "1" or "0", I can't see any other way apart from modifying `OGR_F_SetFieldString` in GDAL, or modifiying the Postgres MapServer driver.

rouault:
I've done a GDAL upstream fix per https://github.com/OSGeo/gdal/pull/7097, but it could make sense for MapServer before calling OGR_F_SetFieldString( hFeat, out_field++, shape->values[i] ) on a boolean field to test shape->values[i] against false or true to replace with 0 or 1.

mapogroutput.cpp
```diff
@@ -494,6 +494,28 @@ static int msOGRWriteShape( OGRLayerH hOGRLayer,
       }
     }
 
+#if GDAL_VERSION_MAJOR <= 2 || (GDAL_VERSION_MAJOR == 3 && GDAL_VERSION_MINOR < 7)
```
rouault, line 497:
```suggestion
    const char* pszValue = shape->values[i];
#if GDAL_VERSION_MAJOR <= 2 || (GDAL_VERSION_MAJOR == 3 && GDAL_VERSION_MINOR < 7)
```

mapogroutput.cpp
```diff
@@ -494,6 +494,28 @@ static int msOGRWriteShape( OGRLayerH hOGRLayer,
...
+    // Boolean true and false values are handled in GDAL 3.7+
+    // otherwise any string values revert to false
+    if (strcmp(shape->values[i], "true") == 0) {
+        OGRFieldDefnH hFieldDefn = OGR_FD_GetFieldDefn(hLayerDefn, out_field);
+        OGRFieldSubType eFieldSubType = OGR_Fld_GetSubType(hFieldDefn);
+        if (eFieldSubType == OFSTBoolean)
+        {
+            shape->values[i] = msStrdup("1");
```
rouault, line 505:
```suggestion
            pszValue = "1";
```

mapogroutput.cpp
```diff
@@ -494,6 +494,28 @@ static int msOGRWriteShape( OGRLayerH hOGRLayer,
...
+        }
+    }
+
+    if (strcmp(shape->values[i], "false") == 0) {
+        OGRFieldDefnH hFieldDefn = OGR_FD_GetFieldDefn(hLayerDefn, out_field);
+        OGRFieldSubType eFieldSubType = OGR_Fld_GetSubType(hFieldDefn);
+        if (eFieldSubType == OFSTBoolean)
+        {
+            shape->values[i] = msStrdup("0");
```
rouault, line 514:
```suggestion
            pszValue = "0";
```

mapogroutput.cpp
```diff
@@ -494,6 +494,28 @@ static int msOGRWriteShape( OGRLayerH hOGRLayer,
...
+        }
+    }
+#endif
+
     OGR_F_SetFieldString( hFeat, out_field++, shape->values[i] );
```
rouault, line 519:
```suggestion

    OGR_F_SetFieldString( hFeat, out_field++, pszValue );
```

rouault (commented):
suggested changes to avoid memory leaks

geographika:
Thanks @rouault for the code review and the `OGR_F_SetFieldString` update!

@arbakker - if you could test again the GeoJSON output that would be great. It should work with this pull request, and/or with GDAL 3.7 when it is released.

arbakker:
@geographika  I just checked and found the following:

|                            | gml-output postgis ds | gml-output geopackage ds | geojson-output postgis ds | geojson-output-geopackage ds |
| -------------------------- | --------------------- | ---------------------- | ---------------------- | ------------------------- |
| `gml_types auto` |                   YES |                     NO |                     YES |                        NO |
| `gml_*_type`    |                   YES |                    YES[^1] |                     YES |                        YES |

[^1]: Boolean values represented as `0` or `1` in GML, as opposed to PostGIS datastore where boolean values are represented as `true` or `false` in GML. 

So the fix works, although the `gml_type auto` config option is not working for boolean values in a GeoPackage datasource, but I suppose that was to be expected considering your earlier comment:

> If the value passed here is "1" then it correctly outputs true. However the only way to do this is to explicitally convert the Boolean to an int in the DATA clause e.g. select id::text, bool_1::int from ahn3_bladindex, and then also overriding "gml_types" "auto" by adding "gml_bool_1_type" "Boolean" to the METADATA.

In any case it is not an issue for me that the  `gml_type auto` config option is not working for boolean values in a GeoPackage datasource, but pointed it out to make clear what the fix fixes and what it doesnt fix :).

geographika:
The latest commit hopefully sets the field to Booleans when reading from any OGR datasource, including GeoPackage. I'll look at adding another test case for this. 

I'm not sure what the "gml_width" should be set to in these cases though - if they are read as a "1", and the size of the field is 1, but then the output is switched to "true", could they get truncated?

geographika:
I've added tests for GeoPackage / Booleans / JSON as well. 
Some of the types for Dates/Times seem strange, but these can be addressed in a separate pull request. @arbakker the Booleans now seem to be correct in the GeoJSON output. 

```
{
"type": "FeatureCollection",
"numberMatched": 1,
"name": "autotypes",
"features": [
{ "type": "Feature", "properties": { "mychar": "abc  ", "myvarchar": "def", "mytext": "ghi", "mybool": true,
"myint2": 10, "myint4": 100, "myint8": 1000, "myfloat4": 1.5, "myfloat8": 2.5, "mynumeric": 3.33,
"mydate": "2023-01-01", "mytime": "00:00:00", "mytimez": "00:00:00", "mytimestamp": "2023-01-01T00:00:00",
"mytimestampz": "2023-01-01T00:00:00Z" },
"geometry": { "type": "Point", "coordinates": [ 1.0, 2.0 ] } }
]
}
```

from `ogrinfo`:

```
mychar (String) = abc
myvarchar (String) = def
mytext (String) = ghi
mybool (Integer(Boolean)) = 1
myint2 (Integer(Int16)) = 10
myint4 (Integer) = 100
myint8 (Integer64) = 1000
myfloat4 (Real(Float32)) = 1.5
myfloat8 (Real) = 2.5
mynumeric (Real) = 3.33
mydate (Date) = 2023/01/01
mytime (String) = 00:00:00
mytimez (String) = 00:00:00
mytimestamp (DateTime) = 2023/01/01 00:00:00
mytimestampz (DateTime) = 2023/01/01 00:00:00+00
```

Corresponding GML types - `TimeInstantType` seems to be used for Dates, and time types are strings:

```
<element name="mychar" minOccurs="0" type="string"/>
<element name="myvarchar" minOccurs="0" type="string"/>
<element name="mytext" minOccurs="0" type="string"/>
<element name="mybool" minOccurs="0" type="boolean"/>
<element name="myint2" minOccurs="0" type="integer"/>
<element name="myint4" minOccurs="0" type="integer"/>
<element name="myint8" minOccurs="0" type="long"/>
<element name="myfloat4" minOccurs="0" type="double"/>
<element name="myfloat8" minOccurs="0" type="double"/>
<element name="mynumeric" minOccurs="0" type="double"/>
<element name="mydate" minOccurs="0" type="gml:TimeInstantType"/>
<element name="mytime" minOccurs="0" type="string"/>
<element name="mytimez" minOccurs="0" type="string"/>
<element name="mytimestamp" minOccurs="0" type="gml:TimeInstantType"/>
<element name="mytimestampz" minOccurs="0" type="gml:TimeInstantType"/>
```

PR #6804, opened 2023-01-23
Small changes missed in the initial 8.0 release.

sdlime:
This does a couple of things discovered through various upgrade work (by myself and others):

1. Allows map.size and map_size CGI parameters as aliases for mapsize. Older versions of OpenLayers (OL) supported a OpenLayers.Layer.MapServer layer type that used an alias instead of just mapsize. This should allow using MapServer 8 without having to re-work an old OL interface.
2. Allows runtime substitution on the web TEMPLATE property.

--Steve

### DIFF
--- a/mapfile.c
+++ b/mapfile.c
@@ -6470,7 +6470,9 @@ static void mapSubstituteString(mapObj *map, const char *from, const char *to) {
       map->outputformatlist[l]->formatoptions[o] = msCaseReplaceSubstring(map->outputformatlist[l]->formatoptions[o], from, to);
     }
   }
+
   hashTableSubstituteString(&map->web.metadata, from, to);
+  if(map->web.template) map->web.template = msCaseReplaceSubstring(map->web.template, from, to);
 }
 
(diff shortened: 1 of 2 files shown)


geographika:
@sdlime - will be great to get this fix in and get the Itasca demo working again. 
There is some ambiguity in the docs about which `TEMPLATE` values can be updated using runtime-substitution:

+ https://mapserver.org/mapfile/layer.html#mapfile-layer-template - is run-time Substitution allowed? Not mentioned here.
+ https://mapserver.org/mapfile/class.html#mapfile-class-template - has a note "This attribute can be replaced using runtime substitution. See Run-time Substitution. "

https://mapserver.org/cgi/runsub.html#runsub lists the following:

```
CLASS: EXPRESSION
CLASS: TEXT
CLASS: TITLE
LAYER: CONNECTION
LAYER: DATA
LAYER: FILTER
LAYER: TILEINDEX
LAYER: METADATA
LAYER: PROCESSING
OUTPUTFORMAT: all FORMATOPTIONs
MAP->WEB: METADATA (values only, not keys)
```

geographika (approved):

PR #6806, opened 2023-01-25
Check for DEBUG level before writing cache warnings to log

geographika:
On a production server with feature caching the logs are currently filling up with lines similar to below:

```
[Mon Jan 23 09:15:24 2023].335000 map->query.max_cached_shape_count = 1000 reached. Next features will not be cached.
[Mon Jan 23 09:15:25 2023].190000 map->query.max_cached_shape_count = 1000 reached. Next features will not be cached.
[Mon Jan 23 09:15:26 2023].547000 map->query.max_cached_shape_count = 1000 reached. Next features will not be cached.
[Mon Jan 23 09:15:28 2023].588000 map->query.max_cached_shape_count = 1000 reached. Next features will not be cached.
[Mon Jan 23 09:15:29 2023].557000 map->query.max_cached_shape_count = 1000 reached. Next features will not be cached.
[Mon Jan 23 09:15:30 2023].604000 map->query.max_cached_shape_count = 1000 reached. Next features will not be cached.
[Mon Jan 23 09:15:31 2023].572000 map->query.max_cached_shape_count = 1000 reached. Next features will not be cached.
[Mon Jan 23 09:15:32 2023].567000 map->query.max_cached_shape_count = 1000 reached. Next features will not be cached.
[Mon Jan 23 09:15:33 2023].587000 map->query.max_cached_shape_count = 1000 reached. Next features will not be cached.
[Mon Jan 23 09:15:34 2023].673000 map->query.max_cached_shape_count = 1000 reached. Next features will not be cached.
```

This update checks for the `DEBUG` level of the map prior to writing these. It will need to be `MS_DEBUGLEVEL_V = 3` or above to include these.

### DIFF
--- a/mapquery.c
+++ b/mapquery.c
@@ -177,9 +177,11 @@ static int canCacheShape(mapObj* map, queryCacheObj *queryCache, int shape_ram_s
       if( !queryCache->cachedShapeCountWarningEmitted )
       {
           queryCache->cachedShapeCountWarningEmitted = MS_TRUE;
-          msDebug("map->query.max_cached_shape_count = %d reached. "
+          if (map->debug >= MS_DEBUGLEVEL_V) {
+              msDebug("map->query.max_cached_shape_count = %d reached. "
                   "Next features will not be cached.\n",
                   map->query.max_cached_shape_count);
+          }
       }
       return MS_FALSE;
   }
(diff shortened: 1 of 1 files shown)


sdlime:
Seems like a reasonable change to me, so +1.

geographika:
@sdlime - as a general rule should all calls to `msDebug` be wrapped in a check against debug level?

sdlime:
> @sdlime - as a general rule should all calls to `msDebug` be wrapped in a check against debug level?

I would think so - with few exceptions.

Issue #6631, opened 2022-09-25
Encoding of shapefile breaks OGCAPI: Features (both HTML and JSON response)


jmckenna:
(this was very difficult to isolate, but I did, took me over a week ha, to isolate this to one single record, with 3 attributes only: `name`, `id`, and `name_hu` where name_hu is `Tanzánia`)

_Declaration: this could be just for my documentation job for mapserver.org, ha, instead of an "issue" here, but, at least by reporting this here other readers will be warned before they hit this wall, which will likely save them a week of effort._

### Description
- testing with `main` branch, on Windows
- this same record works great inside a database (PostgreSQL, SpatiaLite) and served through OGCAPI: Features
- this same record works great served through OGC: WFS (as a Shapefile, PostgreSQL, SpatiaLite)
- however in testing these past few weeks, using the same dataset but in ESRI Shapefile format, breaks the OGCAPI: Features response

### Symptom 1
- request: `ogcapi/collections/countries/items?f=html`
- response: `{"code":"ServerError","description":"General template handling error."}`
  - yes the response is in JSON even though the request is for HTML

### Symptom 2
- request: `/ogcapi/collections/countries/items?f=json`
- response: blank (literally no response, as shown in browser Developer Tools)

### Solution attempt # 100
- change encoding in the templates (header.html, openapi.html) from `<meta charset="utf-8" />` to `<meta charset="ISO-8859-1" />`
- result: no change, same errors

### Solution attempt # 101
- in the mapfile LAYER section, add `ENCODING "ISO-8859-1"`
- result: fixes both the JSON and HTML item response

Are you interested in trying this locally? Here is that single record shapefile (it is only 2KB in size, as it is one polygon feature with 1 record) : https://gatewaygeomatics.com/dl/ogcapi-features-shp.zip

(I also realize that I just spent a week on creating what could be the perfect new shapefile for a new msautotest, ha, yikes)

sdlime:
The first symptom is expected since all fundamental errors are returned as JSON. The second symptom should be trapped somehow.

sdlime:
@jmckenna, by chance is there anything in the Apache or MapServer logs with call 2?

geographika:
@sdlime - related to the OGCAPI and debugging - would it make sense to add in "standard" MS error logging to the `outputError` function?

https://github.com/MapServer/MapServer/blob/2fcc5b6c690b532be512738690625f68bdb9e7cf/mapogcapi.cpp#L87

[...]

PR #6816, opened 2023-02-02
Catch exception when JSON contains invalid UTF-8 data.

sdlime:
Addresses issue #6631. Also updated Inja from version 3.3 to 3.4...


sdlime:
Still need to add msDebug() call to outputError(). Presently a mapObj doesn't make it to outputError() so there's no way to check a debug level. Or should it just write regardless?

sdlime:
Ugh... I'm gonna redo this without Inja update.

PR #6817, opened 2023-02-03
Issue 6631

sdlime:
Fix for issue #6631 and corresponding tests.


sdlime:
@geographika, this doesn't include logging in outputError(). As written, the mapObj isn't available so we can't check map->debug. We could refactor things - could also just call msDebug() whenever an error is encountered. Might be a little chatty though.

sdlime:
Failing test is related to SQL Server and not this change.

geographika:
> Failing test is related to SQL Server and not this change.

Looking into it. A few things broken with the GDAL dependency update. 

Noted re error message. I think the map debug level could be passed into the parent functions in a future pull request.

geographika (approved):

jmckenna:
@sdlime thanks for your effort on this.  I have built your branch here on Windows, with MS4W, and here are the new results:

- items?f=html
  - `{"code":"ServerError","description":"General template handling error."}`
- items?f=json
  - `{"code":"ConfigError","description":"Invalid UTF-8 data, check encoding."}`
  
That's a great improvement, thanks!  (so many MS4W users would have been contacting me with this problem blank page, at least now a meaningful error is displayed)

(as I had mentioned in my previous research in September, setting `ENCODING "ISO-8859-1` at the `LAYER` level does solve it)

Thanks again!

/me approves pull request 🚀

jmckenna:
@sdlime could that error message be enhanced such as "Invalid UTF-8 data, check encoding, which is commonly set through the ENCODING parameter at the LAYER level of mapfile"     ?

sdlime:
> @sdlime could that error message be enhanced such as "Invalid UTF-8 data, check encoding, which is commonly set through the ENCODING parameter at the LAYER level of mapfile" ?

I was trying to keep the error messages as generic as possible, easier to obfuscate. I could see logging a that level though.

sdlime:
Note that the HTML response may vary by the set of templates. The plain html actually does render albeit with the bad UTF-8 characters. There bootstrap templates generate the server error. The easiest way I can think of to always generate the same response for both HTML and JSON would be to do a meaningless j.dump() at the top of the outputTemplate() function and bail if an exception is generated. It would be a quick add if that consistency is important.

jmckenna:
> Note that the HTML response may vary by the set of templates. The plain html actually does render albeit with the bad UTF-8 characters. There bootstrap templates generate the server error. The easiest way I can think of to always generate the same response for both HTML and JSON would be to do a meaningless j.dump() at the top of the outputTemplate() function and bail if an exception is generated. It would be a quick add if that consistency is important.

I vote for consistency.

sdlime:
> > Note that the HTML response may vary by the set of templates. The plain html actually does render albeit with the bad UTF-8 characters. There bootstrap templates generate the server error. The easiest way I can think of to always generate the same response for both HTML and JSON would be to do a meaningless j.dump() at the top of the outputTemplate() function and bail if an exception is generated. It would be a quick add if that consistency is important.
> 
> I vote for consistency.

Added. Would be good to have @rouault confirm this approach seems reasonable code-wise...

rouault:
> Would be good to have @rouault confirm this approach seems reasonable code-wise...

no strong opinion, but sounds reasonable.

jmckenna:
(build issue is related to GDAL issue: https://github.com/OSGeo/gdal/issues/7179 )

Thanks, Steve and all, merging!

rouault:
> (build issue is related to GDAL issue: [OSGeo/gdal#7179](https://github.com/OSGeo/gdal/issues/7179) )

Although they are MSSQL related, the connections between this issue and the GDAL issue you point to is far from obvious to me

jmckenna:
@rouault sorry, this issue was stalled by a CI build error, which is related to https://github.com/OSGeo/gdal/issues/7179 )

jmckenna:
Anyway, merging, sorry for the noise here

rouault:
The "ERROR 1: Unable to find driver `MSSQLSpatial'." error message indicates more that either the gisinternals GDAL build lacks MSSQLSpatial support or that GDAL_DRIVER_PATH isn't properly set

jmckenna:
@rouault true.  I think the GDAL link happens with Tomas' recent buildkit, which was examined to handle the error that you mentioned.   It is sure out of my paygrade, ha.   Sorry for the link here.

PR #6823, opened 2023-02-07
Fix GetMetadata possible buffer overrun

geographika:
As part of updating Appveyor to the new GDAL/PROJ SDK one of the MS SQL Server tests was failing relating to a `GetMetadata` request:

> mapserv -nh "QUERY_STRING=map=C:/projects/mapserver/msautotest/mssql/cluster_mssql.map&REQUEST=GetMetadata&LAYER=cities"

Caused the following: `msIO_vfprintf(): General error message. Possible buffer overrun.`

Note on Windows (or at least Appveyor and a local machine) `HAVE_VSNPRINTF` is false. 

Only similar issue is #2599 - the string to write is larger than the buffer. Looking at other functions that write XML I tried the approach used for WFS 2.0 which fixed the issue. 

https://github.com/MapServer/MapServer/blob/68fd623eddc772137952509a8008209a02cd1e85/mapwfs20.c#L729-L736

Note this also updates the output to use UTF-8 (which is probably what we want for consistency?). 

I've no idea why updating GDAL or PROJ would cause this test to fail though (when fixed the output XML was identical apart from the UTF-8 tag added by the update) - if anyone can provide any insights I'd appreciate it.

Also there are several instances where `xmlDocDumpFormatMemoryEnc` is followed by `msIO_printf`. Are these also at reisk of buffer overruns and should they be changed to `msIO_contextWrite`?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,7 +122,7 @@ check_function_exists("strlen"  HAVE_STRLEN)
 check_function_exists("strncasecmp"  HAVE_STRNCASECMP)
 
 check_symbol_exists(vsnprintf stdio.h HAVE_VSNPRINTF)
-IF(NOT HAVE_SNPRINTF)
+IF(NOT HAVE_VSNPRINTF)
     check_function_exists("vsnprintf"  HAVE_VSNPRINTF)
 ENDIF()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -121,7 +121,7 @@ check_function_exists("strlcpy"  HAVE_STRLCPY)
 check_function_exists("strlen"  HAVE_STRLEN)
 check_function_exists("strncasecmp"  HAVE_STRNCASECMP)
 
-check_symbol_exists(vsnprintf stdio.h HAVE_SNPRINTF)
+check_symbol_exists(vsnprintf stdio.h HAVE_VSNPRINTF)
 IF(NOT HAVE_SNPRINTF)
     check_function_exists("vsnprintf"  HAVE_VSNPRINTF)
 ENDIF()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,7 @@ endif("${PROJECT_SOURCE_DIR}" STREQUAL "${PROJECT_BINARY_DIR}")
 
 include(CheckLibraryExists)
 include(CheckFunctionExists)
+include(CheckSymbolExists)
 include(CheckIncludeFile)
 include(CheckCSourceCompiles)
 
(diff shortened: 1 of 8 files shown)


geographika:
Note the failing Appveyor is unrelated - this fix will help get it moved to the latest SDK and working again.

rouault:
> Note on Windows (or at least Appveyor and a local machine) `HAVE_VSNPRINTF` is false.

That's were things are not OK, because vsnprintf() is available in MSVC 2015+ : https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/vsnprintf-vsnprintf-vsnprintf-l-vsnwprintf-vsnwprintf-l?view=msvc-140

https://github.com/kcat/openal-soft/issues/12 would suggest to perhaps changing in top CMakeLists.txt
```check_function_exists("vsnprintf"  HAVE_VSNPRINTF)```
to
```check_symbol_exists(vsnprintf stdio.h HAVE_VSNPRINTF)```

File: CMakeLists.txt [hunk 1]
```diff
@@ -119,7 +120,12 @@ check_function_exists("strlcat"  HAVE_STRLCAT)
 check_function_exists("strlcpy"  HAVE_STRLCPY)
 check_function_exists("strlen"  HAVE_STRLEN)
 check_function_exists("strncasecmp"  HAVE_STRNCASECMP)
-check_function_exists("vsnprintf"  HAVE_VSNPRINTF)
+
+check_symbol_exists(vsnprintf stdio.h HAVE_SNPRINTF)
```
rouault, line 124:
```suggestion
check_symbol_exists(vsnprintf stdio.h HAVE_VSNPRINTF)
```

geographika:
Thanks @rouault for the CMake link. 
Should I roll back the other changes e.g. use of `context = msIO_getHandler(stdout); `?
Is there any consensus on use of `UTF-8` output for services that return XML?

I'm still uncertain why the update caused this test to fail though. In the last working Appveyor build the function also wasn't found: `looking for vsnprintf - not found`

https://ci.appveyor.com/project/MapServer/mapserver/builds/46069580/job/12wb5g2mfwlwq5u2#L92

rouault:
> Should I roll back the other changes e.g. use of `context = msIO_getHandler(stdout); `?
> Is there any consensus on use of `UTF-8` output for services that return XML?

I guess you can keep the change

geographika:
Unless there are any concerns I'll merge this in the next couple of hours, and can then submit the other fixes for the Appveyor builds.

sdlime:
Is this a candidate for the 7.6 branch as well?

geographika:
@sdlime - I guess the buffer overrun could be a security issue. The update does modify the build process though, so not sure how/if that is tested with the 7.6 branch.

sdlime:
Isn't the error message coming from within mapserver - so not a crash? The cmake changes seem like they would be portable - not an area of strength for me though.

geographika:
@sdlime - sorry yes you're right - the error is caught and `Possible buffer overrun` returned as an error, so not sure it is worth a backport. Not sure CMake is an area of strength for anyone..!
Do you know when/why to use `msIO_contextWrite` as opposed to `msIO_printf`?

sdlime:
Agreed on the backport - that error message gets your attention. I'm not sure on your other question, perhaps @dmorissette recalls.

Issue #6824, opened 2023-02-07
rectObj wrong projection with PROJ9


geographika:
Issue #6478 seems to reappear when using MapServer main branch and PROJ9.1. 
Using the test case attached to #6478 I get different number of features returned based on if the fix below is applied or not. 

When the fix is skipped (which it is when compiling with PROJ9 due to `PROJ_VERSION_MAJOR < 9`) I get less features.
If I change `PROJ_VERSION_MAJOR < 10` then I get 2 additional features. 

https://github.com/MapServer/MapServer/blob/68fd623eddc772137952509a8008209a02cd1e85/mapproject.c#L2187-L2204

Also simply changing `int apply_over = MS_TRUE;` to `int apply_over = MS_FALSE` also appears to fix the issue. 
Proposed fix at https://github.com/geographika/mapserver/commit/d00f46e0d5c72d7b7dda31fbd590f527f0366870

Associated discussion at https://github.com/MapServer/MapServer/issues/6478#issuecomment-1416726125
PROJ fix (added in 9.0) at https://github.com/OSGeo/PROJ/pull/3055
Original patch commit: https://github.com/rouault/mapserver/commit/25171a48476558337dedfe449eb1e8daaa6a5b9a

rouault:
I've built PROJ 9.1 and rebuilt GDAL and MapServer against it and test_reprojection_rect_and_datum_shift() passes fine. I'm wondering if there isn't some PROJ mismatch in your build configuration.

geographika:
> I've built PROJ 9.1 and rebuilt GDAL and MapServer against it and test_reprojection_rect_and_datum_shift() passes fine. I'm wondering if there isn't some PROJ mismatch in your build configuration.

This does seem to be working again. It was failing when I tried to update in #6827. I have re-enabled the test in #6888.

PR #6825, opened 2023-02-07
Fix for #6824 and PROJ9

geographika:
Possible fix to keep tests passing when compiling with PROJ9. 
See details at #6824.

### DIFF
--- a/mapproject.c
+++ b/mapproject.c
@@ -2175,11 +2175,14 @@ int msProjectRect(projectionObj *in, projectionObj *out, rectObj *rect)
    *  parameter in order to disable dateline wrapping.
    */ 
   else {
-    int apply_over = MS_TRUE;
+    int apply_over = MS_FALSE;
 #if PROJ_VERSION_MAJOR >= 6 && PROJ_VERSION_MAJOR < 9
     // Workaround PROJ [6,9[ bug (fixed per https://github.com/OSGeo/PROJ/pull/3055)
     // that prevents datum shifts from being applied when +over is added to +init=epsg:XXXX
     // This is far from being bullet proof but it should work for most common use cases
+
+    apply_over = MS_TRUE;
+
     if(in && in->proj)
     {
         if( in->numargs == 1 && EQUAL(in->args[0], "init=epsg:4326") &&


rouault:
I don't believe this fix is correct. Its effect is to disable adding +over for PROJ >=9

geographika:
Removing `PROJ_VERSION_MAJOR < 9` also allows the features in the test cast to be picked up, and gets the `test_bug_check.py` to pass. From the code comment though https://github.com/OSGeo/PROJ/pull/3055 should address this (running with PROJ 9.1). 

In PROJ9 there should be no issues adding `+over` to all projections no matter what the rect extents are? 

The `<wfs:boundedBy>` differ depending on if `+over` is called or not:

```
        <gml:Envelope srsName="urn:ogc:def:crs:EPSG::3857">
                <gml:lowerCorner>545327.950179 6867645.996197</gml:lowerCorner>
                <gml:upperCorner>545672.823313 6867780.931668</gml:upperCorner>
        </gml:Envelope>
```

```	  
       <gml:Envelope srsName="urn:ogc:def:crs:EPSG::3857">
                <gml:lowerCorner>545327.950179 6867645.996197</gml:lowerCorner>
                <gml:upperCorner>545741.620449 6867933.162223</gml:upperCorner>
        </gml:Envelope>
```

rouault:
> In PROJ9 there should be no issues adding `+over` to all projections no matter what the rect extents are?

Absolutely no. +over in all PROJ versions only works for some projections.

> The <wfs:boundedBy> differ depending on if +over is called or not:

This might be the effect that datum shifts are disable when +over is set.

+over is a kind of nightmare to deal with in modern PROJ. I've done recent fixed in PROJ 9.2dev because there were some issues in PROJ 9.1.0 or 9.1.1

geographika:
It looks extremely complicated indeed..I propose to get the Appveyor Windows builds working again to either:

1. Leave the code as-is and simply disable the failing Python test that checks the #6824 bug.

or

2. Change `PROJ_VERSION_MAJOR >= 6 && PROJ_VERSION_MAJOR < 9` to include a PROJ version minor check as well, and test again with a new PROJ release.

rouault:
> 1. Leave the code as-is and simply disable the failing Python test that checks the

I would do that. Perhaps the test could be just skipped testing an environment variable that would be set on configurations where it fails ?

szekerest:
@geographika Within the scope of this issue, would you fix the appveyor.yml  by replacing the references to the proj7 directory with proj9? I hope that would fix the failing builds.

geographika:
> @geographika Within the scope of this issue, would you fix the appveyor.yml by replacing the references to the proj7 directory with proj9? I hope that would fix the failing builds.

@szekerest pull request on the way! I'll be closing this and making a new pull request with the SDK / PROJ9 updates.

szekerest:
@geographika I'm not sure if that was considered, but now you should also specify the proj9.lib and proj9 include dirs in the cmake configuration, something like:
https://github.com/gisinternals/buildsystem/blob/a80fe7e8daae0f77b4b63af66d28053bf11327cd/Makefile#L902
This is because the SDK still keeps multiple proj versions.

geographika:
Disabling Python test in Appveyor for now.

PR #6827, opened 2023-02-07
Appveyor SDK Update and Fixes

geographika:
This pull request gets the Appveyor builds working again with the new MapServer/GDAL SDKs from https://www.gisinternals.com/ (thanks to @szekerest). 

The new SDKs use PROJ 9.1 and GDAL 3.7. Updating to these new libraries required a few fixes. Related links and discussions below:

- PROJ9.1 issue with `+over` - see #6825 - test disabled for now
- Failing `GetMetadata` test fixed with #6823
- MSSQL GDAL driver (used to load data for test cases) - see discussions at https://lists.osgeo.org/pipermail/gdal-dev/2023-February/056827.html and https://github.com/OSGeo/gdal/issues/7179
- Update the linefeed GHA to use latest Ubtuntu as getting `This is a scheduled Ubuntu-18.04 brownout` failures - https://github.com/MapServer/MapServer/actions/runs/4118462904


szekerest:
@geographika I've modified the SDK yesterday, so that a default MSSQLSpatial driver is built into GDAL by default, against the  legacy ODBC driver. So adding GDAL_DRIVER_PATH is not a strict requirement to get the build working.
However you should still add -DPROJ_LIBRARY to the cmake command line because it should refer to proj9.lib which is not detected by cmake automatically. See and example below for reference:
https://github.com/gisinternals/buildsystem/blob/a80fe7e8daae0f77b4b63af66d28053bf11327cd/Makefile#L902
This is because the SDK maintains multiple proj versions at the moment.

geographika:
> @geographika I've modified the SDK yesterday, so that a default MSSQLSpatial driver is built into GDAL by default, against the legacy ODBC driver. So adding GDAL_DRIVER_PATH is not a strict requirement to get the build working. However you should still add -DPROJ_LIBRARY to the cmake command line because it should refer to proj9.lib which is not detected by cmake automatically. See and example below for reference: https://github.com/gisinternals/buildsystem/blob/a80fe7e8daae0f77b4b63af66d28053bf11327cd/Makefile#L902 This is because the SDK maintains multiple proj versions at the moment.

Thanks for the feedback @szekerest - changes integrated. Will see if the latest SDK is picked up with the default MSSQLSpatial driver.

szekerest:
This needs to be back ported

PR #6833, opened 2023-02-15
Fix build issue when abseil-cpp is present (fixes #6822)

### LINKED ISSUE #6822: hidden random build dep on abseil-cpp ?
support for flatgeobuf added in #6520 adds an undefined build behaviour depending wether abseil-cpp is installed here: https://github.com/MapServer/MapServer/blob/6da3b29110b6a371c7f141bd2f060d00d860b7ad/flatgeobuf/include/flatbuffers/base.h#L218

if built with C++11 (the default ?) and clang 13 (default compiler on OpenBSD) and a recent version of abseil-cpp happens to be installed, the build will fail because abseil-cpp insists on c++14. 
```
absl/policy_checks.h:79:2: error: "C++ versions less than C++14 are not supported."
```
Cf https://marc.info/?t=167567663600006&r=1&w=2 for the reasoning..

rouault:
Port of upstream fix in https://github.com/google/flatbuffers/pull/7824

### DIFF
--- a/flatgeobuf/include/flatbuffers/base.h
+++ b/flatgeobuf/include/flatbuffers/base.h
@@ -236,7 +236,7 @@ namespace flatbuffers {
       } }
       #define FLATBUFFERS_HAS_STRING_VIEW 1
     // Check for absl::string_view
-    #elif __has_include("absl/strings/string_view.h")
+    #elif __has_include("absl/strings/string_view.h") && (__cplusplus >= 201411)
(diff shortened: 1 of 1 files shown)


rouault:
@landryb does that fix your build issue?

landryb:
> @landryb does that fix your build issue?

yes, and that's definitely better. Will also apply it to postgis.. thanks !

PR #6838, opened 2023-02-17
Improve round-trip of mapfiles through mapscript

alexbrault:
The BINDVALS, UTFDATA and UTFITEM attributes in layers were missing from writeLayer()

COLORRANGE and DATARANGE were only written in STYLEs if RANGEITEM was also defined, but kernel density layers also make use of those attributes without using RANGEITEM


jmckenna:
thanks @alexbrault !

PR #6854, opened 2023-04-08
mapfile parser: fix double-free when included file doesn't exist

rouault:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=57788

### DIFF
--- a/maplexer.c
+++ b/maplexer.c
@@ -4418,16 +4418,16 @@ YY_RULE_SETUP
                                                    return(-1);
                                                  }
 
-                                                 include_stack[include_stack_ptr] = YY_CURRENT_BUFFER; /* save state */
-                                                 include_lineno[include_stack_ptr] = msyylineno;
-                                                 include_stack_ptr++;
-
                                                  msyyin = fopen(msBuildPath(path, msyybasepath, msyytext), "r");
                                                  if(!msyyin) {
                                                    msSetError(MS_IOERR, "Error opening included file \"%s\".", "msyylex()", msyytext);
                                                    msyyin = YY_CURRENT_BUFFER->yy_input_file;
                                                    return(-1);
                                                  }
+
+                                                 include_stack[include_stack_ptr] = YY_CURRENT_BUFFER; /* save state */
+                                                 include_lineno[include_stack_ptr] = msyylineno;
+                                                 include_stack_ptr++;
 
                                                  msyy_switch_to_buffer( msyy_create_buffer(msyyin, YY_BUF_SIZE) );
                                                  msyylineno = 1;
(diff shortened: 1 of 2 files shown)


sdlime:
Thanks for fixing this @rouault! I can add a test or two.

landryb:
fwiw, just putting it as a comment here since i was about to file a similar issue against 8.0, but i've checked and it doesnt blow with 8.4, probably thanks to that PR. but since i spent the time writing the issue... might aswell have it somewhere in github :)

--------------------------
i'm having a 'funny' crash with 8.0.0 on debian, adding an INCLUDE to an existing included mapfile (which is INCLUDEd by a *main* mapfile, and already INCLUDEs several other files, which in turn INCLUDE other files..), suddenly fcgi workers started crashing at each call.. added debug symbols, and the trace is:
```
(gdb) bt
#0  0x00007f28cd6949a4 in fclose () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f28cd968433 in msyycleanup_includes () at ./obj-x86_64-linux-gnu/maplexer.l:744
#2  0x00007f28cd9b47e5 in msLoadMap (filename=filename@entry=0x5597f380d01b "/srv/mapserver/mapfile/rtge.map", new_mappath=new_mappath@entry=0x0, config=config@entry=0x5597f3808360) at ./mapfile.c:6331
#3  0x00007f28cd901a63 in msCGILoadMap (mapserv=0x5597f380d3a0, config=0x5597f3808360) at ./mapservutil.c:221
#4  0x00005597f2aae6e4 in ?? ()
#5  0x00007f28cd64624a in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#6  0x00007f28cd646305 in __libc_start_main () from /lib/x86_64-linux-gnu/libc.so.6
#7  0x00005597f2aae791 in ?? ()
```

puzzling a bit, so in the trace there's the `msyycleanup_includes` added in 52f1826c230448b which is in a 'failure to load' codepath.. but when i added the content of the included file in the mapfile at the same spot, things work.

until i figured it out.. the user running the fcgi process didnt have unix rights to read the newly INCLUDE'd file (`chmod +r` fixed it), thus it failed parsing. And if the file isn't opened, then `fclose()` shouldn't be called ? and proper error reporting should be propagated to the user ? i would have expected `Error opening included file \"%s\"."` to be returned safely, but for some reason that's not the case ..

---------------------------

my analysis after finding commit fd70c9933 is that `fclose()` isn't wrongly called anymore, since `fopen()` failed. with 8.4.0, a proper error message is reported:
```
loadLayer(): Premature End-of-File.
msyylex(): Unable to access file. Error opening included file &quot;common/postgis_connection_rtge_reader.map&quot;.
```

PR #6857, opened 2023-04-08
Force HTTP/1.1 for tests that query demo.mapserver.org to workaround current issue with it

rouault:


geographika:
Nice catch @rouault !
Is the `CURL_HTTP_VERSION` config option something that would be useful for other server setups, or is this a workaround just for the msautotests and the demo server?

rouault:
> or is this a workaround just for the msautotests and the demo server?

I just see this mostly as a workaround.

jmckenna:
nice catch @rouault thanks!

rouault:
@jmckenna Do you have hints on what might be wrong with HTTP/2 on the demo server ? Just van den Broeke pointed to https://serverfault.com/questions/937253/https-doesnt-work-with-safari  as possibly related .  The last comment in this page (https://serverfault.com/a/1068684) contains possible "fixes"

jmckenna:
@rouault I'll try those possible fixes today, thanks for this.

Issue #6837, opened 2023-02-17
Restructure repository, to move all source code into /src folder


jmckenna:
## Reasoning

- for the user, it is much easier to understand a repository when the general files (READMEs, markdown files, CMakelists.txt, .yaml files) only exist in the root directory, and the source files (.c, .h, cpp, etc) live in `/src`
- this is the common structure now, in FOSS4G projects (such as the QGIS repo)
- it is so much easier to come to a new repository for the first time, and understand it, as the important markdown files are easily found

## Proposed layout

```
MapServer/
      .github/
      ci/
      cmake/
      etc/
      fonts/
      fuzzers/
      msautotest/
      scripts/
      share/
      src/
          apache/
          flatgeobuf/
          mapscript/
          opengl/
          renderers/
          third-party/
          *.c
          *.cpp
          *.h
      symbols/
      tests/
      xmlmapfile/
      .travis.yml
      appveyor.yml
      CMakeLists.txt
      CONTRIBUTING.md
      HISTORY.md
      LICENSE.md
      README.md
      SECURITY.md
```
## Timing

I would love to do this for the imminent `8.0.1 release`, but this is a major impact to packagers, so instead I propose to tackle this for `8.2.0` instead.  (likely also include #6593  )

landryb:
fine with me as a packager whenever you want, and welcomed !

sebastic:
At first glance it seems like well intended but pointless busywork.

Git makes tracking renamed/moved files easier, but the diff for this source tree reorganisation will likely be a PITA to review. Fortunately the source tree for MapServer is relatively small (unlike QGIS for which I've already spent a few hours today reviewing the changes between 3.22.16 and 3.28.3 being nowhere near finished), so it might not be such a pain.

A source tree reorganisation like this is completely inappropriate for a patch release like 8.0.1, for a new minor release like 8.2.0 it's fine.

sdlime:
I actually like the idea but agree with @sebastic on timing. Might also be worth thinking about separating the apps (e.g. map2img, mapserv, etc...) from the library source. I see gdal has an "apps" directory for that purpose.

PR #6862, opened 2023-04-11
Allow Mapfiles to be accessed via URL Keys

geographika:
The OGC Features API introduced in MapServer 8.0 allows accessing Mapfiles using keys in the config file:

URL: http://example.com/mykey/ogcapi/collections/foo/items.

Config file:

```
CONFIG
  MAPS
    mykey "/path/to/mapfile"
  END
END
```

For all other requests you have to use the format http://example.com/?map=mykey&request=WMS...

This pull request allows the same approach as OGC Features API tbe used all other requests types e.g. a WMS request could be accessed using: http://example.com/mykey/?request=WMS...

Dev list thread at https://lists.osgeo.org/pipermail/mapserver-dev/2023-March/016951.html

### DIFF
--- a/mapserv.c
+++ b/mapserv.c
@@ -279,7 +279,8 @@ int main(int argc, char *argv[])
     }
 #endif
 
-    if(mapserv->request->api_path != NULL) {
+    if(mapserv->request->api_path != NULL && mapserv->request->api_path_length > 1) {
+      // API requests require a map key and a path
       if(msCGIDispatchAPIRequest(mapserv) != MS_SUCCESS) {
 	msCGIWriteError(mapserv);
 	goto end_request;
--- a/mapservutil.c
+++ b/mapservutil.c
@@ -351,7 +351,7 @@ int msCGIIsAPIRequest(mapservObj *mapserv)
           n++;
       }
 
-      if(n < 2) { // not enough non-zero length components for a legitimate API request
+      if(n < 1) { // we need at least a mapfile key for a legitimate request
         msFreeCharArray(tmp_api_path, tmp_api_path_length);
         return MS_FALSE;
       }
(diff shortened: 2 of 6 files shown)


jmckenna:
Thanks @geographika !  Are you ok if this gets added into the 8.0.1 release? Seems like a very useful change.   Thanks again.

sdlime:
This seems more appropriate for 8.2 IMHO. It's a nice add that could drive that release.

jmckenna:
roger, understood (by the way, now that I finally upgraded the demo server, will proceed with the 8.0.1 release this weekend)

jmckenna:
PS. I also look forward to the 8.2.0 release, but for https://github.com/MapServer/MapServer/issues/6837 (I guess we all have different motivation, ha)

sdlime:
> roger, understood (by the way, now that I finally upgraded the demo server, will proceed with the 8.0.1 release this weekend)

7.6.5 as well?

geographika:
Fine with either release, but it would be good to have a few "features" for 8.2 (#6795 is another). 
Once merged I'll update the docs.

pvgenuchten:
I wonder if the maps can also be substituted by an environment variable, so you could indicate from a docker compose which mapfiles to expose... (although you probably also mount the config file into the container)

geographika:
If there are no objections I'll be merging this in the next couple of days.

geographika:
> I wonder if the maps can also be substituted by an environment variable, so you could indicate from a docker compose which mapfiles to expose... (although you probably already mount a customised config file into the container)

I think it is probably simpler to generate different `CONFIG` files and then set the `MAPSERVER_CONFIG_FILE` environment variable in Docker Compose (or point to a single custom config file in the image).

PR #6863, opened 2023-04-15
PROJ_DATA set through config option: take into account possibility of…

rouault:
… multiple paths separated by ; on Windows or : on Unix

### DIFF
--- a/mapproject.c
+++ b/mapproject.c
@@ -37,6 +37,7 @@
 #include "mapaxisorder.h"
 
 #include "cpl_conv.h"
+#include "cpl_string.h"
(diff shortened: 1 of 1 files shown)


geographika:
Thanks @rouault ! I've tested this on a Windows machine, setting `PROJ_LIB` and `PROJ_DATA`) in a config file:

```
CONFIG
  ENV
	PROJ_LIB "C:/MapServer/bin/proj9/share;C:\Temp"
```

I then added a custom projection file in `C:\Temp` and this was correctly picked up by PROJ.

jmckenna:
Excellent, thanks @rouault !

PR #6871, opened 2023-04-22
CMake: Check compiler type for MSVC speicific option

Biswa96:
The '/w' option is specific to MSVC in Windows. So, check the compiler
    type before adding that compiler flag. This fixes the following error
    in mingw toolchain.

    cc: error: no such file or directory: '/w'


jmckenna:
thanks @Biswa96 !

PR #6873, opened 2023-04-22
cmake: Define STDC_HEADERS for regex in mingw toolchain

Biswa96:
Defining STDC_HEADERS macro includes stdlib.h in regex.c which is
    required for free, abort etc. function declarations. This fixes the
    following compiler error with mingw clang.

    regex.c:2042:3: error: call to undeclared library function 'free' with type
    'void (void *)'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
    regex.c:2784:11: error: call to undeclared library function 'abort' with type
    'void (void) __attribute__((noreturn))'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]


jmckenna:
thanks @Biswa96 !

PR #6881, opened 2023-05-25
cmake: respect prefix when installing mapserver-sample.conf

nilason:
Installation of `mapserver-sample.conf` does not respect prefix when installing. This
PR attempt to address this "misbehaviour".

This happens on e.g. MacPorts where the default prefix is `/opt/local` and it is expected 
to install the file at `/opt/local/etc/mapserver-sample.conf`.

With the present variable `CMAKE_INSTALL_FULL_SYSCONFDIR`, however,  it is installed
at `/etc/mapserver-sample.conf`.


jmckenna:
Thanks @nilason, I haven't hit this before, @sebastic does this look ok for your environments too?

sebastic:
> Thanks @nilason, I haven't hit this before, @sebastic does this look ok for your environments too?

For the Debian package build the file is already installed in `/etc`, but it's not included in any package:
```
-- Installing: /build/mapserver-8.0.1/debian/tmp/etc/mapserver-sample.conf
rm -f /build/mapserver-8.0.1/debian/tmp/etc/mapserver-sample.conf
```

nilason:
Thanks!

PR #6884, opened 2023-05-25
Allow expressions in LABEL PRIORITY

geographika:
This allows expressions to be added to a [LABEL PRIORITY](https://mapserver.org/mapfile/label.html#mapfile-label-priority) As well as attribute bindings and integer values. 

This is useful to allow labels to be prioritised on attributes dynamically e.g. calculate a 1-10 value based on population sizes of cities (by calculating a percentage of the highest value). 

The code is based on the [SLD pull request](https://github.com/MapServer/MapServer/pull/5831/commits/fc07dd9168d06a2fbca3e833d0ec56959c52449b) from @jbo-ads. 

There are also [comments](https://github.com/MapServer/MapServer/pull/5831#issuecomment-511933262) from @sdlime on the pull request about using this as a template pull request to allow expressions for other Mapfile values. 

Tests have been added for each of the 3 ways of setting a label priority - numeric value, attribute binding, and the new expression option.

Code reviews would be appreciated. I'm unsure why the following block is required (the SIZE pull request included this). Is it to clean-up expressions if a Mapfile is read a second time?

```
    if (label->exprBindings[MS_LABEL_BINDING_PRIORITY].string) {
        msFreeExpression(&label->exprBindings[MS_LABEL_BINDING_PRIORITY]);
        label->nexprbindings--;
    }
```

Docs to be updated if merged.


rouault:
LGTM

> I'm unsure why the following block is required (the SIZE pull request included this). Is it to clean-up expressions if a Mapfile is read a second time?

I believe this is just to avoid memory leaks if the PRIORITY keyword would appear several times in the same LABEL object.
(by the way my understanding of the work done by msFreeExpression() is that the .string member is freed and reinitialize to NULL, so the explicit msFree( .string) before ``label->exprBindings[MS_LABEL_BINDING_PRIORITY].string = msStrdup(msyystring_buffer)`` is very likely useless, but better keep it for consistency with similar code paths)

geographika:
> I believe this is just to avoid memory leaks if the PRIORITY keyword would appear several times in the same LABEL object. (by the way my understanding of the work done by msFreeExpression() is that the .string member is freed and reinitialize to NULL, so the explicit msFree( .string) before `label->exprBindings[MS_LABEL_BINDING_PRIORITY].string = msStrdup(msyystring_buffer)` is very likely useless, but better keep it for consistency with similar code paths)

Thanks @rouault for the review and clarification on the above - most appreciated. 
I aim to add expression support for [STYLE->SIZE](https://mapserver.org/mapfile/style.html#mapfile-style-size) in a future pull request.

geographika:
Just for reference @rouault is correct that the following block is used to clean-up any duplicate PRIORITY in a label:

```
    if (label->exprBindings[MS_LABEL_BINDING_PRIORITY].string) {
        msFreeExpression(&label->exprBindings[MS_LABEL_BINDING_PRIORITY]);
        label->nexprbindings--;
    }
```

In the case of a Mapfile containing two PRIORITY keywords the first is ignored (and cleaned-up with the above code), and the second expression is used by the label. I added this as part of the msautotest. 

```
LABEL
  PRIORITY ([priority] * 2)
  PRIORITY (10 - [priority])
  ...        
END

PR #6885, opened 2023-05-26
cmake: remove repeated status message

nilason:
Remove repeated configuration status message on `USE_GIF`.


jmckenna:
Nice catch @nilason, thanks

PR #6887, opened 2023-05-26
Coverity Scan fixes

rouault:
No real bugs spotted, but should help making Coverity Scan happier


jmckenna:
thanks @rouault !

rouault:
no need to backport this

PR #6888, opened 2023-05-26
Update Appveyor Python versions

geographika:
This pull request:

- Updates to new Python versions for testing. 3.7 will only be tested on tagged releases in Appveyor. It is EOL next month: https://endoflife.date/python
- Re-enables the "reprojection rect and datum shift" test. This was disabled due to failing previously (see #6824 and #6827), but now appears to be working again


jmckenna:
thanks @geographika

PR #6889, opened 2023-05-26
[Lint] OGCAPI: switch for #define to an emumeration for error types, …

rouault:
…to avoid potential misuses and safe extension of the error types


jmckenna:
thanks @rouault !

PR #6893, opened 2023-05-26
OGC API Features: implement Part 2 - CRS

rouault:
- support crs, bbox, bbox-crs request parameters
- use "oga_srs" metadata to populate the crs list. First item in the list is uses as the storageCrs


sdlime:
Cool @rouault, I'll try to make time to review this evening. --Steve

msautotest/api/ogcapi.map
sdlime, line 1: Should we add empty crs= and bbox-crs= tests?

rouault: > Should we add empty crs and bbox-crs parameter tests?

done in 2072356944443b69c2389b2a95a4c0fc39703589

> Should we add tests where crs and bbox-crs parameters are set multiple times in the same request?

I've had a bit of trouble to figure out what we should do in that situation, but I believe the answer comes indirectly from the fact that the OpenApi query parameters defined in the OGC API spec use "style: form" and "explode: false". Looking at https://swagger.io/docs/specification/serialization/, it means that if "foo=bar&foo=baz" was synthetized, then foo should be interpreted as an array [bar, baz] which doesn't make sense for the requests we are interested in . 

I've tested with a server which does seem to use a OpenApi based implementation:
```
$ curl "https://ogc-api.nrw.de/lika/v1/api?f=json&f=foo"

{
  "status" : 400,
  "title" : "Bad Request",
  "detail" : "Parameter value '[json, foo]' is invalid for parameter 'f': [$: does not have a value in the enumeration [html, json, yaml]]"
}
```

and even the request where it could potentially makes some "sense" (if interpreting literaly) fails:
```
$ curl "https://ogc-api.nrw.de/lika/v1/collections/verwaltungseinheit/items?bbox=2,49&bbox=2,49"
{
  "status" : 400,
  "title" : "Bad Request",
  "detail" : "Parameter value '[2,49, 2,49]' is invalid for parameter 'bbox': [$[0]: string found, number expected]"
```
So I've decided to go and detect this in a general way and error out in 8394797c8f41d86066d47975254337175b7e0b91

sdlime (approved):

msautotest/api/expected/ogcapi_api.json
```diff
@@ -1 +1 @@
...
...pi/ogcapi-features-1.yaml#/components/responses/InvalidParameter"},"404":{"$ref":"http://schemas.opengis.net/ogcapi/features/part1/1.0/openapi/ogcapi-features-1.yaml#/components/responses/NotFound"},"500":{"$ref":"http://schemas.opengis.net/ogcapi/features/part1/1.0/openapi/ogcapi-features-1.yaml#/components/responses/ServerError"}},"summary":"Get Minnesota Population Centers item by id","tags":["mn_population_centers"]}},"/conformance":{"get":{"description":"API conformance definition","operationId":"getConformanceDeclaration","parameters":[{"$ref":"#/components/parameters/f"}],"responses":{"200":{"$ref":"http://schemas.opengis.net/ogcapi/features/part1/1.0/openapi/ogcapi-features-1.yaml#/components/responses/ConformanceDeclaration"},"400":{"$ref":"http://schemas.opengis.net/ogcapi/features/part1/1.0/openapi/ogcapi-features-1.yaml#/components/responses/InvalidParameter"},"500":{"$ref":"http://schemas.opengis.net/ogcapi/features/part1/1.0/openapi/ogcapi-features-1.yaml#/components/responses/ServerError"}},"summary":"API conformance definition","tags":["server"]}}},"servers":[{"description":"Server description","url":"http://localhost/cgi-bin/mapserv/OGCAPI_TEST/ogcapi"}],"tags":[]}
```
sdlime, line 1: Do crs, crs-bbox, etc... need to be added to the parameters hash in the API JSON output? See https://demo.pygeoapi.io/stable/openapi?f=json for example.

rouault: They are added in https://github.com/MapServer/MapServer/pull/6893/files#diff-46e72d45fbe7adb5c2df20d7e4a963164b077f5505152841bdd6a2ce022ac42dR1771 as external references, as we do for bbox and datetime

Cf extract of output of ``mapserv -nh  -conf ../etc/mapserv.conf PATH_INFO=/ogcapi.map/ogcapi/api  "QUERY_STRING=f=json" | jq .`` :
```
    "/collections/mn_counties/items": {
      "get": {
        "description": "This is the standard Minnesota State County Boundary dataset.",
        "operationId": "getmn_countiesFeatures",
        "parameters": [
          {
            "$ref": "#/components/parameters/f"
          },
          {
            "$ref": "http://schemas.opengis.net/ogcapi/features/part1/1.0/openapi/ogcapi-features-1.yaml#/components/parameters/bbox"
          },
          {
            "$ref": "http://schemas.opengis.net/ogcapi/features/part1/1.0/openapi/ogcapi-features-1.yaml#/components/parameters/datetime"
          },
          {
            "$ref": "http://schemas.opengis.net/ogcapi/features/part2/1.0/openapi/ogcapi-features-2.yaml#/components/parameters/bbox-crs"
          },
          {
            "$ref": "http://schemas.opengis.net/ogcapi/features/part2/1.0/openapi/ogcapi-features-2.yaml#/components/parameters/crs"
          },
```

msautotest/api/ogcapi.map
sdlime: I think things look really good although admittedly there was a lot to review but things seems sensible with some nice efficiencies added to the code in general. 

The one thing missing would be changes to the HTML output templates - that could be a separate pull request completed by other devs. I would think that could be as simple as outputting the supported crs's and storageCrs when viewing a collection. As I understand it the crs stuff is implemented at the collection-level so we wouldn't need to expose it anywhere else in the HTML output. That's how the pyogcapi demo handles it. @tomkralidis, what do think?

One question I just though of is should the _links_ attribute for a collection take crs and/or storage crs parameters into account or should that be left to the user? Right now at the collection level we just output the links to fetch items as JSON or HTML. Feels unnecessary to be honest and bloats the response for no good reason.

sdlime:
@rouault, thanks. I noticed when I copied JSON elsewhere and did some searches in it. Sorry for the noise.

msautotest/api/ogcapi.map
rouault: At least for JSON output, given that this is an extension, and that resulting GeoJSON is not compatible with base GeoJSON readers when a non OGC:CRS84 CRS is used, I don't think crs={native-crs} should be included in the links. Looking at https://ogc-api.nrw.de/lika/v1/collections/verwaltungseinheit, they don't do that. This could potentially be an extra link, but I'm not sure we'd want to advertize it as "application/geo+json" as a media type.  If we were to implement https://docs.ogc.org/DRAFTS/21-045.html (which I'm going to work on for a GDAL driver), then we could use a "application/vnd.ogc.fg+json" as a media type for such non-CRS84 /items link

tomkralidis: > The one thing missing would be changes to the HTML output templates - that could be a separate pull request completed by other devs. I would think that could be as simple as outputting the supported crs's and storageCrs when viewing a collection. As I understand it the crs stuff is implemented at the collection-level so we wouldn't need to expose it anywhere else in the HTML output. That's how the pyogcapi demo handles it. @tomkralidis, what do think?
> 

Agree, it makes sense to put forth the CRS specifics at the collection level HTML.

> One question I just though of is should the _links_ attribute for a collection take crs and/or storage crs parameters into account or should that be left to the user? Right now at the collection level we just output the links to fetch items as JSON or HTML. Feels unnecessary to be honest and bloats the response for no good reason.

Agree with @rouault not to put CRS specifics in links.

PR #6895, opened 2023-05-29
Allow expressions to be used on a STYLE SIZE

geographika:
Following on from #6884 this pull request allows expressions to be used in a STYLE SIZE. 
For example:

```
        CLASS
            STYLE
                SYMBOL "circle"
                SIZE ([priority] *2) # double the priority attribute to calculate the size
                COLOR 0 0 255
            END
        END
```

This has a number of real-world use cases e.g. the natural world has a city dataset, which contains a LABELRANK attribute. The attribute itself cannot be used to set the size as there are too many very low values such as 1-4 which would not be visible. 
However the attribute could be subtracted from 15 to give good results. Previously the `DATA` clause would need to be modified and a new attribute created dynamically to symbolise on (therefore mixing data and styling):

```
    DATA "SELECT *, 15-LABELRANK AS POINTSIZE FROM ne_110m_populated_places"
    CLASS
        STYLE
            SYMBOL "circlef"
            COLOR "#6A5ACD"
            SIZE [POINTSIZE]
        END
```

With this pull request the following can be used:

```
    DATA "ne_110m_populated_places"
    CLASS
        STYLE
            SYMBOL "circlef"
            COLOR "#6A5ACD"
            SIZE (15 -[LABELRANK])
        END
```

Another example is when using `CLUSTER`. With this pull request the cluster count could be used for styling as part of an expression:

```
  CLASS
      STYLE
          SYMBOL "circlef"
          COLOR "#6A5ACD"
          SIZE [Cluster_FeatureCount] * 3
      END
```

Without expressions I'm not sure how the above is possible. 

Tests are included as part of the pull request, including duplicate keyword to check there are no memory leaks. There was also no tests to check binding an attribute on `SIZE` so this has been added. 

Finally only the mapfile.c file needed changes. Expressions for STYLE SIZE were already supported in `bindStyle` as they were added as part of https://github.com/MapServer/MapServer/pull/5831


geographika:
@jmckenna MapScript doesn't currently support setting expressions on any properties where other values can be used (e.g. string, integers etc.). This applies to LABEL>SIZE, PRIORITY, as well as STYLE>SIZE. Any objects can however always be loaded, with expressions, using `mapscript.msLoadMapFromString`. 

I can think of a possible way to implement this in MapScript, but it will be in a separate pull request. 

The SLD work also added lots of other properties that can be bound to an attribute using SLD, but not yet via a Mapfile e.g. STYLE>WIDTH. It might be worth allowing these to be supported by in a Mapfile.

PR #6897, opened 2023-05-31
Refresh cached reprojector when changing map projection between drawMap() calls (fixes #6896)

### LINKED ISSUE #6896: Empty image generated after changing projection via MapScript when data in WGS84
Hi,

If the data is in WGS84 the map projection can only be changed once. More changes result in empty images being generated.

I have attached a ZIP file, [test-change-projection.zip](https://github.com/MapServer/MapServer/files/11593570/test-change-projection.zip), containing two example maps, one where the Shape file is in WGS84 and one in SWEREF99 TM (meters). When the following MapScript written in C# is executed on these maps the one having data in WGS84 return an empty image2.png while the map having data in SWEREF 99 TM returns two correct pictures.

``` c#   
using (mapObj map = new mapObj(Path.Combine(path, "test.map")))
{
    map.setProjection("init=epsg:3857");
    map.setSize(256, 256);
    map.setExtent(2003814.264075839, 9577313.687016932, 2003892.9344537726, 9577392.45292971);
    using (imageObj image1 = map.draw())
    {
        image1.save(Path.Combine(path, "image1.png"), map);
    }

    map.setProjection("init=epsg:3006");
    map.setSize(256, 256);
    map.setExtent(642080, 7198656, 642112, 7198688);
    using (imageObj image2 = map.draw())
    {
        image2.save(Path.Combine(path, "image2.png"), map);
    }
}
```

This is a new issue since 8.0.x, the test case above have no issues in 7.6.x. I have performed the tests on Windows 10.

Have I missed something new that I have to do when changing projection using MapServer 8? We generate images in multiple projections from the same map file and for performance reasons do not want to reload the map each time the projection change.

Regards
/Krister

rouault:


geographika:
Thanks for the fix! Just a question on this - is this issue only applicable when changing projections through MapScript, or could the same issue have been present when using CGI or FastCGI requests?

rouault:
> or could the same issue have been present when using CGI or FastCGI requests?

the mapfile is reloaded for each request as far as I remember

jmckenna:
Also added my local PHP MapScript tests used earlier.

rouault:
> Also added my local PHP MapScript tests used earlier.

is that strictly necessary ? I've added an equivalent test in the python test suite (which actually doesn't require new "golden" files), and the fix is not language specific, but in core, so it doesn't need to be tested by all bindings. While adding tests is a good practice, I think we should also try to avoid redundant tests to limit the maintenance burden.

jmckenna:
@rouault Yes, in fact these additional tests are the first inside PHP MapScript to test setProjection, and compare map images generated through PHP MapScript.

jmckenna:
I disagree on redundancy (I did anticipate your pushback here), as in one of the previous tickets it was stated by a PSC member (I wish we had multiple language tests for this issue), well, and here it is, which I think is excellent.  In fact I feel opposite, we nee to stop only testing one single language, and for those language we do not test we should remove that language.

jmckenna:
In other words, when adding a MapScript test, we should be adding a test for each of the 4 or so languages that MapServer supports (this is from experience of the minor differences in how SWIG etc handles languages).

I am happy to discuss this in the next PSC meeting, if in a ticket is too much noise.

jmckenna:
I am also happy to volunteer to handle the MapScript tests (it should be straight forward to test as I can just now use these containers to test).  In the end, this will make a more solid 'product' of MapServer, for the users of our mapscript languages.

rouault:
> In other words, when adding a MapScript test, we should be adding a test for each of the 4 or so languages that MapServer supports (this is from experience of the minor differences in how SWIG etc handles languages).

There's a difference between testing that setProjection() works in all languages, which is a legitimate goal, and testing that the scenario of a particular bug (#6896), which is a bug of core, is tested in all languages, which could become a maintenance burden over the long term (more mapfiles to maintain etc). Changes of PROJ / EPSG database versions might affect reprojection results and require refreshing golden results. We should be careful about the choices of CRS we make to limit our exposure to such changes. Typically you'd want to remain in the same datum (for example EPSG:4326 and EPSG:3857 that are both based on WGS 84)
The test__setProjection() of PHP could just call setProjection() once. It doesn't need to do the multiple calls of #6896

The testing strategy can indeed be discussed at a PSC meeting

jmckenna:
thanks again @rouault for the quick (and tricky) fix.

PR #6898, opened 2023-05-31
Ensure shapes aren't reprojected twice when cached and output to OGR

geographika:
A bug I noticed when working on #6731 - if WFS caching is used, and results are output to an OGR format then any features are reprojected twice. Once during the query phase and then again when retrieved from the cache. 

Query reprojection: https://github.com/MapServer/MapServer/blob/db74212ec9d16030870bc4a280771febfada8091/mapquery.c#L1251
Output reprojection: https://github.com/MapServer/MapServer/blob/db74212ec9d16030870bc4a280771febfada8091/mapogroutput.cpp#L1173

This pull request ensures features read from the cache aren't reprojected. 

```
WEB
  METADATA
    "wfs_features_cache_count" "3"
    "wfs_features_cache_size" "10000" # bytes
```

One of the tests already highlighted this issue - the first 3 features in `wfs_query_cache.json` have very different coordinates from the other features. With the fix the test has been corrected. I also added a WFS 2.0 test as part of the pull request. 

For reference the source dataset is in the https://epsg.io/42304 projection. For comparing features I manually reprojected using the `ogr2ogr` command:

```
ogr2ogr -s_srs "+proj=lcc +lat_0=49 +lon_0=-95 +lat_1=49 +lat_2=77 +x_0=0 +y_0=0 +datum=NAD83 +units=m +no_defs +type=crs" -t_srs epsg:4326 .\msautotest\wxs\data\province4326.shp .\msautotest\wxs\data\province.shp```

### CODE REVIEW

File: msautotest/wxs/wfs_query_cache.map [hunk 1]
```diff
@@ -8,6 +8,10 @@
 
 # RUN_PARMS: wfs_query_cache.json [MAPSERV] QUERY_STRING="map=[MAPFILE]&SERVICE=WFS&VERSION=1.0.0&REQUEST=GetFeature&TYPENAME=province&MAXFEATURES=10&OUTPUTFORMAT=geojson" > [RESULT]
 
+# RUN_PARMS: wfs2_query_cache.json [MAPSERV] "QUERY_STRING=map=[MAPFILE]&SERVICE=WFS&VERSION=2.0.0&REQUEST=GetFeature&TYPENAMES=province&MAXFEATURES=10&OUTPUTFORMAT=geojson" > [RESULT]
+
+# RUN_PARMS: wfs2_query_cache_no_default_extent.json [MAPSERV] "QUERY_STRING=map=[MAPFILE]&SERVICE=WFS&VERSION=2.0.0&REQUEST=GetFeature&TYPENAMES=province2&MAXFEATURES=10&OUTPUTFORMAT=geojson" > [RESULT]
```
rouault, line 13: is this test + added layer really needed with respect to what you changed ? (just for the sake of minimizing things)

geographika: I was worried setting `"wfs_use_default_extent_for_getfeature" "false"` might have skipped the `msQueryByRect` function, and then there could be features in the cache which weren't reprojected. This is not the case as a `MS_INIT_INVALID_RECT` is set and used in the `msQueryByRect`, but thought it might be good to keep this test in case code is refactored in the future.

rouault: ok, fair point

PR #6904, opened 2023-06-06
Allow Expressions to be set for LABEL properties using MapScript

geographika:
This pull request allows the new functionality in #6884 of using an expression in a LABEL PRIORITY in MapScript. 
It takes a similar approach to setting attribute bindings for properties, which requires using the appropriate attribute constant such as `MS_LABEL_BINDING_PRIORITY`:

```python
label = mapscript.labelObj()
label .setExpressionBinding(mapscript.MS_LABEL_BINDING_PRIORITY, "[MY_ATTRIBUTE] * 2")
```

This is a little clunky but is at least consistent with setting attribute bindings:

```python
label = mapscript.labelObj()
label.setBinding(mapscript.MS_LABEL_BINDING_COLOR, "NEW_BINDING")
```

While testing this I noticed that the `writeLabel` function used to write Mapfile objects ignored expressions. This seems an oversight also for the `LABEL` `SIZE` expression. If this approach seems appropriate I can use this for other expression bindings.

Something to note is that a property could have an attribute binding **and** an expression binding set. In this case the attribute binding currently takes precedence when calculating a shape value, so it also takes precedence when writing to a Mapfile. 

In theory setting an expression could also set the attribute binding to NULL. This is not currently the case for MapScript or when reading a Mapfile twice.

### DIFF
--- a/mapscript/swiginc/label.i
+++ b/mapscript/swiginc/label.i
@@ -112,7 +112,51 @@
     return MS_SUCCESS;
   }
 
-  /// Set the label expression.
+  /// Remove an expression binding for a specified label property.
+  int removeExpressionBinding(int binding) 
+  {
+    if(binding < 0 || binding >= MS_LABEL_BINDING_LENGTH) return MS_FAILURE;
+
+    if(self->exprBindings[binding].string) {
+      msFreeExpression(&self->exprBindings[binding]);
+      self->nexprbindings--;
+    }
+
+    return MS_SUCCESS;
+  }
+
+  /// Get the expression binding for a specified label property. Returns NULL if there is no binding for this property.
+  %newobject getExpressionBinding;
+  char *getExpressionBinding(int binding) 
+  {
+    if(binding < 0 || binding >= MS_LABEL_BINDING_LENGTH) return NULL;
+
+    return msGetExpressionString(&(self->exprBindings[binding]));
+  }
+
+  /// Set the expression binding for a specified label property. Binding constants look like this: ``MS_LABEL_BINDING_[attribute name]``:
+  /// >>> new_label.setExpressionBinding(MS_LABEL_BINDING_PRIORITY, "([priority] * 2)")
+  int setExpressionBinding(int binding, char *text) 
+  {
+    if (!text || strlen(text) == 0) {
+      return MS_FAILURE;
+    }
+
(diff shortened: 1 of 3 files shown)


mapscript/swiginc/label.i
```diff
@@ -112,7 +112,52 @@
...
+      msFreeExpression(&self->exprBindings[binding]);
+      self->nexprbindings--;
+    }
+
+    return MS_SUCCESS;
+  }
+
+  /// Get the expression binding for a specified label property. Returns NULL if there is no binding for this property.
+  %newobject getExpressionBinding;
+  char *getExpressionBinding(int binding)
```
rouault, line 130: should probably be:
```suggestion
  const char *getExpressionBinding(int binding) 
```

mapscript/swiginc/label.i
```diff
@@ -112,7 +112,52 @@
...
+  {
+    if(binding < 0 || binding >= MS_LABEL_BINDING_LENGTH) return NULL;
+
+    return msGetExpressionString(&(self->exprBindings[binding]));
+  }
+
+  /// Set the expression binding for a specified label property. Binding constants look like this: ``MS_LABEL_BINDING_[attribute name]``
+  /// Expressions are automatically wrapped in brackets, so do not need to be added to the input string
+  /// >>> new_label.setExpressionBinding(MS_LABEL_BINDING_PRIORITY, "[priority] * 2")
+  int setExpressionBinding(int binding, char *text)
```
rouault, line 140:
```suggestion
  int setExpressionBinding(int binding, const char *text) 
```

geographika:
@rouault - thanks for the review - changes applied.
I presume it would make sense to use `const` for all similar functions throughout MapScript? 

E.g. https://github.com/MapServer/MapServer/blob/257c646a77eafa1f55c7d42448e60714cb211fa5/mapscript/swiginc/label.i#L127

https://github.com/MapServer/MapServer/blob/257c646a77eafa1f55c7d42448e60714cb211fa5/mapscript/swiginc/label.i#L116

https://github.com/MapServer/MapServer/blob/257c646a77eafa1f55c7d42448e60714cb211fa5/mapscript/swiginc/label.i#L142

rouault:
> I presume it would make sense to use `const` for all similar functions throughout MapScript?

yes

PR #6919, opened 2023-08-08
Repo restructure

jmckenna:
- related to plan mentioned through https://github.com/MapServer/MapServer/issues/6837
- resulting tree visible at https://github.com/jmckenna/MapServer/tree/repo-restructure-8.2-release

## Layout

```
MapServer/
      .github/
      ci/
      cmake/
      etc/
      fonts/
      fuzzers/
      msautotest/
      scripts/
      share/
      src/
          apache/
          apps/
            coshp.*
            legend.*
            map2img.*
            mapserv.*
            msencrypt.*
            scalebar.*
            shptree*.*
            sortshp.*
            tile4ms.*
          flatgeobuf/
          mapscript/
          opengl/
          renderers/
          third-party/
          xmlmapfile/
          *.c
          *.cpp
          *.h
      symbols/
      tests/
      .travis.yml
      appveyor.yml
      CMakeLists.txt
      CONTRIBUTING.md
      HISTORY.md
      LICENSE.md
      README.md
      SECURITY.md
```

PR #6920, opened 2023-08-10
Create unique output files for Mapfile by key tests

geographika:
As suggested by @sdlime in #6919, and to avoid overwriting test output, unique copies of expected results have been added to the CONFIG tests.


jmckenna (approved):

jmckenna:
thanks @geographika !

PR #6921, opened 2023-08-11
Allow Mapfile key to be only element in PATH_INFO

geographika:
Follow-up to #6862 - where a missing slash was causing tests to fail in #6919. 
This allows Mapfile keys to be used in a URL path, with or without a trailing slash:

```
http://example.com/mykey/?request=WMS...
http://example.com/mykey?request=WMS...
```

An OGCAPI test is also updated as part of this pull request as it is now valid to use a URL without an additional `ogcapi` in the path. 

I'm unsure why this test is currently passing in main but not when the repo structure was changed. @jmckenna - could you merge this and test with a rebase on #6919?


jmckenna:
thanks @geographika will test through rebase etc   Thanks again

Issue #6912, opened 2023-07-04
Wrong scale computed in tile mode


mikitex70:
## Expected behavior and actual behavior.

Having a layer with `MAXSCALEDENOM` (whatever values are used) and requesting tiles (cgi request with `mode=tile&tilemode=gmap`, but I think the same is with `tilemode=ve`) the layer is never drawn.
Enabling logging show a message like:
```
Skipping layer (fogli) because LAYER.MAXSCALE is too small for this MAP scale
```

Debugging shows that, for example, for a zoom level `13` the computed scale was approximately `8025436685` (too large) instead of about `72224`.

It seems that in `maptile.c`, the function `msTileSetExtent` is missing a `map->units = MS_METERS;` since the calculated extent is now in meters.

Applied the following patch and now seems to work correctly:
```patch
--- a/maptile.c	2023-07-03 15:35:40.032816983 +0200
+++ b/maptile.c	2023-07-03 16:02:17.224862070 +0200
@@ -495,6 +495,8 @@
     msDebug( "msTileSetExtent (%f, %f) (%f, %f)\n", map->extent.minx, map->extent.miny, map->extent.maxx, map->extent.maxy);
   }
 
+  map->units = MS_METERS;  // now the units are meters
+
   return MS_SUCCESS;
 #else
   msSetError(MS_CGIERR, "Tile API is not available.", "msTileSetExtent()");
```

## Steps to reproduce the problem.

Execute the `test.sh` included in the attached zip file in  an environment with Docker (used only to build a test environment).
Or execute a command like:
```shell
mapserv -nh "QUERY_STRING=map=test.map&mode=tile&tilemode=gmap&tile=X+Y+Z&layers=testLayer" > out.jpg
```
using whatever map and tile coordinates you want and with the `testLayer` having the `MAXSCALEDENOM` set to any value.

## Attach simple test case
[tilescale-test.zip](https://github.com/MapServer/MapServer/files/11947725/tilescale-test.zip)

Run the `test.sh` script; it requires Docker to build a test envirtonment, but commenting the `docker build` and the `alias` commands it should work with the system installation of the `mapserv`.

## Operating system

Tested on CentOS 6.7 64 bit and Ubuntu 22.04 64 bit.

## MapServer version and installation method

MapServer 7.6.4 from Ubuntu repository.
MapServer version 7.6.5, self-compiled to enable the Oracle driver.

Not tested but seems that MapServer 8 is also affected by this issue.

[...]

PR #6925, opened 2023-08-20
Fixed tile scale computation (#6912)

mikitex70:
Fixed tile scale computation when the `UNITS` directive is used in the map (for example `UNITS dd`).
Added some test  to verify that the scale are computed correctly.


msautotest/misc/tiles_scales_z2.map
```diff
@@ -0,0 +1,43 @@
...
+
+  LAYER
+    DEBUG 5
+    NAME "canada-poly"
+    TYPE POLYGON
+    STATUS DEFAULT
+    CONNECTIONTYPE OGR
+    CONNECTION "data/canada.dgn"
+    DATA "elements"
+    MINSCALE 30000000
```
geographika, line 30: `MINSCALE` and `MAXSCALE` are deprecated. These should be replaced with `MINSCALEDENOM` and `MAXSCALEDENOM`.

msautotest/misc/tiles_scales_z3.map
```diff
@@ -0,0 +1,43 @@
...
+  LAYER
+    DEBUG 5
+    NAME "canada-poly"
+    TYPE POLYGON
+    STATUS DEFAULT
+    CONNECTIONTYPE OGR
+    CONNECTION "data/canada.dgn"
+    DATA "elements"
+    MINSCALE 30000000
+    MAXSCALE 70000000
```
geographika, line 31: As above

msautotest/misc/tiles_scales_z4.map
```diff
@@ -0,0 +1,43 @@
...
+  LAYER
+    DEBUG 5
+    NAME "canada-poly"
+    TYPE POLYGON
+    STATUS DEFAULT
+    CONNECTIONTYPE OGR
+    CONNECTION "data/canada.dgn"
+    DATA "elements"
+    MINSCALE 30000000
+    MAXSCALE 70000000
```
geographika, line 31: As above

geographika (changes requested):

geographika:
Thanks @mikitex70 for this, and for adding in the tests. 

For the tests could you add in 2 new images files (even if they are duplicates of `tile_scale_empty.png`) with names that match the test Mapfile? This is to ensure if a test fails it doesn't pick up a previously created file of the same name (see discussions in #6919).

geographika (approved):

jmckenna:
thanks @mikitex70 !

PR #6928, opened 2023-09-07
A few code cleanups

rouault:


jmckenna:
Thanks @rouault (I had to re-run one of the actions due to a network issue).  I take it that this cannot be backported.  Squashing now...

rouault:
> I take it that this cannot be backported

probably not worth it indeed, but https://github.com/MapServer/MapServer/pull/6928/commits/51742f481a6623190cbfbd046781271aa692e11b . I'll cherry-pick it

> Squashing now...

well, I would have prefered to get them merged as individual commits, but that's ok.
Squashing is useful/needed when the commits do not result each in a "clean" build. Here they could be taken individually

PR #6932, opened 2023-09-11
reimplementation of imgageObj::pasteImage() and imageObj::saveWebImage()

jbfrd:


geographika:
In the comments for pasteImage is "NOTE : this function only works for 8 bits GD images." - GD support was dropped several versions ago. Is the function out-of-date, or the comment?

We'd want to add tests to the MapScript test suites before adding in.

jbfrd:
> In the comments for pasteImage is "NOTE : this function only works for 8 bits GD images." - GD support was dropped several versions ago. Is the function out-of-date, or the comment?

pasteImage() is a reimplementation of `PHP_METHOD(imageObj, pasteImage)` from `git show 3a100ab0a^:mapscript/php/image.c`, where the note is still included. So, cause the actual image handling is done with `renderer->getRasterBufferHandle()` and `renderer->mergeRasterBuffer()`, it depends on this functions if the note is still valid. But as you mention "GD support was dropped several versions ago.", I think the comment is out-dated. Please drop it.

geographika:
Replaced by #7015

PR #6940, opened 2023-10-06
WMS client (cascaded WMS): restrict the list of GDAL drivers used to …

rouault:
…read the GetMap response; add a 'wms_allowed_gdal_drivers' LAYER.METADATA option that takes a comma separated list of GDAL driver names


jmckenna:
thanks @rouault !

PR #6941, opened 2023-10-08
WFS 2 GetFeature with stored query and GET KVP interface: make sure that values of substituted parameters are XML encoded

rouault:


geographika:
@rouault - is this related to the issue recently reported at https://lists.osgeo.org/pipermail/mapserver-users/2023-October/083244.html ?

rouault:
> is this related to the issue recently reported

no it won't solve it. But I found it when reviewing the code while trying to reproduce the issue of the list (which I didn't manage to)

Issue #6911, opened 2023-06-21
List Expressions not output to SLD


geographika:
## Expected behavior and actual behavior.

Class list expressions such as that below aren't output to SLD files:

```
CLASS
  EXPRESSION {70,960,00,17,13940}
```

## Steps to reproduce the problem.

Add the following block to the listexpression.map file in msautotests:

```
WEB
  METADATA
      "ows_enable_request" "*"
  END
END
```

Next run the following to output SLD:

```
cd ./mapserver/msautotest/misc
SET MAPSERVER_CONFIG_FILE=mapserver.conf
mapserv QUERY_STRING="map=listexpression.map&SERVICE=WMS&VERSION=1.3.0&REQUEST=GetStyles&LAYERS=charts"
```

The resultant SLD does not include the list epxression:

```xml
    <se:PolygonSymbolizer>
      <se:Fill>
        <se:SvgParameter name="fill">#ff0000</se:SvgParameter>
      </se:Fill>
    </se:PolygonSymbolizer>
```

It should include a filter expression similar to:

```xml
<ogc:Or>
   <ogc:PropertyIsEqualTo>
   <ogc:PropertyName>V1</ogc:PropertyName>
   <ogc:Literal>70</ogc:Literal>
   </ogc:PropertyIsEqualTo>
   <ogc:PropertyIsEqualTo>
   <ogc:PropertyName>V1</ogc:PropertyName>
   <ogc:Literal>60</ogc:Literal>
   </ogc:PropertyIsEqualTo>
   </ogc:Filter>
 </ogc:Or>
```

PR #6943, opened 2023-10-09
 Include list expressions in GetStyles SLD output

geographika:
Fix for #6911. This update generates SLD for a class using [list expressions](https://www.mapserver.org/mapfile/expressions.html#list-expressions):

```
  CLASSITEM "FID"
  CLASS
    EXPRESSION {2,4,6}
    STYLE
        COLOR 220 0 0
        WIDTH 1
    END
```

Will include the following SLD with a request such as `map=mymap&SERVICE=WMS&VERSION=1.3.0&REQUEST=GetStyles&layers=MyLayer`

```xml
  <ogc:Filter>
    <ogc:Or>
      <ogc:PropertyIsEqualTo>
        <ogc:PropertyName>FID</ogc:PropertyName>
        <ogc:Literal>2</ogc:Literal>
      </ogc:PropertyIsEqualTo>
      <ogc:PropertyIsEqualTo>
        <ogc:PropertyName>FID</ogc:PropertyName>
        <ogc:Literal>4</ogc:Literal>
      </ogc:PropertyIsEqualTo>
      <ogc:PropertyIsEqualTo>
        <ogc:PropertyName>FID</ogc:PropertyName>
        <ogc:Literal>6</ogc:Literal>
      </ogc:PropertyIsEqualTo>
    </ogc:Or>
  </ogc:Filter>
```

Note when a a list expression is used a `CLASSITEM` has to be set on the layer. 
Tests include lists with 0 and 1 items - the first will not produce a filter, and the second will produce a single filter. When multiple items are in the list (the typical use case) multiple filters will be produced wrapped in an OR tag.

### DIFF
--- a/src/mapogcsld.c
+++ b/src/mapogcsld.c
@@ -5389,22 +5389,34 @@ char *msSLDGetFilter(classObj *psClass, const char *pszWfsFilter) {
 
         char *pszTmp = NULL;
         char *pszTmpFilters = NULL;
-        char *token = strtok((char *)psClass->expression.string, ",");
+        int nArgs = 0;
+        int i = 0;
         int tokenCount = 0;
+        char **papszArgs;
+
+        papszArgs =
+            msStringTokenize(psClass->expression.string, ",", &nArgs, MS_TRUE);
 
         // loop through all values in the list and create a PropertyIsEqualTo
         // for each value
-        while (token != NULL) {
-          tokenCount++;
+        for (i = 0; i < nArgs; i++) {
+
+          if (strlen(papszArgs[i]) == 0) {
+            free(papszArgs[i]);
+            continue;
+          }
+
           snprintf(szBuffer, sizeof(szBuffer),
                    "<ogc:PropertyIsEqualTo><ogc:PropertyName>%s</"
                    "ogc:PropertyName><ogc:Literal>%s</ogc:Literal></"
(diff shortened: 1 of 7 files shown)


src/mapogcsld.c
```diff
@@ -5383,6 +5383,49 @@ char *msSLDGetFilter(classObj *psClass, const char *pszWfsFilter) {
     } else if (psClass->expression.type == MS_EXPRESSION) {
       pszFilter =
           msSLDParseLogicalExpression(psClass->expression.string, pszWfsFilter);
+    } else if (psClass->expression.type == MS_LIST) {
+      if (psClass->layer && psClass->layer->classitem &&
+          psClass->expression.string) {
+
+        char *pszTmp = NULL;
+        char *pszTmpFilters = NULL;
+        char *token = strtok((char *)psClass->expression.string, ",");
```
rouault, line 5392: strtok() is not thread-safe.  The alternative would be strok_r() on Unix and strtok_s() on Windows. But better use msStringTokenize()

geographika: Thanks for the review. Updated to use `msStringTokenize` (although it is only used in one other function in MapServer). Would `msStringSplit` also suffice? List expressions "do not support quote escaping, or attribute values that contain a comma in them".

rouault: > Would `msStringSplit` also suffice?

yes, probably

PR #6947, opened 2023-10-12
msSLDGetFilter(): add missing XML escaping of property names and literals

rouault:
CC @geographika  . Something I noticed when reviewing https://github.com/MapServer/MapServer/pull/6943


geographika:
Thanks for this @rouault - I should have realised the quotes in the XML needed escaping when I made the test output. There are a few other SLD generation issues I'm going to look at - the switch to C++ will make working with strings easier.

PR #6948, opened 2023-10-17
CI: workaround add-apt-repository ppa:ubuntugis/ubuntugis-unstable failures

rouault:
Port of GDAL similar changes of https://github.com/OSGeo/gdal/pull/8571. Credits to @lnicola


jmckenna:
thanks @rouault, I hadn't seen this pop-up here yet, but it sounds like we would eventually..

PR #6949, opened 2023-10-19
Add msAddTrailingSlash function and use for template path

geographika:
An update to the discussion at https://github.com/MapServer/MapServer/pull/6400#issuecomment-927093583

Currently the following Mapfile would fail to find a template file as there is no "/" at the end of the path. This update automatically adds this in. On Windows in particular it is strange to see a string such as "C:\ogcapi\templates\html-plain\". 
```
  WEB
    METADATA
      "oga_html_template_directory" "../../share/ogcapi/templates/html-plain"
```
- added a new `msAddTrailingSlash` function to `mapstring.cpp` that can check any string (typically a path) has a trailing slash
- updated one of the test maps using the OGC Features API to have no trailing slash so the function is tested

A review is appreciated especially around use of `msSmallMalloc` - which normally should be freed. I initially had a C++ function but tried to follow a more MapServer-style with this approach.

### DIFF
--- a/msautotest/api/6631.map
+++ b/msautotest/api/6631.map
@@ -12,7 +12,7 @@ MAP
 
   WEB
     METADATA
-      "oga_html_template_directory" "../../share/ogcapi/templates/html-plain/"
+      "oga_html_template_directory" "../../share/ogcapi/templates/html-plain"
       "oga_title"          "Issue #6631" ## REQUIRED
       "oga_description"	   "This is a test OGC API server description for issue #6631."
       "oga_contactperson"  "Contact person"
--- a/src/mapogcapi.cpp
+++ b/src/mapogcapi.cpp
@@ -360,16 +360,27 @@ static bool getBbox(mapObj *map, layerObj *layer, cgiRequestObj *request,
 /*
 ** Returns the template directory location or NULL if it isn't set.
 */
-static const char *getTemplateDirectory(mapObj *map, const char *key,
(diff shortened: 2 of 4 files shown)


rouault:
> I initially had a C++ function but tried to follow a more MapServer-style with this approach.

I'd go for the C++ version that returns a std::string, as it would be simpler and your implementation actually leaks memory

getTemplateDirectory() should also be modified to return a std::string (empty if it fails)

jmckenna:
Thanks @geographika.  This was part of my initial report to Steve (it seems he deleted that branch and history in his fork), the slashes.   Thanks for fixing.

geographika:
@rouault  - updated to use C++ approach and locally to `getTemplateDirectory()` function. The calling function has also been updated to work with `std::string`.

jmckenna:
(related discussion happening in https://github.com/geopython/pygeoapi/issues/97#issuecomment-1789149696 )

geographika:
The pygeoapi discusses slashes in URLs used by the OGC APIs - this change was to add slashes for the template local path on the operating system. AFAIK MapServer treats the OGC API URLs with and without a slash the same.

jmckenna:
@geographika I am aware of the differences in the topics, however my early tests (my long report was deleted through the original sdlime's repo) had problems with the slashes in ogcapi requests, and were not tackled.  I'll be sure to re-test of course.  Someday that thread may come in handy.  I'll be doing my best to re-test this throughout.  

Sorry for the noise.

Issue #6403, opened 2021-09-06
Internal server error using WFS GetFeature with resulttype=hits and outputformat=geojson


fsteggink:
## Expected behavior and actual behavior.

The following request returns an Internal Server Error:
http://localhost:9906/maps/panden_plus2?service=WFS&version=2.0.0&request=GetFeature&typename=pand&bbox=130000,450000,140000,460000&resulttype=hits&outputformat=geojson
(The CRS is EPSG:28992, but adding this doesn't make a difference.)

I was expecting to get a JSON output indicating somehow, the feature count, although I don't even know if this is supposed to work. At least an Internal Server Error should not be thrown.

The logging to stdout/stderr shows this:
```
[Mon Sep 06 15:30:54.252990 2021] [fcgid:warn] [pid 4120:tid 140526833628928] [client 192.168.16.1:51382] mod_fcgid: error reading data, FastCGI server closed connection
[Mon Sep 06 15:30:54.253043 2021] [core:error] [pid 4120:tid 140526833628928] [client 192.168.16.1:51382] End of script output before headers: mapserv.fcgi
192.168.16.1 - - [06/Sep/2021:15:30:50 +0200] "GET /maps/panden_plus2?service=WFS&version=2.0.0&request=GetFeature&typename=pand&bbox=130000,450000,140000,460000&resulttype=hits&outputformat=geojson HTTP/1.1" 500 908 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:91.0) Gecko/20100101 Firefox/91.0"
[Mon Sep 06 15:30:54.935749 2021] [fcgid:error] [pid 4118:tid 140527389385856] mod_fcgid: process /usr/lib/cgi-bin/mapserv.fcgi(4176) exit(communication error), get unexpected signal 11
```

Relevant MapFile config:
```
  DEBUG 5
  CONFIG "MS_ERRORFILE" "stderr"
```

The following request returns a WFS FeatureCollection indicating the number of features:
http://localhost:9906/maps/panden_plus2?service=WFS&version=2.0.0&request=GetFeature&typename=pand&bbox=130000,450000,140000,460000&resulttype=hits

The only difference is that the `outputformat` parameter is omitted.

Result:
```xml
<?xml version='1.0' encoding="UTF-8" ?>
<wfs:FeatureCollection
   xmlns:panden="<redacted>"
   xmlns:gml="http://www.opengis.net/gml/3.2"
[...]

PR #6960, opened 2023-10-31
Avoid segmentation fault when outputformat is defined for resultType=hits

geographika:
Fix for #6403. This prevents a segmentation fault at: https://github.com/MapServer/MapServer/blob/6412a9287a2cb9e07e47f6570aaa8f4e701127bc/src/mapogroutput.cpp#L1043 as the results in the cache is NULL.

The fix removes/ignores any `outputformat=XXX` and so always returns the result count as XML. I'm unsure if the WFS spec defines what should happen in other format cases, but always returning XML at least prevents a server crash.

### DIFF
--- a/msautotest/wxs/expected/wfsogr20_geojson_resulttype_hits.xml
+++ b/msautotest/wxs/expected/wfsogr20_geojson_resulttype_hits.xml
@@ -1,0 +1,10 @@
+<?xml version='1.0' encoding="UTF-8" ?>
+<wfs:FeatureCollection
+   xmlns:ms="http://mapserver.gis.umn.edu/mapserver"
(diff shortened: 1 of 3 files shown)


rouault:
> I'm unsure if the WFS spec defines what should happen in other format cases, but always returning XML at least prevents a server crash.

The WFS spec is written with GML output in mind, so it doesn't say much. But GeoServer does exactly what you implement here by ignoring the outputFormat when resultType=hits. e.g https://lipas.fi/geoserver/lipas/ows?service=WFS&version=2.0.0&request=GetFeature&outputFormat=CSV&typeName=lipas:lipas_2350_tanssitila&resultType=hits

Issue #6963, opened 2023-11-03
Line OUTLINEWIDTH attribute binding issues


geographika:
Following on from investigations into #6933, two issues were encountered with use of attribute bindings and the creation of outlinewidths for creating lines. 

The [docs](https://www.mapserver.org/mapfile/style.html#mapfile-style-outlinewidth) state that an attribute can be used to set the `OUTLINEWIDTH` in a style. Note `OUTLINEWIDTH` is only used for lines and not polygons - `WIDTH` is used for polygon outline widths. 

However the checks in the code assume an integer value is set for the style. As these checks are performed before `msBindLayerToShape` calls then attribute bindings are ignored. 
https://github.com/MapServer/MapServer/blob/2b263e96d1b44fb74caf20581256c71419da802c/src/mapdraw.c#L1323

An additional condition `style->bindings[MS_STYLE_BINDING_OUTLINEWIDTH].index != -1` should be added to these, or the support for attribute binding dropped. 

A second issue with bindings is if an attribute is used for `WIDTH`. `msOutlineRenderingPrepareStyle` is used to render a line representing the outline, based on the shape `WIDTH` + `OUTLINEWIDTH`. If `WIDTH` is set to an attribute then a default value of 1 is used instead of the shape value. `msBindLayerToShape` needs to be called before `msOutlineRenderingPrepareStyle`. However an additional complication is that if `msBindLayerToShape` is called again after `msOutlineRenderingPrepareStyle` then the outline style is reset. 

This is currently the case here:

https://github.com/MapServer/MapServer/blob/2b263e96d1b44fb74caf20581256c71419da802c/src/mapdraw.c#L2535-L2537

This causes the line outline width to be incorrect when bindings are used.

geographika:
Fixed with #6964

PR #6964, opened 2023-11-03
Allow style bindings when using the feature rendering cache

geographika:
Pull request to address the issues raised in #6963. 

This adds additional checks to ensure `OUTLINEWIDTH` and road/line casing is applied if attribute binding is used for `OUTLINEWIDTH`, and that the correct bound `WIDTH` is used for rendering.

The approach is to apply the binding before calculating the underlying line style, and then avoid rebinding this to preserve the style in `msDrawVectorLayer`. 

From initial tests it no longer seems necessary to disable caching when attribute binding is used in subsequent styles. 
I added a test for https://github.com/MapServer/MapServer/issues/3976 and removed the check in https://github.com/MapServer/MapServer/commit/c63553e99ba89543fb2254e03ae6a1cecb848f57 that disables caching when
bindings are used for multiple styles. This part can easily be rolled back. 

The changes maybe addresses some of the points raised in https://github.com/MapServer/MapServer/issues/3995. I'm unsure how legend creation is meant to work when using attribute binding - if each shape can be rendered differently, what should the legend display?

Reviews and comments very much appreciated as the rendering process is complex!


sdlime:
Will definitely take some time to review!

sdlime:
Regarding legends, it's basically impossible to generate legend icons with bindings because of all the variability and lack of access to features. I think it just punts and returns a blank image. Easy enough to test, just run a mode=legend with your test mapfile.

geographika:
@sdlime - when using bindings and legends, any of the bound attributes of the style will use the style defaults defined in [initStyle](https://github.com/MapServer/MapServer/blob/990d28eed4a120d14aedc6a495cec8037b8abde6/src/mapfile.c#L2665). 

For example `width` will default to 1. Colours will be be `-1 -1 -1 255` and then ignored when generating the legend due to the `MS_VALID_COLOR(style->color)) ` check. Legends using bindings for styles will therefore be blank. 

I discussed this pull request briefly with @rouault at the Vienna codesprint - he sees no issues from a coding-side, but is less familiar with the logic around the rendering cache. 

--

As an aside while debugging I had to comment out the following line as it always returned NULL, even though an `IMAGETYPE` was set on the map, and `map->outputformat` is set.

https://github.com/MapServer/MapServer/blob/990d28eed4a120d14aedc6a495cec8037b8abde6/src/maplegend.c#L721-L723

The `MS_RENDERER_PLUGIN` macro expands to `((map->outputformat)->renderer)` and the `((map->outputformat)->renderer)` value was `105` so I don't know why it was always `False`. The msautotests in `/renderers/legend.map` pass when using a compiled mapserv so I guess this is a strange Visual Studio debugging issue?

sdlime:
Yes, this is old and somewhat complicated logic. Caching was mostly related to GD and when that support was yanked I wonder if some remnants still exist.

geographika:
> Caching was mostly related to GD and when that support was yanked. I wonder if some remnants still exist.

@sdlime the caching here only seems to be relevant to line layers to allow multiple overlain styles, and was added in https://github.com/MapServer/MapServer/commit/bcaae2c124548920bf1f3cf6f1305e6a021cdbc9

The GD caching (also in that commit) and using the variable ` gdImagePtr img_cache=NULL;` seems to have been completely removed. 

I've been testing this branch for a month with no issues, and if anything arises I'm happy to fix it and add a new test. Any further comments or is it ok to merge this in a few days time?

sdlime:
+1 on merging...

PR #6986, opened 2023-12-18
Remove trailing spaces from stroke-dasharray SLD output

geographika:
The following `STYLE`

```
        STYLE
          COLOR 255 0 0
          WIDTH 2
          PATTERN 10 10 END
        END
```

Produces SLD output with trailing whitespace:

```xml
<se:SvgParameter name="stroke-dasharray">10.00 10.00 </se:SvgParameter>
```

When working with GeoStyler this style fails to render in OpenLayers. As I understand it this whitespace in nodes are part of the node value so the fix should be on the MapServer side. 

This change ensures the trailing space is trimmed. A new test mapfile has been created for tests for more advanced SLD output from styles.


geographika:
Failing Appveyor test appears to be unrelated. New ticket created at #6987

PR #6992, opened 2023-12-20
Remove .travis.yml and migrate ASAN job to github actions

rouault:

PR #6993, opened 2023-12-21
Migration to GHA from Travis

geographika:
This pull request removes the Travis CI setup, and migrates the testing to GitHubActions as discussed in https://lists.osgeo.org/pipermail/mapserver-dev/2023-December/017060.html

@rouault - I've incorporated the changes you changes you made in #6992 - I'd be grateful if you could have a look at these changes and let me know your thoughts. 

## Changes

- Renamed scripts:

  - `/ci/travis/script.sh` to `/ci/build.sh`
  - `ci/travis/before_install.sh` to `ci/setup.sh`
  - `ci/travis/after_success.sh` removed (only a couple of relevant lines and these are added to `build.sh`

Added two new GitHub Actions:

- `build-mapscript-python.yml` - to build and test Python MapScript against different versions
- `build-mapscript-php.yml` - to build and test PHP MapScript against different versions

I've added a new environment variable `MAPSCRIPT_PYTHON_ONLY` that is used to skip running the msautotests, and without building PHP, Perl etc. to try and create a more streamlined build that can be run faster against multiple versions.
Other MapScript versions could follow this approach, but also require Python for other build steps. 

Added in two other environment variables that are used to switch various build and testing settings:

- `RUN_MSAUTOTESTS` run or not run the full msautotest suite
- `WITH_ASAN` - build and run with or without the AddressSanitizer

Other changes:

- Update CI badge in README
- lots of Python hacks for Travis can be removed

## Other Notes

1. The line in the `after_success.sh`: https://github.com/MapServer/MapServer/blob/9d888affe5cd3916337b084f4f7af73412620091/ci/travis/after_success.sh#L8C27-L8C145 I think stopped running a few years ago, and was used to upload test coverage to  https://github.com/mapServer/coverage/. This was last updated 3 years ago - perhaps the whole repo could be removed?
See log https://api.travis-ci.com/v3/job/615239849/log.txt

The coveralls library now takes care of uploading to coveralls directly e.g. https://coveralls.io/builds/64652487

2. RC is notified in a separate GitHub action - irc_notify.yml and so no need to port from Travis


.github/workflows/build-asan.yml
```diff
@@ -0,0 +1,20 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    env:
+      TEST_MAPSCRIPT_PYTHON: 'true'
+      TEST_MAPSCRIPT_PHP: 'true'
+      RUN_MSAUTOTESTS: 'true'
+      RUN_COVERALLS: 'false'
+      WITH_ASAN: 'true'
```
rouault: you need to explicitly pass them with the -e option of docker run

.github/workflows/build.yml
```diff
@@ -5,6 +5,12 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: ubuntu-20.04
+    env:
+      TEST_MAPSCRIPT_PYTHON: 'true'
```
rouault: you need to explicitly pass them with the -e option of docker run

rouault:
> 1. This was last updated 3 years ago - perhaps the whole repo could be removed?

yes sounds good

> The PHP builds require the PHP development headers. CMake always seems to pick up the latest PHP version on the test runner. This may mean shivammathur/setup-php@v2 is not an option, and there may have to be a clean-up of the test runner to get it working.

If testing several PHP versions isn't easy, I would let that aside for now, so we can unsubscribe from Travis ASAP. We have at least some PHP testing. If more is needed, it can always been added later.
Looking at https://github.com/pcolby/cmake-modules/blob/main/FindPHP.cmake, perhaps this is a matter of having the right php-config executable at the front of the PATH ? Just guessing. I don't do PHP (my suggestion of maybe shivammathur/setup-php@v2 wasn't backed up by experience)

.github/workflows/build-asan.yml
geographika: @rouault - not sure how this showed up in the pull request - this was one of several attempts that was since discarded.

.github/workflows/build.yml
geographika: Also outdated.

geographika:
The failing `build` seems to be due to running coveralls. Looking at https://coveralls.io/github/MapServer/MapServer only travis-ci seems to be pushing results. 

Errors are:

```
2023-12-21T16:52:42.1509353Z fatal: detected dubious ownership in repository at '/home/runner/work/MapServer/MapServer'
2023-12-21T16:52:42.1510015Z To add an exception for this directory, call:
2023-12-21T16:52:42.1510292Z 
2023-12-21T16:52:42.1510641Z 	git config --global --add safe.directory /home/runner/work/MapServer/MapServer
2023-12-21T16:52:43.7838556Z {'message': "Couldn't find a repository matching this job.", 'error': True}
```

I'm not sure how the https://github.com/MapServer/MapServer/actions/runs/7278796572/job/19833738707 action is passing though for #6992 as this should also have run coveralls in the case `[ "$WITH_ASAN" != "YES" ];`.

rouault:
> I'm not sure how the https://github.com/MapServer/MapServer/actions/runs/7278796572/job/19833738707 action is passing though for #6992 as this should also have run coveralls in the case `[ "$WITH_ASAN" != "YES" ];`.

looking at the log, it doesn't seam it ran coveralls for some reason.

You probably juts need to add ``git config --global --add safe.directory " ${WORK_DIR:=..}"`` in build.sh before running coveralls

geographika:
I think the `set -eu` in the `after_success.sh` script probably hid any coveralls upload errors. I've made this return true for now.

geographika:
> If testing several PHP versions isn't easy, I would let that aside for now, so we can unsubscribe from Travis ASAP. We have at least some PHP testing. If more is needed, it can always been added later. Looking at https://github.com/pcolby/cmake-modules/blob/main/FindPHP.cmake, perhaps this is a matter of having the right php-config executable at the front of the PATH ? Just guessing. I don't do PHP (my suggestion of maybe shivammathur/setup-php@v2 wasn't backed up by experience)

Agreed. We have PHP8.1 MapScript tested under the main builds. I'll continue looking into this in follow-up pull requests (along with other changes I'd like to make such as adding Python3.10). I think this pull request covers all other functionality previously in Travis. coveralls might require setting up a secret in the MapServer GitHub settings, but I don't have access to this.

rouault:
> coveralls might require setting up a secret in the MapServer GitHub settings, but I don't have access to this.

I've just added a COVERALLS_REPO_TOKEN secret in this repo with the token from https://coveralls.io/github/MapServer/MapServer/settings

geographika:
PHP testing on multiple versions (8.1 and 8.2 as with current Travis setup) now fixed. Turned out the GitHub Action runners have SWIG pre-installed, and that was taking priority with the PHP builds - lots of time looking in the wrong place..

There were also 4 failing unit tests when running on a GHA runner rather than the Docker image. These were also failing under Travis and skipped with a `IGNORE_COMPARISON_RESULT_ON_TRAVIS` variable, so I've updated this to `IGNORE_COMPARISON_RESULT_ON_CI`. It is likely there is a preinstalled library causing different results. 
There is a list of software at https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2004-Readme.md

I believe that is all Travis functionality now migrated. Coveralls should hopefully start working once merged (thanks @rouault). 
MapServer docs also migrated with https://github.com/MapServer/MapServer-documentation/pull/862 - one last action on this is for someone with MapServer org access to setup a SSH key).

sdlime:
@geographika, you're an animal. Thanks for taking this on!

rouault:
> Coveralls should hopefully start working once merged

Probably that the secret should be somehow passed with a section

env:
     COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}

and a "-e COVERALLS_REPO_TOKEN" passed to Docker

geographika:
> @geographika, you're an animal. Thanks for taking this on!

It all makes sense in hindsight, but you don't want to see the amount of failed CI builds in my MapServer fork..!

rouault:
Looking at the latest build logs I can see
```
2023-12-22T15:19:08.1186198Z Traceback (most recent call last):
2023-12-22T15:19:08.1186927Z   File "/usr/local/bin/coveralls", line 8, in <module>
2023-12-22T15:19:08.1187489Z     sys.exit(run())
2023-12-22T15:19:08.1188475Z   File "/usr/local/lib/python3.8/dist-packages/cpp_coveralls/__init__.py", line 92, in run
2023-12-22T15:19:08.1189541Z     raise ValueError("\nno coveralls.io token specified and no travis job id found\n"
2023-12-22T15:19:08.1190270Z ValueError: 
2023-12-22T15:19:08.1190728Z no coveralls.io token specified and no travis job id found
```
The first part ("no coveralls.io token specified") is expected given we are a PR. The "no travis job id" part will perhaps need later adjustments, but we can figure out that later. 
I remember to have struggle "a bit" with GDAL (I do know what you mean with try, rinse, repeat cycles when setting up CI stuff...) and ended up using the coverallsapp/github-action@v2 action, that sets everything for GHA: https://github.com/OSGeo/gdal/blob/master/.github/workflows/linux_build.yml#L308

Merging this PR, and pulling off Travis-CI subscription

PR #6996, opened 2023-12-23
coveralls.io fixes

geographika:
Follow-up to #6993 to fix coveralls.io updates, noted by @rouault in https://github.com/MapServer/MapServer/pull/6993#issuecomment-1867815481

There was an issue with quoting around the security token. I've also updated the service name which defaults to `travis-ci` in cpp-coveralls - https://github.com/eddyxu/cpp-coveralls/blob/47f638727512126bc16b1365560854a86db70e7e/cpp_coveralls/__init__.py#L79

The upload now will only happen when merging,and on the main branch rather than ignoring errors. 

Tested in my fork with a token and coveralls.io setup, and working fine: https://coveralls.io/github/geographika/mapserver

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,6 +21,7 @@ jobs:
     env:
       WITH_ASAN: ${{ matrix.WITH_ASAN }}
       COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
+      COVERALLS_SERVICE_NAME: "GitHub Actions"
 
     steps:
       - name: Checkout repository contents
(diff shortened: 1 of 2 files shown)


geographika:
When testing on my fork I was using git push. I though merging pull requests would trigger this too, but it appears I need to capture a different GHA event. Final (hopefully) pull request to follow.

rouault:
Note that coveralls also work with pull requests (the resulting builds are presented in a separate way in coveralls.io). Which is very useful to check if a new feature/bug fix has sufficient test coverages.
I'm not totally sure if this is possible with the way we use coveralls, but this is definitely handled by using the github action coverallsapp/github-action@v2 which doesn't require the coveralls secret but uses the regular github secret available in pull requests

PR #6997, opened 2023-12-23
Upload to coveralls.io on pull requests

geographika:
From https://github.com/MapServer/MapServer/pull/6996#issuecomment-1868267998

> Note that coveralls also work with pull requests (the resulting builds are presented in a separate way in coveralls.io). Which is very useful to check if a new feature/bug fix has sufficient test coverages.
> I'm not totally sure if this is possible with the way we use coveralls, but this is definitely handled by using the github action coverallsapp/github-action@v2 which doesn't require the coveralls secret but uses the regular github secret available in pull requests

This should make it easier to control when to upload - it should now be triggered on any github event, but it should be easy to modify the logic to filter certain events with the `UPLOAD_COVERALLS` variable.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,11 +29,10 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Set RUN_COVERALLS Environment Variable
-        if: env.WITH_ASAN != 'true'
-        # if: (github.event.pull_request.merged == true || github.event_name == 'push') && env.WITH_ASAN != 'true'
-        run: echo "Setting UPLOAD_COVERALLS to true"
-        env:
-          UPLOAD_COVERALLS: true
+        if: (github.event.pull_request.merged == true || github.event_name == 'push') && env.WITH_ASAN != 'true'
+        run: |
+            echo "Setting UPLOAD_COVERALLS to true"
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -22,17 +22,25 @@ jobs:
...
     steps:
       - name: Checkout repository contents
         uses: actions/checkout@v3
 
+      - name: Set RUN_COVERALLS Environment Variable
+        if: env.WITH_ASAN != 'true'
+        # if: (github.event.pull_request.merged == true || github.event_name == 'push') && env.WITH_ASAN != 'true'
+        run: echo "Setting UPLOAD_COVERALLS to true"
+        env:
+          UPLOAD_COVERALLS: true
```
rouault, line 36: This isn't propagated to the next action.
You would need to rather add the following as the "run":
```echo "UPLOAD_COVERALLS=true" >>${GITHUB_ENV}```

But I'm doubtful that will work for pull requests as the COVERALLS_REPO_TOKEN env var will be empty, since secrets are not available for pull requests.

geographika: Yes sorry, that just updates it for the step. Will set to run on pushes and merged pull requests for now and look at moving to the GitHub Action in the future.

PR #7002, opened 2024-01-10
Add CodeQL static analyzer as a CI check, and fix related warnings

rouault:


sdlime:
This is great @rouault! I'm assuming we'd still keep the Coverity scans going as well since they may identify different defects. The per commit capability here is really nice though...

rouault:
> This is great @rouault! I'm assuming we'd still keep the Coverity scans going as well since they may identify different defects

yes, I see them as complementary. Both are somehow proprietary (Coverity fully,  CodeQL has a mixed model where the checks are open source but the "engine" is not), so better have fallback plans

PR #7009, opened 2024-01-13
CI: use coverallsapp/github-action@v2

rouault:


geographika:
Thanks @rouault ! It would have taken me awhile to figure that approach out..

PR #7012, opened 2024-01-15
fixed `=*` operator translation for Oracle database

mikitex70:
This PR fixes a query issue on Oracle databases.
An expression such as `'[FIELD]' =* 'value'` was translated into a SQL expression such as `REGEXP_LIKE( FIELD, '^value$'`, which lacks the closing parenthesis.
The fix has been verified and already used on a production server.


jmckenna:
@mikitex70 makes sense to me, thanks.

@msmitherdc can you review?

msmitherdc (approved):
Yeah, not a function I've used but it does generate a syntax error without this fix, good catch.

jmckenna:
thanks again @mikitex70

Issue #7014, opened 2024-01-16
ability to use non existing properties in (FI) templates


pvgenuchten:
i want to reuse featureinfo templates in multiple grid layers, the number of bands varies between grids. The fi template should not fail if a grid does have limited number of bands

## Expected behavior and actual behavior.

Current: 
- if I use [value_1] on a single band grid the text is rendered as is.
- if i use [item name="value_1" nullformat="na"] on a single band grid i get error: `processItemTag(): Web application error. Item name (value_1) not found in layer item list.`

Expected:

3 options:
- If the "nullformat" attribute is set, do not error in the case a property does not exist, but use the value of nullformat='na'
- Introduce a new "missingfield" attribute (like nullformat), for if the field isn't found in the dataset? 
- Add a config option to neglect field-missing errors, if set, just leave missing fields blank

My favourite is 1st

## Simple test case 

```
<!-- MapServer Template -->
Band 1: [value_0]<br/>
[item name="value_1" format="Band 2: $value" nullformat=""]<br/>
```

## Operating system

ubuntu

## MapServer version and installation method

8.01 Docker-camp2camp

geographika:
This seems a useful enhancement to me. From https://mapserver.org/mapfile/template.html#general

```
nullformat = String to output if the attribute value is NULL, empty or doesn’t match the pattern (if defined). If not set and any of these conditions occur the item tag is replaced with an empty string.
```

So `nullformat` is already used for a couple of cases - both NULL values and values that don't meet a regex, so could be expanded to missing fields. The downside would be possibly breaking backwards compatibility with current templates - although as noted in this issue if a field is missing the template would be throwing an error.

sdlime:
> This seems a useful enhancement to me. From https://mapserver.org/mapfile/template.html#general
> 
> ```
> nullformat = String to output if the attribute value is NULL, empty or doesn’t match the pattern (if defined). If not set and any of these conditions occur the item tag is replaced with an empty string.
> ```
> 
> So `nullformat` is already used for a couple of cases - both NULL values and values that don't meet a regex, so could be expanded to missing fields. The downside would be possibly breaking backwards compatibility with current templates - although as noted in this issue if a field is missing the template would be throwing an error.

Perhaps setting an item-level attribute to ignore missing a item - potentially paired with the nullformat - would be a way to avoid backwards compatibility issues.

geographika:
[...]

PR #7017, opened 2024-01-19
Add a ignoremissing option to template items

geographika:
As discussed in #7014 this adds a new `ignoremissing` property to template items. Example usage:

```
[item name="value_99" nullformat="missing" ignoremissing="true"]
```

If the field `value_99` is missing from the features then the `nullformat` value will be displayed (in this case "missing").
This allows the same template to be used for different datasets which may have slightly different fields, without throwing an error. 

Docs to be updated at https://mapserver.org/mapfile/template.html#format if merged.


sdlime:
Looks good to me!

Issue #6113, opened 2020-07-10
Issues with the [feature] substitution tag and includes


kungfujohnjon:
The feature substitution tag [(from RFC 36)](https://mapserver.org/output/template_output.html) isn't parsed when it's inside an include.
e.g.
main.tmpl
```
// mapserver template
[
[resultset layer=xlayer][include src="itemlist.tmpl"][/resultset]
]
```

itemlist.tmpl
```
// mapserver template
[feature trimlast=","]{"something":"[item name="whatever"]"},[/feature]
```
For this template, main.tmpl will include itemlist.tmpl, but doesn't parse any of the substitution tags inside it. The output will look like
```
[
[feature trimlast=","]{"something":"[item name="whatever"]"},[/feature]
]
```

Also, the trimlast option for features doesn't work with includes.
e.g.
main.tmpl
```
// mapserver template
[
[resultset layer=xlayer]
[feature trimlast=","][include src="itemlist.tmpl"][/feature]
[/resultset]
]
```
itemlist.tmpl
```
//mapserver template
{"something":"[item name="whatever"]"},
```

In this case, the item tag substitution inside the include does work, but the last comma never gets trimmed, e.g.
```
[
{"something":"data"},
{"something":"data"},
...
{"something":"data"},
]
```

By contrast, if main.tmpl is just:
```
// mapserver template
[
[resultset layer=xlayer]
[feature trimlast=","]{"something":"[item name="whatever"]"},[/feature]
[/resultset]
]
```
then the output is trimmed as expected.

I'm testing this on MapServer 7.6.0 built from source on RHEL 7.

Just for completeness, this would be the bones of a map file that would call this template.
Map File:
```
MAP
...
OUTPUTFORMAT
  NAME "json"
  DRIVER "TEMPLATE"
  MIMETYPE "application/json"
  FORMATOPTION "FILE=/path/to/main.tmpl"
END
...
LAYER
  NAME "xlayer"
  TYPE LINE
  STATUS ON
  INCLUDE "DatabaseConnection.map"
  DATA "gis_geometry FROM (SELECT gis_geometry, uid, whatever FROM x  ) as foo using unique uid using SRID=4326"
  METADATA
    ...
    "wms_include_items"  "uid,whatever"
    "wms_feature_info_mime_type" "application/json"
    ...
  END
  TEMPLATE "json"
  ...
END
END
```

jmckenna:
[...]

PR #7020, opened 2024-02-01
Allow includes in feature templates

geographika:
Fix for #6113. 

The issue was that `[feature]` tags are only processed within a `[resultset]` tag. If included templates don't contain the `[resultset]` tag the template content is left as a raw string. 
When processing `[resultset]` the code searches through the file for the closing `[resultset]` tag and gets the content inbetween, without processing `[include]` tags.
This pull request adds a call to `processIncludeTag` within `processResultSetTag` which resolves the issue.

In addition several typos were corrected (as there was quite a bit of time debugging this file to find the issue). Sorry this makes the actual code changes harder to review. 

In the end the only required code changes were adding the following, and moving `processIncludeTag` higher in the file to ensure it was checked:

```c
    /* process any includes within the resultset tag */
    if (processIncludeTag(mapserv, &tag, stream, 0) != MS_SUCCESS) {
      msFreeHashTable(tagArgs);
      return (MS_FAILURE);
    }
```

There was also a couple of (existing) memory leaks in the case where templates were missing that have been fixed. These were highlighted by the Address Sanitizer CI when new tests were added.
This pull request includes test files to check the issues reported in #6113 - including a JSON example of stripping the last comma when using includes. 

One minor (again I believe existing) issue, is that the templates write to a stream, so if errors are encountered after output has been sent then any template text has already been written.
This leads to slightly messed up error responses - such as in the test result template_test005.html

```
This template tests a missing feature template
Content-Type: text/html

<HTML>
<BODY BGCOLOR="#FFFFFF">
processIncludeTag(): Unable to access file. missing.tmpl
</BODY></HTML>
```

As this error only occurs when templates are missing it will probably only be seen by the developers of a system rather than users.


rouault:
@geographika is this good to merge?

geographika:
> @geographika is this good to merge?

Yes, all good from my side.

PR #7024, opened 2024-02-18
Issue 6907

sdlime:
Fixes a regression with the use of the WEB EMPTY attribute. Also adds capability to set a global setting via MS_EMPTY in the config file.


rouault:
@sdlime Do you have issues to install pre-commit ? Cf https://mapserver.org/development/dev_practices.html#commit-hooks and then `pre-commit run --all-files` to fix current issues

sdlime:
@rouault, do you know the best way to remove a few source files from a pull request? I inadvertently included the mapparser/maplexer generated files. It looks like I can just remove them from the pull request via the GUI.

sdlime:
> @sdlime Do you have issues to install pre-commit ? Cf https://mapserver.org/development/dev_practices.html#commit-hooks and then `pre-commit run --all-files` to fix current issues

Just getting caught up on this stuff. Thanks for the tip.

rouault:
> do you know the best way to remove a few source files from a pull request? I inadvertently included the mapparser/maplexer generated files.

I would do "git reset master", which will not modify your local modifications but just uncommit them. Then you can just "git add ...." and "git commit"

sdlime:
I'm gonna redo this properly... never mind.

PR #7028, opened 2024-02-23
Issue 6956

mapcentia:
#6956


mapcentia:
The second commit should fix the linting check

geographika:
Actually it appears as though the expected output may itself be wrong. The output was modified as part of #6430 - see https://github.com/MapServer/MapServer/commit/3a100ab0ac304700cb16ff12f8a892131b31e89e

The original file looked like this https://github.com/MapServer/MapServer/blob/e2dc1451d22a73ca10f4f930436f2c73e7b496e5/msautotest/sld/expected/map_linemark.png

map_linemark

So this update may actually be fixing a regression.

mapcentia:
I get this now with the linemark.map

image

mapcentia:
Notice: This issue is related to librsvg >= 2.46

geographika:
> I get this now with the linemark.map
> 
> image

Yes this looks to be the original, and correct, output. I've updated the expected results in the branch so we'll see if the tests now pass.

PR #7034, opened 2024-03-03
RFC 138 Implementation - Reference SLD files in Mapfiles

geographika:
Pull request for https://mapserver.org/development/rfc/ms-rfc-138.html - allowing SLD files to be referenced in a Mapfile:

```
LAYER
  STYLEITEM "sld://mysldfile.xml"
  ...
END
```
 
The logic of applying the LAYER created from the SLD to the MAP LAYER was moved to a new function ``msApplySldLayerToMapLayer``. This allows the code to be reused to set a default style from the SLD file
if there are no matching names.

Following discussions in https://lists.osgeo.org/pipermail/mapserver-dev/2024-February/017105.html, the following approaches have been implemented:

- For SLD files with multiple NamedLayers, if the Name matches the layer names then this will be used. 
- If there are no matches then by default the first named layer will be used. A new ``SLD_USE_FIRST_NAMEDLAYER`` metadata item is created to apply this logic by default. It can be set in the Mapfile to only apply NamedLayer if there is a match in the SLD file using ``METADATA "SLD_USE_FIRST_NAMEDLAYER" "false"`` (or any other value than true). 
Suggestions on any better name for this metadata key appreciated. 
- I have also added in support for SLD to be used in attributes, similar to the OGR styling support added in https://mapserver.org/development/rfc/ms-rfc-61.html. This had a few issues to resolve in ``msApplySldLayerToMapLayer`` as ``msSLDApplySLD`` had up until now only worked on layers that were closed. As the SLD is applied on a per-feature basis while looping through features logic was added to avoid re-opening or closing the layer. As suggested by @szekerest ``msLayerGetFeatureStyle`` was updated to call ``msSLDApplySLD``. Applying SLD is done at a layer level, so there will be cases such as filters that won't work correctly here - but the basics of styling with SLD will work. Many MapServer services such as GetStyles will already fail if attempting to use on layers using per-feature styling, so the SLD support will work in the same way. 

## Implementation Details

I went on a few dead-ends in terms of where to apply the SLD to the layer. Initially I had this in the drawing functionality. However it became apparent the file would also need to be applied in other cases, such as in ``GetStyle`` or legend requests. This created much more complicated logic throughout the codebase. A simple approach was to apply the SLD styling when the Mapfile was parsed by adding logic to ``msApplyDefaultSubstitutions``. This seems to fit as the SLD STYLEITEM is in effect replacing a CLASS with classes generated from the SLD. Cases such as overriding the styling with WMS SLD_BODY requests were then handled by default. 

Initially I thoughy maybe returning the SLD file directly for WMS ``GetStyle`` requests could be an option. However users can request the SLD in different versions, so the SLD would need to be processed anyway - it was easier to convert to CLASS/STYLEs and then generate the SLD back from this as needed. There maybe some differences in the input/output SLD as part of this roundtrip. 

Note the Mapfile will be modified by applying the SLD file (as with other items such as WFS filters, SLD form URLs etc.) so Mapfiles cannot be reused in subsequent requests (see similar discussions at #4018). 

If an SLD file is missing the logs will have the error ``msSLDApplyFromFile(): Unable to access file. Invalid SLD filename: "sld/data/missing.sld".``
If there is a CLASS in the LAYER then this will be used as a fallback for styling, otherwise the layer will produce an empty output. 

Along with a test suite for the new functionality a couple of missing tests have been added for existing functionality to ensure it is not broken by changes in this pull request:

+ A test for ``msautotest/misc/ogrpen.map`` added to check the output of feature styling
+ ``wms_get_map_mark_symbol_sld_legend`` to test GetLegendGraphic request when SLD_BODY applied

Reviews and comments appreciated! Docs will follow once this pull request is finalised.


src/mapdraw.c
```diff
@@ -1106,7 +1106,8 @@ int msDrawVectorLayer(mapObj *map, layerObj *layer, imageObj *image) {
 
   /* build item list. STYLEITEM javascript needs the shape attributes */
   if (layer->styleitem &&
-      (strncasecmp(layer->styleitem, "javascript://", 13) == 0)) {
+      ((strncasecmp(layer->styleitem, "javascript://", 13) == 0) ||
```
rouault, line 1109: for your information, if you include "cpl_port.h" from GDAL, you can juste use ``STARTS_WITH_CI(layer->styleitem, "javascript://") || lSTARTS_WITH_CI(layer->styleitem, "sld://")`` which saves you from a potential wrong string length

src/mapfile.c
```diff
@@ -7296,6 +7298,18 @@ void msApplyDefaultSubstitutions(mapObj *map) {
         layer, &(layer->validation)); /* ...then layer settings... */
     applyLayerDefaultSubstitutions(
         layer, &(map->web.validation)); /* ...and finally web settings */
+
+    // apply any styles referenced in STYLEITEM to the layer
```
rouault, line 7302: I'm a bit unconfortable seeing that in msApplyDefaultSubstitutions(). Especially since this function is not called if map->web.validation is set to "immutable". Perhaps having a dedicated function and calling it from towards the end of msLoadMap() would make sense?

src/mapfile.c
```diff
@@ -7296,6 +7298,18 @@ void msApplyDefaultSubstitutions(mapObj *map) {
...
+    if (layer->styleitem && strncasecmp(layer->styleitem, "sld://", 6) == 0) {
+      char *filename = layer->styleitem + 6;
```
rouault, line 7304:
```suggestion
      const char *filename = layer->styleitem + 6;
```

src/maplayer.c
```diff
@@ -1411,6 +1418,8 @@ int msLayerGetFeatureStyle(mapObj *map, layerObj *layer, classObj *c,
              strncasecmp(stylestring, "symbol", 6) == 0 ||
              strncasecmp(stylestring, "label", 5) == 0) {
     msOGRUpdateStyleFromString(map, layer, c, stylestring);
+  } else if (strcasestr(stylestring, "StyledLayerDescriptor>") != NULL) {
```
rouault, line 1421: what about patterns like ``<StyledLayerDescriptor version='1.1.0'>`` ?

src/mapogcsld.cpp
```diff
@@ -37,11 +37,12 @@ extern "C" {
...
 }
 
-#if defined(USE_CURL)
 static inline void IGUR_sizet(size_t ignored) {
   (void)ignored;
 } /* Ignore GCC Unused Result */
-#endif
+static inline void IGUR_voidp(void *ignored) {
+  (void)ignored;
+} /* Ignore GCC Unused Result */
```
rouault, line 45: this seems to be unused

src/mapogcsld.cpp
```diff
@@ -147,6 +148,58 @@ int msSLDApplySLDURL(mapObj *map, const char *szURL, int iLayer,
 #endif
 }
 
+/************************************************************************/
+/*                             msSLDApplyFromFile                       */
+/*                                                                      */
+/*      Apply SLD from a file on disk to the layerObj                   */
+/************************************************************************/
+int msSLDApplyFromFile(mapObj *map, layerObj *layer, char *filename) {
```
rouault, line 156:
```suggestion
int msSLDApplyFromFile(mapObj *map, layerObj *layer, const char *filename) {
```

src/mapogcsld.cpp
```diff
@@ -147,6 +148,58 @@ int msSLDApplySLDURL(mapObj *map, const char *szURL, int iLayer,
...
+
+  FILE *fp = NULL;
+
+  int nStatus = MS_FAILURE;
+  const char *realpath;
```
rouault, line 161:
```suggestion
```

src/mapogcsld.cpp
```diff
@@ -147,6 +148,58 @@ int msSLDApplySLDURL(mapObj *map, const char *szURL, int iLayer,
...
+
+  /* open SLD file */
+  char szPath[MS_MAXPATHLEN];
+  char *pszSLDbuf = NULL;
+  realpath = msBuildPath(szPath, map->mappath, filename);
```
rouault, line 166:
```suggestion
  const char *realpath = msBuildPath(szPath, map->mappath, filename);
```

src/mapogcsld.cpp
```diff
@@ -147,6 +148,58 @@ int msSLDApplySLDURL(mapObj *map, const char *szURL, int iLayer,
...
+
+  if ((fp = fopen(realpath, "rb")) != NULL) {
+    int nBufsize = 0;
+    fseek(fp, 0, SEEK_END);
+    nBufsize = ftell(fp);
+    if (nBufsize > 0) {
+      rewind(fp);
+      pszSLDbuf = (char *)malloc((nBufsize + 1) * sizeof(char));
```
rouault, line 174: it would prudent to check pszSLDbuf for null

src/mapogcsld.cpp
```diff
@@ -196,6 +249,206 @@ static void msSLDApplySLD_DuplicateLayers(mapObj *map, int nSLDLayers,
 }
 #endif
 
+int msApplySldLayerToMapLayer(layerObj *sldLayer, layerObj *lp) {
+
+  if (sldLayer->numclasses > 0) {
+    int iClass = 0;
+    int k;
```
rouault, line 256:
```suggestion
```

src/mapogcsld.cpp
```diff
@@ -196,6 +249,206 @@ static void msSLDApplySLD_DuplicateLayers(mapObj *map, int nSLDLayers,
...
+    int bSLDHasNamedClass = MS_FALSE;
```
rouault, line 257:
```suggestion
    bool bSLDHasNamedClass = false;
```

src/mapogcsld.cpp
```diff
@@ -196,6 +249,206 @@ static void msSLDApplySLD_DuplicateLayers(mapObj *map, int nSLDLayers,
...
+
+    lp->type = sldLayer->type;
+    lp->rendermode = MS_ALL_MATCHING_CLASSES;
+
+    for (k = 0; k < sldLayer->numclasses; k++) {
```
rouault, line 262:
```suggestion
    for (int k = 0; k < sldLayer->numclasses; k++) {
```

src/mapogcsld.cpp
```diff
@@ -196,6 +249,206 @@ static void msSLDApplySLD_DuplicateLayers(mapObj *map, int nSLDLayers,
...
+      if (sldLayer->_class[k]->group) {
+        bSLDHasNamedClass = MS_TRUE;
```
rouault, line 264:
```suggestion
        bSLDHasNamedClass = true;
```

src/mapogcsld.cpp
```diff
@@ -196,6 +249,206 @@ static void msSLDApplySLD_DuplicateLayers(mapObj *map, int nSLDLayers,
...
+        break;
+      }
+    }
+
+    for (k = 0; k < lp->numclasses; k++) {
```
rouault, line 269:
```suggestion
    for (int k = 0; k < lp->numclasses; k++) {
```

src/mapogcsld.cpp
```diff
@@ -196,6 +249,206 @@ static void msSLDApplySLD_DuplicateLayers(mapObj *map, int nSLDLayers,
...
+      msFree(lp->classgroup);
+      lp->classgroup = msStrdup(sldLayer->classgroup);
+    } else {
+      /*unset the classgroup on the layer if it was set. This allows the
+      layer to render with all the classes defined in the SLD*/
+      msFree(lp->classgroup);
+      lp->classgroup = NULL;
+    }
+
+    for (k = 0; k < sldLayer->numclasses; k++) {
```
rouault, line 292:
```suggestion
    for (int k = 0; k < sldLayer->numclasses; k++) {
```

src/mapogcsld.cpp
```diff
@@ -196,6 +249,206 @@ static void msSLDApplySLD_DuplicateLayers(mapObj *map, int nSLDLayers,
...
+
+      initClass(lp->_class[iClass]);
+      msCopyClass(lp->_class[iClass], sldLayer->_class[k], NULL);
+      lp->_class[iClass]->layer = lp;
+      lp->numclasses++;
+
+      /*aliases may have been used as part of the sld text symbolizer
+        for label element. Try to process it if that is the case #3114*/
+
+      int layerWasOpened = msLayerIsOpen(lp);
```
rouault, line 304:
```suggestion
      const int layerWasOpened = msLayerIsOpen(lp);
```

src/mapogcsld.cpp
```diff
@@ -196,6 +249,206 @@ static void msSLDApplySLD_DuplicateLayers(mapObj *map, int nSLDLayers,
...
+
+      if (layerWasOpened || msLayerOpen(lp) == MS_SUCCESS) {
+        if (layerWasOpened || msLayerGetItems(lp) == MS_SUCCESS) {
```
rouault, line 307: will require re-indenting the block below
```suggestion
      if (layerWasOpened || msLayerOpen(lp) == MS_SUCCESS || msLayerGetItems(lp) == MS_SUCCESS) {
```

src/mapogcsld.cpp
```diff
@@ -196,6 +249,206 @@ static void msSLDApplySLD_DuplicateLayers(mapObj *map, int nSLDLayers,
...
+              }
+              msFree(pszTmp1);
+            }
+          }
+          if (!layerWasOpened) { // don't close the layer if already
+                                 // open
+            msLayerClose(lp);
+          }
+        }
+      }
```
rouault, line 338:
```suggestion
      }
```

src/mapogcsld.cpp
```diff
@@ -196,6 +249,206 @@ static void msSLDApplySLD_DuplicateLayers(mapObj *map, int nSLDLayers,
...
+
+      if (layerWasOpened || msLayerOpen(lp) == MS_SUCCESS) {
+        if (layerWasOpened || msLayerGetItems(lp) == MS_SUCCESS) {
+
+          if (lp->_class[iClass]->text.string) {
+            int z;
+            for (z = 0; z < lp->numitems; z++) {
+              const char *pszFullName;
+              char szTmp[512];
+              char *pszTmp1;
```
rouault, line 314:
```suggestion
            for (int z = 0; z < lp->numitems; z++) {
 ``

File: src/mapogcsld.cpp [hunk 20]
```diff
@@ -196,6 +249,206 @@ static void msSLDApplySLD_DuplicateLayers(mapObj *map, int nSLDLayers,
...
+              if (!lp->items[z] || strlen(lp->items[z]) == 0)
+                continue;
+              snprintf(szTmp, sizeof(szTmp), "%s_alias", lp->items[z]);
+              pszFullName = msOWSLookupMetadata(&(lp->metadata), "G", szTmp);
+              pszTmp1 = msStrdup(lp->_class[iClass]->text.string);
```
rouault, line 319:
```suggestion
              const char* pszFullName = msOWSLookupMetadata(&(lp->metadata), "G", std::string(lp->items[z]).append("_alias").c_str());
              pszTmp1 = msStrdup(lp->_class[iClass]->text.string);
```

src/mapogcsld.cpp
```diff
@@ -196,6 +249,206 @@ static void msSLDApplySLD_DuplicateLayers(mapObj *map, int nSLDLayers,
...
+              if (pszFullName != NULL &&
+                  (strstr(pszTmp1, pszFullName) != NULL)) {
+                pszTmp1 =
+                    msReplaceSubstring(pszTmp1, pszFullName, lp->items[z]);
+                char *pszTmp2 =
+                    (char *)malloc(sizeof(char) * (strlen(pszTmp1) + 3));
+                sprintf(pszTmp2, "(%s)", pszTmp1);
+                msLoadExpressionString(&(lp->_class[iClass]->text), pszTmp2);
+                msFree(pszTmp2);
```
rouault, line 328:
```suggestion
                msLoadExpressionString(&(lp->_class[iClass]->text), std::string("(").append(pszTmp1).append(")").c_str());
```

src/mapogcsld.cpp
```diff
@@ -196,6 +249,206 @@ static void msSLDApplySLD_DuplicateLayers(mapObj *map, int nSLDLayers,
...
+            else
+              snprintf(szTmp, sizeof(szTmp), "%s", " OR ");
+
+            pszBuffer = msStringConcatenate(pszBuffer, szTmp);
+            pszBuffer = msStringConcatenate(pszBuffer,
+                                            lp->_class[k]->expression.string);
+          }
+
+          snprintf(szTmp, sizeof(szTmp), "%s", "))");
+          pszBuffer = msStringConcatenate(pszBuffer, szTmp);
```
rouault, line 436: could be simplified with std::string concatenation

src/mapogcsld.cpp
```diff
@@ -241,8 +494,11 @@ int msSLDApplySLD(mapObj *map, const char *psSLDXML, int iLayer,
       /* compare layer name to wms_name as well */
       pszWMSLayerName = msOWSLookupMetadata(&(lp->metadata), "MO", "name");
 
+      int bSldApplied = 0;
+      layerObj *sldLayer = NULL;
```
rouault, line 498:
```suggestion
      bool bSldApplied = false;
```

src/mapogcsld.cpp
```diff
@@ -241,8 +494,11 @@ int msSLDApplySLD(mapObj *map, const char *psSLDXML, int iLayer,
...
+
       for (j = 0; j < nSLDLayers; j++) {
-        layerObj *sldLayer = &pasSLDLayers[j];
+        sldLayer = &pasSLDLayers[j];
```
rouault, line 501:
```suggestion
        layerObj *sldLayer = &pasSLDLayers[j];
```

src/mapogcsld.cpp
```diff
@@ -269,202 +525,30 @@ int msSLDApplySLD(mapObj *map, const char *psSLDXML, int iLayer,
...
-              msFree(lp->classgroup);
-              lp->classgroup = NULL;
-            }
-
-            for (k = 0; k < sldLayer->numclasses; k++) {
-              if (msGrowLayerClasses(lp) == NULL)
+          if (msApplySldLayerToMapLayer(sldLayer, lp) == MS_FAILURE) {
+            return MS_FAILURE;
+          };
+          bSldApplied = 1;
```
rouault, line 531:
```suggestion
          bSldApplied = true;
```

src/mapogcsld.cpp
```diff
@@ -269,202 +525,30 @@ int msSLDApplySLD(mapObj *map, const char *psSLDXML, int iLayer,
...
+          break;
+        }
+      }
+      if (bUseSpecificLayer) {
+        if (bSldApplied == 0) {
```
rouault, line 536:
```suggestion
        if (!bSldApplied) {
```

src/mapogcsld.cpp
```diff
@@ -269,202 +525,30 @@ int msSLDApplySLD(mapObj *map, const char *psSLDXML, int iLayer,
...
+          // there was no name match between the map layer and the SLD named
+          // layer - check if we should apply the first SLD layer anyway
+          const char *pszSLDUseFirstNamedLayer =
+              msLookupHashTable(&(lp->metadata), "SLD_USE_FIRST_NAMEDLAYER");
+          if (pszSLDUseFirstNamedLayer) {
+            if (strcasecmp(pszSLDUseFirstNamedLayer, "true") == 0) {
+              sldLayer = &pasSLDLayers[0];
```
rouault, line 543:
```suggestion
              layerObj *sldLayer = &pasSLDLayers[0];
```

src/mapogcsld.h
```diff
@@ -40,6 +40,8 @@ MS_DLL_EXPORT int msSLDApplySLDURL(mapObj *map, const char *szURL, int iLayer,
 MS_DLL_EXPORT int msSLDApplySLD(mapObj *map, const char *psSLDXML, int iLayer,
                                 const char *pszStyleLayerName,
                                 char **ppszLayerNames);
+MS_DLL_EXPORT int msSLDApplyFromFile(mapObj *map, layerObj *layer,
+                                     char *filename);
```
rouault, line 44: Internal function only. Doesn't need to be exported:

```suggestion
int msSLDApplyFromFile(mapObj *map, layerObj *layer, const char *filename);
```

src/mapogcsld.cpp
geographika: Ah the luxury of boolean types in C++!

src/mapogcsld.cpp
geographika: Is it ok to re-declare the same variable name in a nested block (it is also declared on line 498), or should a different variable name be used?

rouault: > Is it ok to re-declare the same variable name in a nested block (it is also declared on line 498)

sldLayer = &pasSLDLayers[j]; at line 501 seems to be in a unrelated scope of the one at line 543. Or perhaps I don't read correctly the scopes in the github review UI?

src/maplayer.c
geographika: This is to try and find the closing tag of the SLD e.g. `</StyledLayerDescriptor>` or if namespaces are used `</sld:StyledLayerDescriptor>`

rouault: ah good point. A comment in the code would help clarify that for future readers

src/mapogcsld.cpp
geographika: Yes different scope (assuming C doesn't use hoisting like JS). I've renamed the variable anyway for clarity.

src/mapogcsld.cpp
geographika: Removed

src/mapogcsld.cpp
geographika: Updated

src/mapdraw.c
geographika: I've updated the changes in this pull request to use `STARTS_WITH_CI` (note to self - CI is for Case Insensitive). There are plenty of places in the MapServer codebase that could also switch to this approach.

src/maplayer.c
geographika: Added a comment

src/mapogcsld.cpp
geographika: Added a check to cover both NULL and 0 (which was hinted as a possible error) using the following code. If this check makes sense I can add it to the original function I copied it from. 

```c
      if (!pszSLDbuf) {
        msSetError(MS_MEMERR, "Failed to read SLD file.",
                   "msSLDApplyFromFile()");
        nStatus = MS_FAILURE;
      
```

src/mapogcsld.cpp
geographika: Updated

src/mapogcsld.cpp
geographika: This code was from the original SLD logic, but moved to a new function. Will review.

src/mapogcsld.cpp
geographika: Updated, but `msLoadExpressionString` expects `char *` and not `const char *`. Changing this would impact lots of places in the code so probably outside the scope of this pull request. I used the following construct to cast the c_str, but not sure if this is correct:

```
              char *pszTmp2 = const_cast<char *>(
                  std::string("(").append(pszTmp1).append(")").c_str());
```

src/mapogcsld.cpp
rouault: > to cover both NULL and 0 

not sure to understand what 0 means here?

actually a very edgy case would be nBufSize == INT_MAX, in which case nBufSize + 1 would overflow...

src/mapogcsld.cpp
rouault: > Updated, but `msLoadExpressionString` expects `char *` and not `const char *`

should be fixed per https://github.com/MapServer/MapServer/pull/7035 . You can rebase on top of latest master once this will be merged

src/mapogcsld.cpp
geographika: Checking just for NULL gets rid of the warning in VS
image

This code is used in a couple of other places in MapServer.

src/mapogcsld.cpp
rouault: yes I realized that. no obligation to fix that. up to you

src/mapfile.c
geographika: Moved to its own function `applyStyleItemToLayer` and added calls to `msLoadMapFromString` and `msLoadMap`.

src/mapdraw.c
```diff
@@ -1106,7 +1107,8 @@ int msDrawVectorLayer(mapObj *map, layerObj *layer, imageObj *image) {
 
   /* build item list. STYLEITEM javascript needs the shape attributes */
   if (layer->styleitem &&
-      (strncasecmp(layer->styleitem, "javascript://", 13) == 0)) {
+      ((STARTS_WITH_CI(layer->styleitem, "javascript://") == 0) ||
+       (STARTS_WITH_CI(layer->styleitem, "sld://") == 0))) {
```
rouault, line 1111: STARTS_WITH_CI() returns a boolean. You shouldn't compare to 0

```suggestion
      (STARTS_WITH_CI(layer->styleitem, "javascript://")  ||
       STARTS_WITH_CI(layer->styleitem, "sld://")) {
```

src/maplayer.c
```diff
@@ -894,7 +895,8 @@ static void buildLayerItemList(layerObj *layer) {
     layer->filteritemindex =
         string2list(layer->items, &(layer->numitems), layer->filteritem);
   if (layer->styleitem && (strcasecmp(layer->styleitem, "AUTO") != 0) &&
-      (strncasecmp(layer->styleitem, "javascript://", 13) != 0))
+      STARTS_WITH_CI(layer->styleitem, "javascript://") &&
+      STARTS_WITH_CI(layer->styleitem, "sld://"))
```
rouault, line 899:
```suggestion
      !STARTS_WITH_CI(layer->styleitem, "javascript://") &&
      !STARTS_WITH_CI(layer->styleitem, "sld://"))
```

src/maplayer.c
```diff
@@ -1067,7 +1069,8 @@ int msLayerWhichItems(layerObj *layer, int get_all, const char *metadata) {
   if (layer->filteritem)
     nt++;
   if (layer->styleitem && (strcasecmp(layer->styleitem, "AUTO") != 0) &&
-      (strncasecmp(layer->styleitem, "javascript://", 13) != 0))
+      STARTS_WITH_CI(layer->styleitem, "javascript://") &&
+      STARTS_WITH_CI(layer->styleitem, "sld://"))
```
rouault, line 1073:
```suggestion
      !STARTS_WITH_CI(layer->styleitem, "javascript://") &&
      !STARTS_WITH_CI(layer->styleitem, "sld://"))
```

geographika: Missed these comments, but eventually came to the same conclusion! Thanks for all the assistance @rouault. I think I messed up the alias logic, but will rebase on to #7035 and review tomorrow.

rouault (approved):

geographika:
Final commits include:

- add a test to apply SLD generated in QGIS to a raster file. 
See https://docs.geoserver.org/main/en/user/styling/qgis/index.html for exporting SLD from QGIS. However by default QGIS export `<UserLayer>`. This needs to be changed to `<NamedLayer>` to work with MapServer.

- The style attribute SLD test was updated to use the new METADATA key `"SLD_USE_FIRST_NAMEDLAYER" "true"` and one of the SLD strings had a NamedLayer Name set to a different value than its layer. The METADATA key
allows the SLD to be applied to the feature even though the names don't match. 

- I also added the NULL check when opening the SLD file, suggested in the code review, to the original function I copied. 

All good to go from my side (apart from checking why Appveyor isn't starting) and ready for the 8.2 release. 

Thanks very much for the review @rouault!

geographika:
Looks like the new raster test found a memory leak in an existing function `msSLDParseRasterSymbolizer`: https://github.com/MapServer/MapServer/actions/runs/8164584864/job/22320124941?pr=7034#step:3:9222

Will investigate further.

rouault:
@geographika I've pushed an extra commit in your branch for a better fix for the memleak. Your initial change could cause a potential use-after-free when classitem was not null and different from "[pixel]".

pvgenuchten:
Nice work Seth and Even, looking forward to test

PR #7039, opened 2024-03-07
flatgeobuf: fix off-by-one error which skipped last feature with index search

PPazderski:
This bug I (mostly) found by accident while adding an msautotest for another bug. The stop condition to iterate the flatgeobuf index search results is off by one
and therefore the last found feature is missing.

It can even be seen in the [test image](https://github.com/MapServer/MapServer/blob/8f4586637176f883c521c49abd28ef7939b5c8b3/msautotest/misc/expected/flatgeobuf-continent.png) that [Congo](https://en.wikipedia.org/wiki/Republic_of_the_Congo) is missing. (but I also looked at this image multiple times before I even noticed)

The XML producing tests for this mapfile work because they seem to skip the flatgeobuf index.


rouault:
Looks good! Thanks!

PR #7040, opened 2024-03-07
flatgeobuf: fix out of bounds read after index search without results

PPazderski:
This bug is a bit spread out, involves some randomness and the outcome depends on the environment so please bear with my rather long description.

If an index search is performed in `msFlatGeobufLayerWhichShapes` and the search has an empty result then
`flatgeobuf_index_search` will set `search_result_len` to zero but also allocate a buffer for zero results.

https://github.com/MapServer/MapServer/blob/8f4586637176f883c521c49abd28ef7939b5c8b3/src/flatgeobuf/flatgeobuf_c.cpp#L358-L360

From what I read such zero size allocations can return a null pointer or a pointer to a random location. Both cases are problematic.

`msFlatGeobufLayerNextShape` interprets a null pointer as `search_result` as index was skipped and will try to read all features but the file handle is at the wrong offset if the index was actually used.

If `search_result` was given a non-null pointer `msFlatGeobufLayerNextShape` will check if there are more results found with the following condition:

https://github.com/MapServer/MapServer/blob/8f4586637176f883c521c49abd28ef7939b5c8b3/src/mapflatgeobuf.c#L235-L238

With an empty search the result length is zero and because it is unsigned it will underflow and the condition evaluate to false. As consequence it will read the first search result where none is.

The outcome is that item.offset is a random value and either the following seek fails (which results in a maperror) or if it succeeds the following read most likely fails with an EOF (which then produces the correct output by accident and no error) or if the read succeeds the following buffer allocation most likely fails as "too huge" or if that also succeeds decoding the random file content as feature will likely fail.

How to reproduce
----

It is actually not so easy to reproduce because it depends on so many factors and in many cases it appears to work correctly.
I had this issue when I made many requests to a mapserver running with fcgid.
The best I found to see the problem reliable is to trace the system calls with the following command:

`strace --trace lseek map2img -m msautotest/misc/flatgeobuf.map -l africa-continent -e -7 -17 -4 -14 -o /dev/null`

(the extends lies in the ocean therefore the expected output would be an empty image)

In the strace output you should see an SEEK_SET with an offset which makes (most likely) no sense. Either the offset is ridiculous huge or at least larger than the `africa.fgb` file.

Maybe you will also see one of the possible error messages mentioned above or it will silently succeed. The output is kinda unpredictable. From what I read the outcome also depends on the used filesystem because fseek might fail if the offset is larger than the maximum possible file size.

Regards the integer underflow
----

As mentioned before one condition to trigger this error is the underflow which happens in the condition if `search_result_len` is zero. With my change it is not necessary to change it because this point is not reached anymore.

Actually I intentionally do not change it in this pull request because I later figured out it is an off-by-one error I propose to fix in #7039.

Notes on my fix
----

At first I attempted to set `search_result = NULL` if nothing was found but then it is indistinguishable from an index skip situation.
I could have given `search_result_len` a special value like "int max" to indicate an index skip but that felt unnecessary dirty and changing the context object also felt excessive.

In my opinion stopping early from `msFlatGeobufLayerWhichShapes` if the search has no result is a good solution. Especially since it already returns `MS_DONE` if the request area is outside map extends so it is expected that the return value is handled correctly.

I also added the above described map2img invocation to trigger the error as a test case but it only has limited value because even without the fix it is unlikely that it will recognize the error.


msautotest/misc/flatgeobuf.map
```diff
@@ -5,6 +5,8 @@
 #
 #RUN_PARMS: flatgeobuf-continent.png [MAP2IMG] -m [MAPFILE] -i png -o [RESULT] -l africa-continent
 #RUN_PARMS: flatgeobuf-classes.png [MAP2IMG] -m [MAPFILE] -i png -o [RESULT] -l africa-classes
+# Given extend is fully in the Atlantic Ocean therefore no feature is selected and the result empty
```
rouault, line 8:
```suggestion
# Given extent is fully in the Atlantic Ocean therefore no feature is selected and the result empty
```

rouault:
Thank you for your high quality pull requests, both in description, code and added unit tests!

rouault (approved):

PR #7041, opened 2024-03-07
flatgeobuf: add missing error name for MS_FGBERR

PPazderski:
For native flatgeobuf support a new error code was added but no name / description text to the `ms_errorCodes` array.

To reproduce it would be enough to trigger any `msSetError` which uses `MS_FGBERR` but I think the change is clear enough by itself. In all cases I saw this error the string "(null)" was printed instead because printf was involved at some point and the memory after the last array entry was always null. (and I produced a lot of those error messages while reproducing #7040)

### DIFF
--- a/src/maperror.c
+++ b/src/maperror.c
@@ -86,7 +86,8 @@ static char *const ms_errorCodes[MS_NUMERRORCODES] = {
     "OpenGL renderer error.",
     "Renderer error.",
     "V8 engine error.",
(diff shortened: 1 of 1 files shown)


rouault:
good catch! I've added a comment in f8f9d528da25ab01aa3de5ceb4faf2b8a43e2f4f so we hopefully no longer fall into that trap

PR #7042, opened 2024-03-07
map2img: improve error handling for missing option arguments

PPazderski:
The map2img command is too optimistic when parsing commandline arguments. In many cases it simply access the option argument without checking if it was actually provided leading sometimes to segfaults for incomplete commandlines.

Note that my change not only improves error handling for options which had non so far but also slightly changes behaviour for some other options.
The `-e` has already a check and custom error message. That message changes a bit with my commit.
Also other options (e.g. all the debug options) had a check before but simply skipped the option if the argument was missing. E.g. `map2img -m file.map -all_debug` would silently ignore the debug argument due to the missing level value before but report an error with my change.

### DIFF
--- a/msautotest/misc/expected/map2img-argparse-c.txt
+++ b/msautotest/misc/expected/map2img-argparse-c.txt
@@ -1,0 +1,1 @@
+Argument -c requires an additional argument.
--- a/msautotest/misc/expected/map2img-argparse-d.txt
+++ b/msautotest/misc/expected/map2img-argparse-d.txt
@@ -1,0 +1,1 @@
+Argument -d needs 2 space separated arguments.
--- a/msautotest/misc/expected/map2img-argparse-e.txt
+++ b/msautotest/misc/expected/map2img-argparse-e.txt
@@ -1,0 +1,1 @@
+Argument -e needs 4 space separated arguments.
--- a/msautotest/misc/expected/map2img-argparse-layer_debug.txt
+++ b/msautotest/misc/expected/map2img-argparse-layer_debug.txt
@@ -1,0 +1,1 @@
+Argument -layer_debug needs 2 space separated arguments.
--- a/msautotest/misc/expected/map2img-argparse-o.txt
+++ b/msautotest/misc/expected/map2img-argparse-o.txt
@@ -1,0 +1,1 @@
+Argument -o requires an additional argument.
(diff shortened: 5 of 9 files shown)


rouault (approved):

rouault (changes requested):
There's actually an issue with the address sanitizer build (perhaps a memory leak?)

PPazderski (commented):
> perhaps a memory leak?

Actually yes. Good jobs.

PPazderski:
Ok. Still not satisfied. Will try to reproduce it tomorrow.

PPazderski:
Found the problem. For asan test it adds an error output redirect which screws my test.

https://github.com/MapServer/MapServer/blob/9240625b87880421b48bb5e66230275e08a2c6eb/msautotest/pymod/mstestlib.py#L669-L672

I added a workaround for that but that makes the test a bit wonky and dependent on the availability of `tee`. If you prefer I can take out the tests.

rouault:
> For asan test it adds an error output redirect

can you be more specific on what asan adds?

PPazderski:
> can you be more specific on what asan adds?

Sure. From my understanding the address sanitizer build runs the msautotest with --run_under_asan option.

The test runner then modifies the executed command in
https://github.com/MapServer/MapServer/blob/9240625b87880421b48bb5e66230275e08a2c6eb/msautotest/pymod/mstestlib.py#L669-L672
by adding a `2> asan.log` to collect error report from the address sanitizer.

My test was intended to run something like.

`map2img -m msautotest/misc/map2img-argparse.map -s 640 > map2img-argparse-s.txt 2>&1`

to get the error output from the incomplete commandline and compare it to the expected result.

But for the address sanitizer build the executed command becomes

`map2img -m msautotest/misc/map2img-argparse.map -s 640 > map2img-argparse-s.txt 2>&1 2>result/map2img-argparse-s.txt.asan.txt`

The second error redirect overrides the first and the test output file remains empty.

rouault:
hum maybe then add a [ASAN_SKIP] marker for those tests and have mstestlib.py recognize it to ignore them?

PPazderski:
> hum maybe then add a [ASAN_SKIP] marker for those tests and have mstestlib.py recognize it to ignore them?

Added it as [ASAN_SKIP_LOG]. The test can still be run with address sanitizer.

rouault:
@PPazderski Looks good to me. Would you mind reworking the git commit history a bit to have the msautotest/pymod/mstestlib.py change in a first commit, and all the other changes in a following one?

geographika:
@rouault - it looks like @PPazderski updated the commit history - is this ok to merge now?

PR #7048, opened 2024-03-10
msRedactCredentials(): fix various issues (fixes #7046)

### LINKED ISSUE #7046: msRedactCredentials removes too much content especially for postgis connections
The function to hide passwords in debug and error output removes to much content.

Expected:

Only the password is removed.

```
msPostGISLayerOpen called: the_geom from something
msPostGISLayerOpen: No connection in pool, creating a fresh one.
msPostGISLayerOpen: Connection failure.
Database connection failed (invalid port number: "99999"
) with connect string 'dbname=nonexistingdatabase user=AzureDiamond password=******* port=99999'
Is the database running? Is it allowing connections? Does the specified user exist? Is the password valid? Is the database on the standard port? in msPostGISLayerOpen()msPostGISLayerOpen(): Query error. Database connection failed. Check server logs for more details.Is the database running? Is it allowing connections? Does the specified user exist? Is the password valid? Is the database on the standard port?
msDrawMap(): Image handling error. Failed to draw layer named 'not_working_connection'.
msDrawMap(): Image handling error. Failed to draw layer named 'not_working_connection'. <br>
msPostGISLayerOpen(): Query error. Database connection failed. Check server logs for more details.Is the database running? Is it allowing connections? Does the specified user exist? Is the password valid? Is the database on the standard port? <br>
msPostGISLayerIsOpen called.
```

Actual:

In most cases the password and all content after is removed.

```
msPostGISLayerOpen called: the_geom from something
msPostGISLayerOpen: No connection in pool, creating a fresh one.
msPostGISLayerOpen: Connection failure.
Database connection failed (invalid port number: "99999"
) with connect string 'dbname=nonexistingdatabase user=AzureDiamond password=*********************************************************************************************************************************************************************************************msPostGISLayerOpen(): Query error. Database connection failed. Check server logs for more details.Is the database running? Is it allowing connections? Does the specified user exist? Is the password valid? Is the database on the standard port?
msDrawMap(): Image handling error. Failed to draw layer named 'not_working_connection'.
msDrawMap(): Image handling error. Failed to draw layer named 'not_working_connection'. <br>
msPostGISLayerOpen(): Query error. Database connection failed. Check server logs for more details.Is the database running? Is it allowing connections? Does the specified user exist? Is the password valid? Is the database on the standard port? <br>
msPostGISLayerIsOpen called.
```

The attached example shows the problem by logging a failed postgis connection.

[redact_credentials.zip](https://github.com/MapServer/MapServer/files/14540978/redact_credentials.zip)

## MapServer version

Since 8.0.1.

## My opinion on the issue

This is an annoying and at the same time nasty and difficult problem. It is related to #6784.

https://github.com/MapServer/MapServer/blob/9240625b87880421b48bb5e66230275e08a2c6eb/src/maperror.c#L355-L363

When processing a debug or error message it will first search for `password=` and remove anything after until the first space.
Then it will take the redacted message will again search for `password=` and will find it again and now remove anything after until the first semicolon.

Apparently my example message has no semicolon so everything after `password=` is replaced.

I have no good idea how to fix this in general. At the point this redaction is applied there is no context, only a plain message and no way to know the required delimiter for a particular message for sure.
In the extrem there is no way to distinguish a OGR MySQL connection with connection string `MySQL:test,user=root,password=mysql,port=3306` and a PostGIS connection where my connection string is for some reason `host=myhost dbname=gis user=name password=mysql,port=3306 port=5432`.
And keywords could be case insensitive... :/

But I have two suggestions to improve the situation.

1. One annoyance is that the redaction also removes the newline (it is not clearly visible in my example because the postgis debug message happens to miss the newline anyways). Would already help if the replace function stops on newline.

2. While writing this issue I noticed that there are more possible database connection strings than the three mentioned in `msRedactCredentials`. I added such a case in my example.zip inspired from https://mapserver.org/input/vector/mysql.html#create-mapserver-layer.
The second thing I noticed is that the MapServer documentation and #6784 mention `pwd` for MSSQL as keyword but in code the MSSQL case (according to the comment) has `password`. If this was a mistake and `pwd` till semicolon is enough for MSSQL the second `msRedactString` could be removed which would fix the problem for PostGIS.

rouault:
and add a minimum C/C++ unit test

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1126,3 +1126,12 @@ install(
   FILES ${PROJECT_SOURCE_DIR}/etc/mapserver-sample.conf
   DESTINATION ${CMAKE_INSTALL_SYSCONFDIR}/
 )
+
+option(BUILD_TESTING "Build unit test" ON)
+if(BUILD_DYNAMIC AND BUILD_TESTING)
+    enable_testing()
+    add_executable(unit_test tests/unit/test.cpp)
+    target_include_directories(unit_test PRIVATE src)
+    target_link_libraries(unit_test PRIVATE mapserver)
+    add_test(NAME unit_test COMMAND unit_test)
+endif()
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ all: cmakebuild
 
 cmakebuild: lexer parser
 	if test ! -s build/Makefile; then  mkdir -p build ; cd build ; cmake .. $(CMAKEFLAGS); fi
-	cd build && $(MAKE) $(MFLAGS)
+	cd build && $(MAKE) $(MFLAGS) && ctest -V
 
 cmakebuild_nocoverage: lexer parser
 	if test ! -s build_nocoverage/Makefile; then  mkdir -p build_nocoverage ; cd build_nocoverage ; cmake .. $(CMAKEFLAGS_NOCOVERAGE); fi
--- a/src/maperror.c
+++ b/src/maperror.c
@@ -340,12 +340,34 @@ char *msGetErrorString(const char *delimiter) {
   return (errstr);
 }
 
-void msRedactString(char *str, const char *keyword, const char delimeter) {
+static void msRedactString(char *str, const char *keyword) {
 
   char *password = strstr(str, keyword);
   if (password != NULL) {
+    const char chOptionDelimeter = password - str > 0 ? password[-1] : 0;
     char *ptr = password + strlen(keyword);
-    while (*ptr != '\0' && *ptr != delimeter) {
+    char chStringSep = *ptr;
+    if (chStringSep == '\'' || chStringSep == '"') {
+      ++ptr;
+    } else {
+      chStringSep = '\0';
+    }
+    /* Replace all characters from after equal sign to end of line, end of
+     * string, or end of quoted string.
+     */
+    while (*ptr != '\0' && *ptr != '\r' && *ptr != '\n') {
+      if (chStringSep == '\0') {
+        if (*ptr == chOptionDelimeter)
+          break;
+      } else {
+        if (*ptr == chStringSep) {
+          break;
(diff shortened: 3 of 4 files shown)


geographika:
Very nice!

There is one case I've come across from MSSQL when using the ODBC driver, curly braces are used to allow passwords to contain the delimiter, e.g. 

```
"\\SQL2019;DATABASE=msautotest;Driver={ODBC Driver 17 for SQL Server};pwd={Password;12!};uid=sa;
```

See bottom of https://learn.microsoft.com/en-us/sql/relational-databases/security/strong-passwords?view=sql-server-ver16

> If used in an OLE DB or ODBC connection string, a login or password containing special characters must be enclosed in braces and right braces must be escaped. For example, the password my}Pass;word must be specified in the connection string like PWD={my}}Pass;word}.

I'm not sure if this is a case worth handling. 

In my view, it may be best to update all places in the code that log connection strings and remove them completely from logging output. In the case of the MSSQL driver the connection string is visible as an error message sent to the client when, for example, a WFS request timesout when connecting to the database. Even without a password server names and user names are data that shouldn't be leaked.

PPazderski:
Good fix. Should work reliable.

rouault:
> There is one case I've come across from MSSQL when using the ODBC driver, curly braces are used to allow passwords to contain the delimiter, e.g.

new commit added

> it may be best to update all places in the code that log connection strings and remove them completely from logging output. In the case of the MSSQL driver the connection string is visible as an error message sent to the client when, for example, a WFS request timesout when connecting to the database

oh definitely for messages sent back to the client, we need to be even less verbose. You're probably in a better position to spot where this comes from if you can reproduce

rouault:
I've also added  "msRedactString(): modify to trim password to a single redacted character" because giving the exact length of a password is already giving too much information

PR #7054, opened 2024-03-22
Fix SVG symbol sizing when width and height are with physical units 

mapcentia:
SVG symbols having width and height attributes with physical units (px, in, cm, mm, pt, pc) will get sized from the viewBox if present. so a symbol like this:

```svg 
<svg xmlns="http://www.w3.org/2000/svg" width="5.5mm" height="5.5mm" viewBox="0 0 15 15">..</svg>
```

will get sized according to the viewBox. But 5.5mm is 20px, so the symbol will end up cropped on the map.

This patch will test if the symbol has intrinsic size and use rsvg_handle_get_intrinsic_size_in_pixels to get the pixel size and size the symbol from that.

### DIFF
--- a/src/mapcairo.c
+++ b/src/mapcairo.c
@@ -1024,7 +1024,16 @@ int msPreloadSVGSymbol(symbolObj *symbol) {
         height.unit == RSVG_UNIT_PX) {
       symbol->sizex = width.length;
       symbol->sizey = height.length;
-    } else if (!has_viewbox) {
+    } else if (has_width && width.unit != RSVG_UNIT_PERCENT && has_height &&
+               height.unit != RSVG_UNIT_PERCENT) {
+      gdouble width_px;
+      gdouble height_px;
+      rsvg_handle_get_intrinsic_size_in_pixels(cache->svgc, &width_px,
+                                               &height_px);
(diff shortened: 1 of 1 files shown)


mapcentia:
The build test fails because it uses Ubuntu 20.04 with rsvg 2.48. `get_intrinsic_size_in_pixels` is added in 2.52.

It's worth reading the docs for `get_intrinsic_dimensions`: https://gnome.pages.gitlab.gnome.org/librsvg/Rsvg-2.0/method.Handle.get_intrinsic_dimensions.html

It's not necessary to test the symbol for having a width and height any longer, because it will always have them (it a geometry property of the svg symbol)

geographika:
@mapcentia - thanks for the pull request.
Could it be updated to check for the RSVG version number so the code works with multiple versions, with a  check similar to:

https://github.com/MapServer/MapServer/blob/ba0d946f31fccc503da2e7c821043253b0e8dbad/src/mapcairo.c#L1009C1-L1009C36

```c
#if LIBRSVG_CHECK_VERSION(2, 46, 0)
```

mapcentia:
@geographika Yes. We'll try to setup a build environment based on 20.04.

PR #7055, opened 2024-03-25
Add -D_DEFAULT_SOURCE for strlcat/strlcpy provided by glibc >= 2.38.

sebastic:
As reported in [Debian Bug #1067651](https://bugs.debian.org/1067651) MapServer failed to build on Ubuntu noble with glibc 2.39.

`strlcpy` and `strlcat` were introduced in glibc 2.38 but require `-D_DEFAULT_SOURCE` to actually work after being detected by `check_function_exists()`.


rouault:
Failure on "Build / regular" CI target unrelated to that pull request, will be fixed per https://github.com/MapServer/MapServer/pull/7056

PR #7056, opened 2024-03-25
Add compatibility for CMake 3.29 for CSharp bindings on Linux

rouault:


geographika:
Good catch @rouault !
I was looking at recent SWIG changes for a possible breaking change. 
Would it better to pin the SWIG and CMake versions used in the CI to avoid random breaking changes, or is catching these issues as early as possible easier to find the cause?

https://github.com/MapServer/MapServer/blob/12a2fe6beb28e81d3336b10aa7203746b92b85f7/ci/setup.sh#L25

https://github.com/MapServer/MapServer/blob/12a2fe6beb28e81d3336b10aa7203746b92b85f7/ci/setup.sh#L33

rouault:
> I was looking at recent SWIG changes for a possible breaking change.

yes, took me a while before realizing that CMake was the culprit here.

There are upsides & downsides in having CI track moving targets. It is painful when it breaks, but it is also good to be able to catch early breaking changes for us, and potentially report bugs in those dependencies as early as possible

Issue #7058, opened 2024-03-28
centerline function causes GEOS errors


geographika:
When using the `centerline` function for labels GEOS can return an error, which then stops all rendering. For example:

```
  LAYER
    NAME "lake-labels"
    STATUS OFF
    TYPE LINE
    CONNECTIONTYPE OGR
    CONNECTION "/vsicurl/https://github.com/MapServer/MapServer-demo/raw/main/data/lakespy2.shp"

    GEOMTRANSFORM (centerline([shape]))
    # GEOMTRANSFORM (centerline(densify([shape], 0.1)))
    CLASS
    LABEL
      COLOR 255 255 255
      TEXT (initcap("[LAKE_NAME]"))
      TYPE truetype
      SIZE 12
      PARTIALS false
      POSITION cc
      FORCE TRUE
      ANGLE FOLLOW
    END #label

    END
```

Without using `densify` when zoomed out MapServer returns the following:

```xml
<ServiceExceptionReport xmlns="http://www.opengis.net/ogc" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.3.0" xsi:schemaLocation="http://www.opengis.net/ogc http://schemas.opengis.net/wms/1.3.0/exceptions_1_3_0.xsd">
<ServiceException> msDrawMap(): Image handling error. Failed to draw layer named 'lake-labels'. msGeomTransformShape(): Expression parser error. Failed to process shape expression: centerline([shape]) yyparse(): Expression parser error. Executing centerline failed. msGEOSCenterline(): GEOS library error. Centerline generation failed, try densifying the shapes. </ServiceException>
</ServiceExceptionReport>
```

Probably best to log this as an error, but to proceed with rendering the rest of the image as best as possible?
Note - adding `densify` causes rendering to timeout when using a dataset with `vsicurl`.

sdlime:
I can understand doing that. It would be nice to identify the features that
cause the issues to understand what's going on and maybe avoid the error in
the first place.

--Steve

geographika:
@sdlime I found the features causing the issue. I see you already have a check for error-ing on features like this at https://github.com/MapServer/MapServer/blob/5a943d5a00c815d562210337297df9138ffc03dc/msautotest/misc/centerline.map#L155

Below is a Mapfile to recreate the issue - just comment out `PROCESSING "NATIVE_FILTER=FID NOT IN (127, 1112)"` to see the error. 

[...]

PR #7060, opened 2024-03-29
Automatically Densify simple polygons when creating centerlines

geographika:
Automatically densify simple polygons before creating msGEOSVoronoiDiagram. 
Possible fix for #7058. 

The number of points and the value of 3 for msDensify are based on a test dataset. I'm not sure of general applicability.

### DIFF
--- a/src/mapgeos.c
+++ b/src/mapgeos.c
@@ -1284,7 +1284,13 @@ shapeObj *msGEOSCenterline(shapeObj *shape) {
     return NULL;
   }
 
-  shape2 = msGEOSVoronoiDiagram(shape, 0.0, MS_TRUE);
+  if (shape->numlines > 0 && shape->line[0].numpoints <= 6) {
+    // automatically densify simple polygons
+    shape2 = msDensify(shape, 3);
+    shape2 = msGEOSVoronoiDiagram(shape2, 0.0, MS_TRUE);
+  } else {
+    shape2 = msGEOSVoronoiDiagram(shape, 0.0, MS_TRUE);
+  }
   if (!shape2) {
     msSetError(MS_GEOSERR, "Voronoi diagram generation failed.",
                "msGEOSCenterline()");
--- a/src/mapgeos.c
+++ b/src/mapgeos.c
@@ -1286,8 +1286,12 @@ shapeObj *msGEOSCenterline(shapeObj *shape) {
 
   if (shape->numlines > 0 && shape->line[0].numpoints <= 6) {
     // automatically densify simple polygons
-    shape2 = msDensify(shape, 3);
-    shape2 = msGEOSVoronoiDiagram(shape2, 0.0, MS_TRUE);
+    shapeObj *densifiedShape = msDensify(shape, 3);
+    if (densifiedShape) {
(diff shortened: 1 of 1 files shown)


sdlime:
I think I actually had this in place early on with this work but teased things apart. I don’t recall why though. What you propose seems like a good idea. I was wondering the same thing though on number of points and value for densify. I thought I used some trivial test cases with a triangle and square early on. 

Really will need to test on this end.

src/mapgeos.c
```diff
@@ -1284,7 +1284,13 @@ shapeObj *msGEOSCenterline(shapeObj *shape) {
     return NULL;
   }
 
-  shape2 = msGEOSVoronoiDiagram(shape, 0.0, MS_TRUE);
+  if (shape->numlines > 0 && shape->line[0].numpoints <= 6) {
+    // automatically densify simple polygons
+    shape2 = msDensify(shape, 3);
+    shape2 = msGEOSVoronoiDiagram(shape2, 0.0, MS_TRUE);
```
rouault, line 1290: to fix memory leak:
```suggestion
    shapeObj *densifiedShape = msDensify(shape, 3);
    if (densifiedShape) {
        shape2 = msGEOSVoronoiDiagram(densifiedShape, 0.0, MS_TRUE);
        msFreeShape(densifiedShape);
        free(densifiedShape);
    }
```

geographika:
> I think I actually had this in place early on with this work but teased things apart. I don’t recall why though. What you propose seems like a good idea. I was wondering the same thing though on number of points and value for densify. I thought I used some trivial test cases with a triangle and square early on.

I thought you'd have considered this, and the arbitrary values are a worry. At first, I was trying to simply ignore/log the errors, but that seemed much harder to implement as the shape it processed during parsing, so the function can't simply log an error and return NULL. 

I ran into the issue whilst creating the MapServer tutorial - `centerline` is a great option to have, but running `densify` on a full dataset kills performance.

sdlime:
> > I think I actually had this in place early on with this work but teased things apart. I don’t recall why though. What you propose seems like a good idea. I was wondering the same thing though on number of points and value for densify. I thought I used some trivial test cases with a triangle and square early on.
> 
> I thought you'd have considered this, and the arbitrary values are a worry. At first, I was trying to simply ignore/log the errors, but that seemed much harder to implement as the shape it processed during parsing, so the function can't simply log an error and return NULL.
> 
> I ran into the issue whilst creating the MapServer tutorial - `centerline` is a great option to have, but running `densify` on a full dataset kills performance.

What constitutes a simple feature is more problematic. I'd think the we could figure out a densify value based on the distance between the points (there aren't that many of then - perhaps the shortest distance between points (or a fraction thereof).

geographika:
I was hoping there were some criteria for which features would work in the GEOS codebase (I presume this is where the feature fails) but not many details from what I can find https://github.com/search?q=repo%3Alibgeos%2Fgeos++GEOSVoronoiDiagram&type=code

src/mapgeos.c
geographika, line 1: Thanks @rouault for the code review!

geographika:
Closed - see discussions at #7058

Issue #5343, opened 2016-10-31
OPACITY not working with GetMap Requests for 7.0.2


jmckenna:
- small mapfile (and data and commands) to reproduce problem: http://gatewaygeomatics.com/dl/ticket-opacity-wms.zip (400 KB)

 - no problems using normal CGI

 - however, using 7.0.2 opacity is not applied for a GetMap request

 - GetMap output for 7.0.1 (notice parks polygons are transparent)
![mapserv1](https://cloud.githubusercontent.com/assets/1611709/19867792/02e9eb4e-9f84-11e6-884c-c00fb6f7ffbe.png)

 - GetMap output for 7.0.2 (parks polygons are not transparent)
![mapserv2](https://cloud.githubusercontent.com/assets/1611709/19867840/30c13978-9f84-11e6-9e5b-70511356b8f8.png)

 - interesting, if I try to execute the same GetMap request at the commandline....

```
mapserv -nh QUERY_STRING="MAP=opacity-wms.map&SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&BBOX=50.43646181818179741,-137.2127006060605936,81.69507636363634617,-85.115009696969679&CRS=EPSG:4326&WIDTH=990&HEIGHT=594&LAYERS=bathymetry,park&STYLES=,&FORMAT=image/png&DPI=96&MAP_RESOLUTION=96&FORMAT_OPTIONS=dpi:96&TRANSPARENT=TRUE%20HTTP/1.1" > test.png
```
the following message is returned (only with 7.0.2), and a map image without transparency is generated:
```
*** BUG ***
In ???: The expression bits == NULL || (rowstride_bytes % sizeof (uint32_t)) == 0 was false
Set a breakpoint on '_pixman_log_error' to debug
```

rouault:
@jmckenna Your request for me with latest state of branch-7-0 (so nearly identical to 7.0.2) on Linux with cairo-1.10.2 (the polygons are transparent both in standalone mode "mapserv -nh" or within mongoose web server).
Running it under Valgrind doesn't show any problem reported. Reviewing (quickly) changes between 7.0.1 and 7.0.2, nothing strikes as being related. Could it be that something else than MapServer changed between your 7.0.1 and 7.0.2 build, like the cairo version used ?

Hum, I just found https://github.com/mapserver/mapserver/issues/5181 which seems related. Is it possible that your build have pixman enabled ? (my build has pixman disabled)

[...]

PR #7065, opened 2024-05-30
Add 4 new COMPOSITE.COMPOP blending operations (in particular hsl-color)

rouault:
Cf https://pycairo.readthedocs.io/en/latest/reference/enums.html#cairo.Operator.HSL_HUE

- "hsl-hue": Creates a color with the hue of the source and the saturation and luminosity of the target.
- "hsl-luminosity": Creates a color with the luminosity of the source and the hue and saturation of the target. This produces an inverse effect to hsl-color
- "hsl-saturation": Creates a color with the saturation of the source and the hue and luminosity of the target. Painting with this mode onto a gray area produces no change.
- "hsl-color": Creates a color with the hue and saturation of the source and the luminosity of the target. This preserves the gray levels of the target and is useful for coloring monochrome images or tinting color images.

That last mode is typical the most useful, for example when draping a colorized terrain onto a hillshade layer, or to colorize a greyed LIDAR shaded surface model with a green mask of a canopy model.

Ambiant occlusion LIDAR:
ambiant_occlusion_lidar
combined with Canopy Model blended on top of it with "hsl-color" at 100% opacity
canopy_model
results in:
ambiant_occlusion_lidar_colored_with_canopy

----

Implemented in the Cairo backend using the corresponding Cairo operators (which ultimately go to using Pixman), or in the AGG backend, by manually extracting the related Pixman code and adapting it to AGG (or when building WITH_PIXMAN=ON, by using the Pixman operation)

Note: if using OPACITY < 100 with the Cairo backend (or AGG WITH_PIXMAN=ON), there's a bug in current Pixman versions that result in wrong colors. Cf https://gitlab.freedesktop.org/pixman/pixman/-/merge_requests/97 The adaptation in the AGG backend isn't affected by this bug.

(note: this is intended to be 8.4.0 material, at least shouldn't be a reason to delay 8.2.0)


jmckenna:
@rouault thanks for these additions.  Regarding Pixman, I cringe to remember the many [issues with OPACITY](https://github.com/MapServer/MapServer/issues/5343) in MapServer with `-DWITH_PIXMAN=1`, as all packagers since then have to disable PIXMAN in MapServer with `-DWITH_PIXMAN=0` ...or maybe I am misunderstanding your comments above with AGG?

rouault:
> Regarding Pixman, I cringe to remember the many [issues with OPACITY](https://github.com/MapServer/MapServer/issues/5343) in MapServer with `-DWITH_PIXMAN=1`,

This older pixman issue is not the same as the one I noticed for this development, which is only used by the 4 new operators.

rouault:

sdlime:
+1 from me.

PR #7067, opened 2024-05-31
Raster rendering: skip Byte alpha band with PROCESSING "SCALE=AUTO"

rouault:
Currently when using PROCESSING "SCALE=AUTO" on a raster with an alpha band, the alpha band is all scaled, which is undesirable, in particular if it is fully opaque, as the result of auto scaling on it is to generate a fully transparent result.


rouault:

PR #7068, opened 2024-05-31
Try to fix (as far as I can tell) false-positive CodeQL warnings

rouault:


rouault:
there are remaining warnings I can't figure out how to workaround. Filed upstream as https://github.com/github/codeql/issues/16643

PR #7069, opened 2024-05-31
GDAL/raster rendering: add a GAMMA processing option

rouault:
to perform gamma correction on rasters (except in raw modes)


rouault:

PR #7073, opened 2024-06-08
[DO NOT MERGE] Add support for using PCRE2 instead of POSIX regex

rouault:
This is an attempt at using pcre2posix [https://www.pcre.org/current/doc/html/pcre2posix.html] which has the same API as POSIX regex
This builds. However we have failures at runtime like the following, likely do the regex dialect of PCRE2 and GNU regex not being exactly the same:
```
msCGILoadMap(): Web application error. CGI variable &quot;map&quot; fails to validate.
msEvalRegex(): Regular expression error. Failed to compile expression ([/\]{2}|[/\]?\.+[/\]|,).
```


rouault:
Pull request slightly modified to use PCRE2 only if explicitly enabled with -DWITH_PCRE2=ON,  except on
Windows if PCRE2 is found and REGEX_DIR is not defined.

cmake/FindPCRE2.cmake
geographika, line 1: Just a question in the Find*.cmake files - is there a collection of "official" files, or is it just a case of finding and picking one? Should we keep track of the original location in case there are fixes in the future we want to copy into the MapServer versions?

rouault: > is there a collection of "official" files,

CMake itself sometimes ship with some of them, but that's not the case for PCRE2 (https://gitlab.kitware.com/cmake/cmake/-/tree/master/Modules?ref_type=heads).
I've started from the FindPCRE2.cmake from GDAL, but I've adapted it since GDAL doesn't use the libpcre2-posix library but the libpcre2-8 one. So basically MapServer FindPCRE2.cmake is its own upstream

rouault:
This PR is ready as far as I'm concerned

Issue #7019, opened 2024-01-25
data loss on getmap requests when vector data is reprojected EPSG:2056 to EPSG:4326


ltclm:
## Expected behavior and actual behavior.

When zooming into a reprojected polygon layer we expect that all the features are displayed in all scales. The original data and the mapfile configuration is in EPSG:2056. The strange behaviour and the data loss has been observed when requesting the map in EPSG:4326.

## Steps to reproduce the problem.
Using the mapfile config [1] and the data index.shp from the attachment [2],  start zooming into the center of the 4 polygons in EPSG:4326. After a certain zoom/scale some data gets lost. The same effect occurs with tiled wms queries (256x256), but much earlier.

## Attach simple test case (drag/drop it here)
here are some screenshots and getmap requests of the observed data loss with EPSG:4326 GetMap Requests:
http://localhost:9178/local/?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&BBOX=47.70122100713624036,8.631010777621343166,47.73459661396240961,8.673423389810198003&CRS=EPSG:4326&WIDTH=1591&HEIGHT=1252&LAYERS=test&STYLES=&FORMAT=image/jpeg&DPI=96&MAP_RESOLUTION=96&FORMAT_OPTIONS=dpi:96
![image](https://github.com/MapServer/MapServer/assets/5286659/ee7b6b65-1b2c-4e0f-b65b-4cefbff32d35)
```
service-wms_1  |  [warn] [pid 10016] mod_fcgid: stderr: CGI Request 4 on process 10071
service-wms_1  |  [warn] [pid 10016] mod_fcgid: stderr: msDrawMap(): rendering using outputformat named jpeg (AGG/JPEG).
service-wms_1  |  [warn] [pid 10016] mod_fcgid: stderr: msDrawMap(): WMS/WFS set-up and query, 0.000s
service-wms_1  |  [warn] [pid 10016] mod_fcgid: stderr: msDrawMap(): Layer 0 (test), 0.016s
service-wms_1  |  [warn] [pid 10016] mod_fcgid: stderr: msDrawMap(): Drawing Label Cache, 0.000s
service-wms_1  |  [warn] [pid 10016] mod_fcgid: stderr: msDrawMap() total time: 0.017s
service-wms_1  |  [warn] [pid 10016] mod_fcgid: stderr: msSaveImage(stdout) total time: 0.014s
[...]

PR #7074, opened 2024-06-09
Fix workaround for Proj versions 9.0.0 to 9.4.0

oshawa-connection:
Hello,

This is a fix for a workaround for a bug in Proj versions between version 9.0.0 and 9.4.0 (inclusive), which is described in this issue:

#7019 

The source of this problem was a workaround in mapserver for a bug in proj prior to version 9.4.1. The workaround checks for proj major versions between 6 and 9, and then applies a fix. However, versions 9.0.0 through 9.4.0 suffer from the original proj bug, and do not have the workaround applied.

I'm fairly confident of my change on line 1903, but on line 1909 I had not much idea what I was doing. I only tested requests in EPSG:4326 and EPSG:3857 which the workaround targets, I did not test with expanded projection strings. 

I am happy to write some tests + test data for this change if required!

### DIFF
--- a/src/mapproject.c
+++ b/src/mapproject.c
@@ -1900,13 +1900,13 @@ int msProjectRect(projectionObj *in, projectionObj *out, rectObj *rect) {
    */
   else {
     int apply_over = MS_TRUE;
-#if PROJ_VERSION_MAJOR >= 6 && PROJ_VERSION_MAJOR < 9
+#if PROJ_VERSION_MAJOR >= 6 && PROJ_VERSION_MAJOR < 9 || (PROJ_VERSION_MAJOR == 9 && PROJ_VERSION_MINOR <= 4 && PROJ_VERSION_PATCH < 1)
     // Workaround PROJ [6,9[ bug (fixed per
     // https://github.com/OSGeo/PROJ/pull/3055) that prevents datum shifts from
     // being applied when +over is added to +init=epsg:XXXX This is far from
     // being bullet proof but it should work for most common use cases
     if (in && in->proj) {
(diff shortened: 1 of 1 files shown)


rouault:
> The source of this problem was a workaround in mapserver for a bug in proj prior to version 9.4.1. The workaround checks for proj major versions between 6 and 9, and then applies a fix. However, versions 9.0.0 through 9.4.0 suffer from the original proj bug, and do not have the workaround applied.

hum, it is not obvious to me which change between PROJ 9.4.0 and 9.4.1 would have an influence on this, and PROJ 9 did have a fix related to +over handling. This would require more careful analysis. Could you do a git bisect on PROJ between 9.4.0 and 9.4.1 to find out which revision of PROJ would have changed behavior?

oshawa-connection:
Yes no problem, I will take a look!

oshawa-connection:
I made a mistake when installing proj from source - so mapserver's cmake picked up my system installed proj (version 8)... Sorry for this, I don't use cmake often, hopefully I haven't wasted much of your time on this.

However, the bug definitely still exists in proj 9 for some cases when +over is added- I am trying to hunt down more specific information on when/ why. If it's ok with you, I will leave this PR open until the problem is fixed in proj 9, and then the PROJ_VERSION_PATCH/ PROJ_VERSION_MINOR number could be changed to cover those cases. 

In the meantime I will try to get some more specifics of the problem with proj.

rouault:
@oshawa-connection Thanks for the thouroughful analysis. I believe a better fix would be https://github.com/MapServer/MapServer/pull/7121  . Can you give this a try?

oshawa-connection:
No problem, glad I could help :+1: Yes I agree, this PR isn't needed now!

Issue #7081, opened 2024-06-17
Breaking changes to Python MapScript in SWIG 4.3


geographika:
As discovered in #7080. Recent changes in SWIG has caused MapScript Python to stop building:

```
2024-06-15: wsfulton
            [Python] Removed deprecated pytuplehlp.swg file and t_output_helper.
            Use SWIG_Python_AppendOutput instead of t_output_helper.
```

https://github.com/swig/swig/commit/c2d20fbd6b8cbc51d2e46894b09ac3cfda192dfc
https://github.com/swig/swig/commit/462fb665af060feb113aacbd6f70f384aba9dcdd

Errors are as follows:

```
/home/runner/work/MapServer/MapServer/build/src/mapscript/python/CMakeFiles/pythonmapscript.dir/mapscriptPYTHON_wrap.c: In function ‘_wrap_styleObj_pattern_get’:
/home/runner/work/MapServer/MapServer/build/src/mapscript/python/CMakeFiles/pythonmapscript.dir/mapscriptPYTHON_wrap.c:13970:17: warning: implicit declaration of function ‘t_output_helper’ [-Wimplicit-function-declaration]
13970 |     resultobj = t_output_helper(resultobj,r);
      |                 ^~~~~~~~~~~~~~~
/home/runner/work/MapServer/MapServer/build/src/mapscript/python/CMakeFiles/pythonmapscript.dir/mapscriptPYTHON_wrap.c:13970:15: warning: assignment to ‘PyObject *’ {aka ‘struct _object *’} from ‘int’ makes pointer from integer without a cast [-Wint-conversion]
13970 |     resultobj = t_output_helper(resultobj,r);
      |               ^
	  
	  2024-06-16T21:37:37.0081961Z /home/runner/work/MapServer/MapServer/src/mapscript/python/pyextend.i:433: Warning 490: Fragment 't_output_helper' not found.
```	  

`t_output_helper` is used in pymodule.i:

```
%typemap(argout,fragment="t_output_helper,CreateTupleFromDoubleArray") (double** argout, int* pnListSize) 
{
   /* %typemap(argout) (double* argout, int* pnListSize)  */
  PyObject *r;
  r = CreateTupleFromDoubleArray(*$1, *$2);
  free(*$1);
  $result = t_output_helper($result,r);
}
```

geographika:
Note there are also several uses of  `t_output_helper` in `https://github.com/OSGeo/gdal/blob/master/swig/include/python/typemaps_python.i`

PR #7082, opened 2024-06-17
Pin SWIG version used for CI

geographika:
The master branch of SWIG is breaking Python MapScript - #7081. 
I've not found a backwards (or indeed forwards!) fix for this yet, so I think it may be simplest to pin SWIG to the last working version / released version, rather than using the master branch which could break the CI at any time. 

cc @rouault


geographika:
Replaced with a backwards compatible fix in #7083

PR #7083, opened 2024-06-17
Remove deprecated SWIG t_output_helper

geographika:
See #7081.

These changes work with the latest master branch of SWIG (4.3.0). However I can't seem to get both the main build and the MapScript builds to both work, see various logs/attempts at https://github.com/geographika/mapserver/actions

SWIG_Python_AppendOutput has been updated in 4.3.0 to have a new `is_void` parameter, that should be set to 1. 
Both build actions report SWIG 4.3.0, and most strangely both use the same setup script to install SWIG: https://github.com/MapServer/MapServer/blob/0b9fb9bc504d8f0ecbf3d479dd37a3a2f9746682/ci/setup.sh#L33

- build-mapscript-python.yml - this doesn't seem to pick up the latest SWIG master branch and fails with the following:

```
2024-06-17T13:59:11.6733264Z /home/runner/work/mapserver/mapserver/build/src/mapscript/python/CMakeFiles/pythonmapscript.dir/mapscriptPYTHON_wrap.c: In function ‘_wrap_styleObj_pattern_get’:
2024-06-17T13:59:11.6736854Z /home/runner/work/mapserver/mapserver/build/src/mapscript/python/CMakeFiles/pythonmapscript.dir/mapscriptPYTHON_wrap.c:13241:17: error: too many arguments to function ‘SWIG_Python_AppendOutput’
2024-06-17T13:59:11.6741095Z 13241 |     resultobj = SWIG_Python_AppendOutput(resultobj,r,1);
2024-06-17T13:59:11.6742811Z make[3]: Leaving directory '/home/runner/work/mapserver/mapserver/build'
```

- build.yml - this does pick up the latest SWIG version. If I don't add a 3rd parameter to SWIG_Python_AppendOutput I get the following errors:

```
[ 91%] Building C object src/mapscript/java/CMakeFiles/javamapscript.dir/edu/umn/gis/mapscript/mapscriptJAVA_wrap.c.o
/home/runner/work/mapserver/mapserver/build/src/mapscript/python/CMakeFiles/pythonmapscript.dir/mapscriptPYTHON_wrap.c: In function ‘_wrap_styleObj_pattern_get’:
/home/runner/work/mapserver/mapserver/build/src/mapscript/python/CMakeFiles/pythonmapscript.dir/mapscriptPYTHON_wrap.c:13970:17: error: too few arguments to function ‘SWIG_Python_AppendOutput’
13970 |     resultobj = SWIG_Python_AppendOutput(resultobj,r);
```


src/mapscript/python/pymodule.i
```diff
@@ -82,13 +82,13 @@ CreateTupleFromDoubleArray( double *first, unsigned int size ) {
...
 
-%typemap(argout,fragment="t_output_helper,CreateTupleFromDoubleArray") (double** argout, int* pnListSize) 
+%typemap(argout,fragment="CreateTupleFromDoubleArray") (double** argout, int* pnListSize) 
 {
    /* %typemap(argout) (double* argout, int* pnListSize)  */
   PyObject *r;
   r = CreateTupleFromDoubleArray(*$1, *$2);
   free(*$1);
-  $result = t_output_helper($result,r);
+  $result = SWIG_Python_AppendOutput($result,r,1);
```
rouault, line 91: I presume we should do something along those lines (untested):
```suggestion
%#if SWIGVERSION >= 0x040300
  $result = SWIG_Python_AppendOutput($result,r,$isvoid);
%#else
  $result = SWIG_Python_AppendOutput($result,r);
%#endif
```

rouault:
I guess we can backport this to 8.2 so it is SWIG 4.3 ready and drop #7082 ?

geographika:
> I guess we can backport this to 8.2 so it is SWIG 4.3 ready and drop #7082 ?

Yes. Although the mystery of what version of SWIG is used when building build-mapscript-python.yml. There was a missing space in-between flags that may have caused it..

I'm still not sure if it is best to know a dependency broke something early via unrelated changes, or to pin everything and make updates when ready to try and address them.

rouault:
> I'm still not sure if it is best to know a dependency broke something early via unrelated changes, or to pin everything and make updates when ready to try and address them.

yes there are pros and cons to both approaches. Sometimes it is easier to catch the breakage as early as possible as looking at the latest commits of the dependency that broke helps quickly spotting the reason, while if you realize it months later, that might be much more difficult

src/mapscript/python/pymodule.i
```diff
@@ -82,13 +82,17 @@ CreateTupleFromDoubleArray( double *first, unsigned int size ) {
...
-%typemap(argout,fragment="t_output_helper,CreateTupleFromDoubleArray") (double** argout, int* pnListSize) 
+%typemap(argout,fragment="CreateTupleFromDoubleArray") (double** argout, int* pnListSize) 
 {
    /* %typemap(argout) (double* argout, int* pnListSize)  */
   PyObject *r;
   r = CreateTupleFromDoubleArray(*$1, *$2);
   free(*$1);
-  $result = t_output_helper($result,r);
+  %#if SWIG_VERSION >= 0x040300
+    $result = SWIG_Python_AppendOutput($result,r,1);
```
rouault, line 92: Did you try the following ? That's what they suggest in SWIG readme:
```suggestion
    $result = SWIG_Python_AppendOutput($result,r,$isvoid);
```

geographika: I tried that originally, but without the SWIG version check. Updated now.

Issue #5697, opened 2018-11-07
Layer compositer problems


szekerest:
1. When using the layer compositer with the agg renderer and set 
layer.compositer.comp_op = CompositingOperation.MS_COMPOP_SRC_OVER
expliticly (or specify a COMPOSITE block in the mapfile), the layer pixel values are decremented.

If the compositer block is not set (ie. layer.compositer = null, or not COMPOSITE block in the mapfile) the pixel values are not changed.

2. The COMPOSITE setting is not properly exposed in mapscript, we can initialize the compisite block by using setOpacity, but no reverse operation added (to safely destroy the compositer without leaking memory).

szekerest:
I've run into this problem again (after 6 years) with the latest mapsever version, but actually, the issue is related to the incorrect opacity calculation in **mapagg.cpp**, where **unsigned(opacity * 2.55)** can yield 254 instead of 255 is some cases. The correct formula should be **unsigned(MS_NINT(opacity * 2.55))** with the same way as what we already do in other locations like **mapfile.cpp,mapkmlrenderer.cpp, mapogsld.c, maputil.c**.

PR #7085, opened 2024-06-21
Fix opacity calculation problem with the agg renderer (#5697)

szekerest:


szekerest:
@jmckenna @rouault I wonder how to recreate the expected files for the failed tests. I created an ubuntu VM, and run msautotest, but most of the generated files are binary different than the files from the expected directories.

rouault:
> I created an ubuntu VM, and run msautotest, but most of the generated files are binary different than the files from the expected directories.

you need a 20.04 one to get the "right" version of the libraries. As I've it handy, I've just checkout your branch and pushed the updated expected results

szekerest:
> > I created an ubuntu VM, and run msautotest, but most of the generated files are binary different than the files from the expected directories.
> 
> you need a 20.04 one to get the "right" version of the libraries. As I've it handy, I've just checkout your branch and pushed the updated expected results

Great, thanks @rouault 
I'll give it a try creating a 20.04 VM. I wonder if we could add that requirement into the msautotests docs for further reference.

rouault:
backport done in https://github.com/MapServer/MapServer/pull/7087

geographika:
@szekerest @rouault - the [vagrant](https://mapserver.org/development/tests/vagrant.html) approach works well for me to get the exact binary outputs on a Windows machine. Isn't this the recommended approach?

rouault:
> Isn't this the recommended approach?

yes, that can be convenient .Added to https://github.com/MapServer/MapServer/pull/7086

PR #7087, opened 2024-06-21
[8.2 backport] Fix opacity calculation problem with the agg renderer (#5697)

### LINKED ISSUE #5697: Layer compositer problems

rouault:
Fixes #5697

Backport of https://github.com/MapServer/MapServer/pull/7085


szekerest:
@rouault Can we merge this in?

Issue #7079, opened 2024-06-16
Potential memory corruption in mapproject.c


szekerest:
The projection context handling in mapproject.c is not thread safe. Under certain contitions it may result in double free corruption issues.
We have managed to extract a test case with an asan dump log related to this problem:

```
==10772==ERROR: AddressSanitizer: heap-use-after-free on address 0x12105767a3a0 at pc 0x7ffa9e2514b7 bp 0x0098e12ff3a0 sp 0x0098e12ff3a8
READ of size 4 at 0x12105767a3a0 thread T-1
    #0 0x7ffa9e2514b6 in proj_context_errno D:\Devl\NiallG\PROJ\src\4D_api.cpp:2405
    #1 0x7ffa9e26f30f in proj_destroy D:\Devl\NiallG\PROJ\src\malloc.cpp:108
    #2 0x7ffaac79e627 in msFreeProjection E:\builds-t1\src\mapserver-intramaps\mapproject.c:635
    #3 0x7ffaac81868c in freeLayer E:\builds-t1\src\mapserver-intramaps\mapfile.c:3697
    #4 0x7ffaca5b37ca in SWIGRegisterByteArrayCallback_mapscript+0x24ca (D:\Users\NiallG\Downloads\MapEngine Proj Test\intramaps-mapengine\MapEngine.LoadTest\bin\Release\mapscript.DLL+0x1800037ca)
    #5 0x7ffaca5ce855 in CSharp_OSGeofMapServer_delete_layerObj___+0x25 (D:\Users\NiallG\Downloads\MapEngine Proj Test\intramaps-mapengine\MapEngine.LoadTest\bin\Release\mapscript.DLL+0x18001e855)
    #6 0x7ffa691d6fd8  (<unknown module>)

0x12105767a3a0 is located 32 bytes inside of 520-byte region [0x12105767a380,0x12105767a588)
freed by thread T0 here:
    #0 0x7ffa9e6c0183 in operator delete D:\a\_work\1\s\src\vctools\asan\llvm\compiler-rt\lib\asan\asan_win_delete_scalar_size_thunk.cpp:41
    #1 0x7ffa9e25145c in proj_context_destroy D:\Devl\NiallG\PROJ\src\4D_api.cpp:2487
    #2 0x7ffaac79fde4 in msProjectionContextUnref E:\builds-t1\src\mapserver-intramaps\mapproject.c:408
[...]

PR #7088, opened 2024-06-22
Make projection context ref counting thread safe (#7079)

szekerest:


rouault:
@szekerest I'm wondering if it is not hiding a more serious concern: a PROJ context should not be used by multiple threads simultaneously. So if you need to make ref counting thread-safe, this is a sign that a PROJ context might be used simultaneously at some point. Can you share more details about the issue?

szekerest:
@rouault According to the stack traces attached to the corresponding ticket, the handling of refcount in msProjectionContextUnref is not thread safe so proj_destroy can be called multiple times. This causes an issue of a client running an ASP.NET (mapscript) app on the server under a heavy load condition and causes the host process to crash approx 20 times a day. They have just reported that by applying this fix, the issue has gone after 3 days of continuous operation.

rouault:
> the handling of refcount in msProjectionContextUnref is not thread safe so proj_destroy can be called multiple times.

yeah, I've no doubt that this fix improves things, but I'm wondering if it is not just a workaround for a more serious issue, and multi-threading bugs could still happen (perhaps less frequent)
I suspect msProjectionInheritContextFrom() should refuse to reuse the initial context if the threads are different, or perhaps clone more things. I'll try to come up with an alternative fix

rouault:
@szekerest can you give https://github.com/MapServer/MapServer/pull/7089 a try ?

szekerest:
> @szekerest can you give #7089 a try ?

@rouault That seems to solve the problem in my test case rendering 1200 maps, however it is approx 3 times slower than the previous approach and the memory consumption is bigger. I'll let the client to test that in their production env.

szekerest:
@rouault We finally got a chance to test the mapserver proj9 fix on the client's production servers (the one that made its way into the mapserver code) and the results under load wasn't good, customers started to log support issues to complain about performance.
We had to revert to using the initial fix (as per https://github.com/MapServer/MapServer/pull/7088/commits/19abdfcac45c86cc979605478c445434f7859fa8), see CPU usage with the mapserver fix, then it drops back dramatically after we revert to using the initial fix

image

jmckenna:
nudge here as this was tagged for the 8.2.2 release (see [Milestone](https://github.com/MapServer/MapServer/milestone/63)) which is set to be released this Friday

Issue #7090, opened 2024-06-25
SWIG mapscript failure with out of source builds


szekerest:
When doing out of source build with SWIG mapscript enabled (ie JAVA or CSHARP) and PROTOBUFC (MVT) is not enabled, we get the following build error:

`..mapscript.i(255): error : Unable to find '..\..\mapserver-version.h'`

This is because `include_directories(${PROJECT_BINARY_DIR}/src/mapscript/)` hasn't been added to the corresponding CMakeLists.txt files so that the compiler generated mapserver-version.h in the build directory cannot be included by SWIG.

However, when PROTOBUFC is enabled, it includes `${PROJECT_BINARY_DIR}/renderers/mvt` which causes the `'..\..\mapserver-version.h'` reference to be found "accidentally".

szekerest:
Closing as completed

akrherz:
I think I am hitting this issue presently on the `branch-8-2` within a created build folder

```
[ 94%] Swig compile ../mapscript.i for php7
/home/akrherz/projects/mapserver/src/mapscript/mapscript.i:255: Error: Unable to find '../../mapserver-version.h'
make[2]: *** [src/mapscript/phpng/CMakeFiles/php_mapscriptng_swig_compilation.dir/build.make:76: src/mapscript/phpng/CMakeFiles/php_mapscriptng.dir/mapscriptPHP7.stamp] Error 1
make[2]: *** Deleting file 'src/mapscript/phpng/CMakeFiles/php_mapscriptng.dir/mapscriptPHP7.stamp'
make[1]: *** [CMakeFiles/Makefile2:585: src/mapscript/phpng/CMakeFiles/php_mapscriptng_swig_compilation.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
```

jmckenna:
Thanks @akrherz .  It seems that the fix submitted through https://github.com/MapServer/MapServer/pull/7091 fixed all MapScripts (CSharp, Java, Perl, Perl, Python, Ruby) but missed PHPNG.  I'll add it now through a PR.  Thanks again @akrherz for catching this!

jmckenna:
Fixed through https://github.com/MapServer/MapServer/pull/7108, closing.

PR #7091, opened 2024-06-25
Fix SWIG mapscript failure with out of source builds (#7090)

szekerest:


jmckenna:
thanks @szekerest great catch!

PR #7096, opened 2024-06-28
Use EPSGTreatsAsLatLong and EPSGTreatsAsNorthingEasting to determine inverted axis

PatrikSylve:
Hi, I updated the script that generates the table of EPSGs that have inverted axis order.  It will append EPSGs if EPSGTreatsAsNorthingEasting or EPSGTreatsAsLatLong return true. 

I encountered some issues when downloading over WFS using srsName=EPSG:32661, the output GML had the coordinates in EN-order. Looking into the proj definition I saw that the axes were defined as NE, but the orientation were the same (south) which is why it was not added in mapaxisorder.h.  

 ```
projinfo EPSG:32661
...
        AXIS["northing (N)",south,
            MERIDIAN[180,
                ANGLEUNIT["degree",0.0174532925199433]],
            ORDER[1],
            LENGTHUNIT["metre",1]],
        AXIS["easting (E)",south,
            MERIDIAN[90,
                ANGLEUNIT["degree",0.0174532925199433]],
            ORDER[2],
            LENGTHUNIT["metre",1]],
```


rouault:
Merging. The CI failure on AppVeyor is unrelated, as it also occurs on recent commits in main (such as https://github.com/MapServer/MapServer/commit/20886024151c5befc9c2876765353929abf23b4f)

PR #7102, opened 2024-07-09
Implement Windows, Linux, and OSX CI builds

hobu:
This PR implements CI builds for Windows, Linux, and OSX based on @conda-forge.


geographika:
Thanks for this @hobu !
I'm still getting familiar with the conda environment, so are the statements below accurate?

This pull request:

- uses the conda build approach from https://github.com/conda-forge/mapserver-feedstock and packages from the conda ecosystem to build MapServer for Windows, Linux, and OSX
- will ensure that any MapServer source code changes are tested against this build approach so any changes that could break conda MapServer packaging are caught early
- introduces CI building of MacOS to MapServer (would conda be the recommended way to build MapServer on a Mac?)
- `DWITH_PYTHON=0` is not required as this is to test the MapServer build, and https://github.com/conda-forge/mapserver-feedstock will be used to build and create a conda package with MapScript (or is there a more specific reason Python is excluded)

@msmitherdc was showing me `mapscript` on the Linux conda build at FOSS4G-Europe. Being able to run `conda install mapserver` on Windows (including Python MapScript) will be a massive boost for MapServer, I can think of several use cases already.

hobu:
> uses the conda build approach from https://github.com/conda-forge/mapserver-feedstock and packages from the conda ecosystem to build MapServer for Windows, Linux, and OSX

This PR does not use `conda build`. It instead creates a conda environment using the definition in `./ci/conda/environment.yml` and then invokes cmake and ninja to build. For GDAL and PDAL, we do have the `conda build` (or `conda mambabuild` rather 😉) scenarios tested but I didn't want to start there as it is extra complication.

> introduces CI building of MacOS to MapServer 

Yep, arm64 is the architecture built. It would be possible to do x86_64, but maybe not worth it going forward.

>  is not required as this is to test the MapServer build,

With a little more CMake configuration work, it will be possible to get Python MapScript building on OSX, Windows, and Linux for the Conda Forge recipes.

rouault:
(I've reworked history a bit to separate fixes from the CI addition)

geographika:
> This PR does not use `conda build`. It instead creates a conda environment using the definition in `./ci/conda/environment.yml` and then invokes cmake and ninja to build. For GDAL and PDAL, we do have the `conda build` (or `conda mambabuild` rather 😉) scenarios tested but I didn't want to start there as it is extra complication.

So the main difference is it uses cross-platform Conda packages listed in `environment.yml` to get all the MapServer build dependencies, rather than relying on the SDK from GISInternals.com (for Windows) or Debian packages on Linux?

geographika (approved):

PR #7103, opened 2024-07-09
msautotest: avoid failed tests on AppVeyor to break the build

rouault:
For a yet-to-be-determined reasons, test_rq_10() and test_rqtest_12() have started to fail on AppVeyor builds somewhere between June 28th (https://github.com/MapServer/MapServer/commit/8401b7b6f86a8723aa38ce958894df0162b6fb9f) and July 8th 2024
(https://github.com/MapServer/MapServer/commit/20886024151c5befc9c2876765353929abf23b4f) No MapServer source code change seems to be involved.

### DIFF
--- a/msautotest/mspython/test_rq.py
+++ b/msautotest/mspython/test_rq.py
@@ -397,6 +397,9 @@ def test_rq_9():
 # Open a classified map and post a point query.
 
 
+@pytest.mark.xfail(
+    "APPVEYOR" in os.environ, reason="fail for unknown reason on AppVeyor builds"
+)
 def test_rq_10():
 
     map = mapscript.mapObj(get_relpath_to_this("../gdal/classtest1.map"))
@@ -452,6 +455,9 @@ def test_rq_10():
 # Issue another point query, on colored text.
 
 
+@pytest.mark.xfail(
+    "APPVEYOR" in os.environ, reason="fail for unknown reason on AppVeyor builds"
+)
 def test_rqtest_12():
 
     map = mapscript.mapObj(get_relpath_to_this("../gdal/classtest1.map"))


geographika:
@rouault - I think the cause is the switch to GDAL 3.9.1 in the Windows SDK which was pushed last week: https://github.com/gisinternals/buildsystem/commit/8d6e000e47842dd9e88c8fe13330b821b7230742

Both failing tests use `classtest1.map` which has a PNG8 driver. Any idea if the `AGG/PNG8` driver had changes from GDAL 3.8 to 3.9?

```
OUTPUTFORMAT
  NAME png8_t
  DRIVER "AGG/PNG8"
  IMAGEMODE RGB
  TRANSPARENT ON
END
```

There is also a note in test_rq.py:

```
    # In the future the raster query may well stop returning
    # "offsite" pixels and we will need to update this test.
```

rouault:
@geographika Interesting catch. Those 2 tests actually pass on my dev setup on Linux which is on GDAL 3.10dev. The connection between AGG/PNG8 and GDAL isn't immediately obvious to me

rouault:
@geographika Would be great if you had a chance of investigating. In the meantime, I'm going to merge this to keep CI green for pull requests

geographika:
@rouault - I can confirm the test passed with an older version of the SDK with GDAL 3.8, and when I updated the SDK the test fails. 
Looking at SDK changes, the only recent change is the GDAL update: https://github.com/gisinternals/buildsystem/commits/main/

```python
import mapscript
map = mapscript.mapObj(r"D:\GitHub\mapserver\msautotest\gdal\classtest1.map")
layer = map.getLayer(0)

pnt = mapscript.pointObj()
pnt.x = 88.5
pnt.y = 7.5

# when using the newer SDK the following returns MS_FALSE, in an 
# older SDK (with same MapServer main branch it returns 0)
# with an older SDK it returns MS_TRUE and a single result
layer.queryByPoint(map, pnt, mapscript.MS_SINGLE, 10.0)
```

`msRasterQueryByPoint` uses GDAL objects, so maybe there is something in that. Will continue debugging.

geographika:
`msRasterQueryByRect` fails to open https://github.com/MapServer/MapServer/blob/main/msautotest/gdal/data/grid.gif and `hDS` is always NULL. `gdalinfo` runs fine on this file though:

```c
    msAcquireLock(TLOCK_GDAL);
    if (!layer->tileindex) {
      char **connectionoptions =
          msGetStringListFromHashTable(&(layer->connectionoptions));
      hDS = GDALOpenEx(decrypted_path, GDAL_OF_RASTER, NULL,
                       (const char *const *)connectionoptions, NULL);
```

https://github.com/MapServer/MapServer/blob/389f8401903f3326f805a9bfeb0dbae0ce25f290/src/maprasterquery.c#L783

Hard coding an absolute path to this file in the Mapfile also fails.

rouault:
@geographika Does that happen with any GIF file ?
What if you add ``fprintf(stderr, "gif driver = %p\n", GDALGetDriverByName("GIF"));`` before GDALOpenEx() ?

geographika:
@rouault - seems to happen with any GIF.

The output statement when adding the `fprintf` gives:
```
>>> layer.queryByPoint(map, pnt, mapscript.MS_SINGLE, 10.0)
gif driver = 0000000000000000
gif driver = 0000000000000000
1
```

If I list the GDAL formats in the SDK build GIF is listed:

```
D:\MapServer\release-1930-x64-dev\release-1930-x64>gdalinfo --formats
Supported Formats:
  ECW -raster- (rw+): ERDAS Compressed Wavelets (SDK 5.5) (*.ecw)
  JP2ECW -raster,vector- (rw+v): ERDAS JPEG2000 (SDK 5.5) (*.jp2)
  FITS -raster,vector- (rw+): Flexible Image Transport System (*.fits)
  GIF -raster- (rwv): Graphics Interchange Format (.gif) (*.gif)
```

`gdal_GIF.dll` is present in "D:\MapServer\release-1930-x64-dev\release-1930-x64\bin\gdal\plugins\gdal_GIF.dll"

**Solved!!**

Was the GIF driver moved to a plugin in GDAL 3.9?

Adding the following prior to running the Python script resolves the issue and returns the result!

```
$env:GDAL_DRIVER_PATH="D:\MapServer\release-1930-x64-dev\release-1930-x64\bin\gdal\plugins"
```

I'll create a pull request with a fix in the morning.

rouault:
> Was the GIF driver moved to a plugin in GDAL 3.9?

it is indeed now an *option*.  And I see gisinsternals does enable GDAL_ENABLE_PLUGINS=ON at https://github.com/gisinternals/buildsystem/blob/8d6e000e47842dd9e88c8fe13330b821b7230742/Makefile#L792 . Perhaps adding -DGDAL_ENABLE_DRIVER_GIF_PLUGIN=OFF to that line could be appropriate

geographika:
cc @szekerest - any thoughts on the above? we can set the `GDAL_DRIVER_PATH` for the tests on Windows, or include the GIF driver.

PR #7105, opened 2024-07-09
Windows MapScript for Conda build

hobu:
Follow up to #7102 to add MapScript builds as well on all three platforms


geographika (approved):

geographika:
Thanks @hobu !

Note to self: [shell-paths](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#shell-paths) for converting slashes in paths dependent on OS.

ci/conda/cmake.sh
```diff
@@ -1,5 +1,14 @@
+#!/bin/sh
+
+if [ "$RUNNER_OS" == "Windows" ]; then
```
rouault, line 3: It is weird that we have to explicitly specify SWIG_EXECUTABLE. We don't need to do that for GDAL builds

hobu:
> Note to self: [shell-paths](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#shell-paths) for converting slashes in paths dependent on OS.

IMO the Python MapScript install stuff should be refactored to use "modern" Python packaging installation stuff instead of hand copying files around to paths. Some of the items that would entail include:

* A proper `MANIFEST.in` file to describe the examples and data and such that should be installed with Python MapScript
* Use `pyproject.toml` for all of the package metadata (readme, version, etc)
* Proper rpath'ing to get rid of `MAPSERVER_DLL_PATH`  (and the rest of the dirty hackery in `__init__.py`)

ci/conda/cmake.sh
hobu: This could probably be removed. It was in the mapserver feedstock and I carried it along from there.

rouault: > This could probably be removed

indeed: https://github.com/MapServer/MapServer/pull/7109

geographika:
@hobu - there are definitely some improvements that can be made such as a move to a .toml file, but I'm not sure if there is a "clean" way when building with a combination of SWIG, CMake, and Python packaging.

Setting `MAPSERVER_DLL_PATH` is required on Windows to allow the MapServer DLLs to be referenced from MapScript - see https://github.com/MapServer/MapServer/issues/6101.
Including the `mapserver/bin` folder on the system PATH no longer works since Python 3.8. 

GDAL also relies on several path manipulations in https://github.com/rouault/gdal/blob/master/swig/python/osgeo/__init__.py

Copying files with CMake is required to gather the files from the source repository such as tests so they can be alongside the SWIG generated .py file and a .pyd file and added to the wheel.

PR #7106, opened 2024-07-09
Update security policy to mention 8.2 release

jmckenna:
@sdlime am I understanding the plan correctly? (see proposed [SECURITY.md](https://github.com/jmckenna/MapServer/blob/update-security-policy-8.2-release/SECURITY.md) for a better view)

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -11,9 +11,11 @@ submissions, when describing the vulnerability (see https://mapserver.org/develo
 ## Supported Versions
 
 The MapServer PSC (Project Steering Committee) will release patches for security vulnerabilities 
-for the last release branch of the **two most recent release series** (such as 8.x, 7.x, 6.x, etc...). 
+for the last release branch of the **two most recent release series** (such as 8.x, 7.x, 6.x, etc..., 
+where "x" is the most recent release in the series, such as: 7.6.6 being supported, but 
+not 7.6.5). 
 Patches will only be provided **for a period of three years** from the release date of the current series.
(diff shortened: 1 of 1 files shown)


jmckenna:
@sdlime nudge

sdlime:
Being explicit with the 7.6.x date is good. Two things:

1. Do we want to support 8.0.x? If 8.4 comes out do we still support it? I was thinking originally that we’d only support the most recent release in the 2 most recent major releases so 8.0.x goes out of support once 8.2 is released. 8.0 is relevant in terms of dropping 7.6.x.
2. I wonder if we should be explicit and define “x” as the most recent release in a major branch series. So, 7.6.6 is supported but 7.6.5 is not… perhaps that’s obvious though.

jmckenna:
Thanks @sdlime, I agree with your points (to clarify dropping 8.0.x)

sdlime:
@jmckenna, we could say we support 2 most recent minors in the current series if folks want, so 8.2, 8.0 and then 8.4, 8.2 and so forth. Then only latest in last series, so 7.6, for 2 years.

jmckenna:
@sdlime I like your initial thinking (support one latest release in the major series), I just needed to clarify the wording.  I hope it makes more sense now to readers.

sdlime:
I guess we can patch backwards (e.g. 8.0) but not release.

jmckenna:
Exactly my thinking (releasing 8.2.1 and 7.6.7 now...)

PR #7111, opened 2024-07-11
Fix CLUSTER layers to return EXTENT

### LINKED ISSUE #7070: If clustering is used, the capabilities are not completely created
We use clustering in a layer in our mapservice. This works fine. But if we ask for the capabilities through the WMS we get only a part of the capabilities.

We use the following lines in our mapfile in the LAYER section:
  LAYER
    NAME 'xxx'
    STATUS ON
    PROJECTION
      'init=epsg:4326'
    END
    ...
    CLUSTER
      MAXDISTANCE 15
      BUFFER 10
      REGION "ellipse" # This line is responsible for not being able to get the capabilities
    END
    LABELITEM "Cluster_FeatureCount"
    CLASSITEM "Cluster_FeatureCount"
    CLASS
      NAME 'xxx'
      EXPRESSION "1"
      STYLE
        COLOR 240 228 66
        OUTLINECOLOR 0 0 0
        SYMBOL "square"
        MINSIZE 16
        MAXSIZE 120
      END
    END
    CLASS
      NAME 'xxx (geclusterd)'
      EXPRESSION ("[Cluster_FeatureCount]" != "1")
      STYLE
        COLOR 240 228 66
        OUTLINECOLOR 0 0 0
        SYMBOL "square"
        MINSIZE 16
        MAXSIZE 120
      END
      LABEL
        SIZE 8
        COLOR 0 0 0
        ALIGN CENTER
        PRIORITY 10
        BUFFER 1
        PARTIALS TRUE
        POSITION cc
        FORCE TRUE
      END
    END
  END

If we comment out the line REGION “ellipse” we can get the complete capabilities. If the line is not commented out we get the partial capabilities.

For example (please modify !!!): 
I expected the following request to work:
http://localhost:6080/cgi-bin/mapserv.exe?map=/ms4w/apps/local-demo/services.map&version=1.0.9&service=WMS&request=GetCapabilities

The WMS version number is not important. If the line REGION "ellipse" is not commented out the GetCapabilities request for every WMS version goes wrong.

Create a mapfile with a layer based on the layer information above. Use a points layer from an Oracle database (our data is saved in an Oracle database). Use the GetCapabilities request to get the error.
Comment out the line REGION "ellipse" and use again the GetCapabilities request to get the capabilities.

## Solution/Workaround
If I add an extent in the layer the GetCapabilities request works. With of without commenting out the line REGION "ellipse".
So to get everything working I added the line EXTENT -100000 200000 650000 950000 beyond the PROJECTION

## Attach simple test case (drag/drop it here)
- Save the shapefile in an Oracle database
- Change the connection file
- Use the mapfile to get the capabilities: http://localhost:6080/cgi-bin/mapserv.exe?map=/ms4w/apps/local-demo/services.map&version=1.0.9&service=WMS&request=GetCapabilities
- Activate line 68 with the text REGION "ellipse"
- Get the capabilities again: http://localhost:6080/cgi-bin/mapserv.exe?map=/ms4w/apps/local-demo/services.map&version=1.0.9&service=WMS&request=GetCapabilities
[Uploading test-data.zip…]()

## Operating system
Windows
RHEL Linux (running container in podman)

MS4W 4.0.5 with Mapserver 7.7.0-dev
Mapserver 8 build from source for Linux with Oracle support

geographika:
Attempt to fix #7070. 

The root cause is when an EXTENT is not set on a layer, when GetCapabilities is called the code tries to calculate this based on the underlying PostGIS layer. 
The cluster layer defaults to using the `vtable->LayerGetExtent` of the source layer, in this case the `msPostGISLayerGetExtent` function. This calls  `msPostGISParseData where `layer->layerinfo` is cast to `msPostGISLayerInfo`, but it is a struct of type `msClusterLayerInfo`.  

https://github.com/MapServer/MapServer/blob/15eb094b81a89c54813ad2a1476970d7946bb8d2/src/mappostgis.cpp#L1201

When the code tries to access properties that only belong to `msPostGISLayerInfo` the code crashes and the GetCapabilities output stops at the layer extent. 

https://github.com/MapServer/MapServer/blob/15eb094b81a89c54813ad2a1476970d7946bb8d2/src/mappostgis.cpp#L1234-L1237

This pull request adds a new function to the cluster vtable: `vtable->LayerGetExtent = msClusterLayerGetExtent;` which uses the `LayerGetExtent` of the underlying source layer. 

Tests have been added to check this, and the cluster output in general, which was previously untested. I believe this should also fix the issue in layers with other types of connections such as Oracle or SQL Server. 

One outstanding issue is that there are errors when debugging (on both Windows and Linux using vagrant) when the layer is closing connections:

https://github.com/MapServer/MapServer/blob/15eb094b81a89c54813ad2a1476970d7946bb8d2/src/mappool.c#L289-L290

```
Unhandled exception at 0x00007FF9D017C8F9 (ntdll.dll) in mapserv.exe: 0xC0000374: A heap has been corrupted (parameters: 0x00007FF9D01F38B0).
```

This seems to be thrown even if CLUSTER is unused, so could be a separate bug that has only surfaced due to the new tests.


rouault:
> The cluster layer defaults to using the `vtable->LayerGetExtent` of the source layer

aouch!  Did you spot the root cause for this ? This feels very wrong.
Your fix is OK, but there's probably something more fundamental to address.

geographika:
> > The cluster layer defaults to using the `vtable->LayerGetExtent` of the source layer
> 
> aouch! Did you spot the root cause for this ? This feels very wrong. Your fix is OK, but there's probably something more fundamental to address.

Without overriding `vtable->LayerGetExtent` any calls to get the layer extent will use whichever function was associated with the original layer type, however the `layerinfo` is a struct specific to the cluster layer. 
There could be other issues like this e.g. `layer->vtable->LayerGetShapeCount`, but I've not checked WFS services against a cluster layer yet. 

I tried a few other fixes - hardcoding extents for cluster layers, calculating the extent when the layer was opened (slow), but overriding `vtable->LayerGetExtent` seemed the cleanest approach. Maybe it was an oversight during implementation.

sdlime:
Maybe @szekerest could comment?

PR #7115, opened 2024-07-17
Fix style scale factor inheritance

marisn:
Commit ede4953 introduced `SIZEUNITS` inheritance layer->class->style/label. When `LAYER STYLEITEM` is in use, for each feature `msUpdateStyleFromString` is called but no attempts are made to ensure correct scale factor values for newly created style item thus all features get scale factor 1.0, that might be different from value it should have. This wasn't a problem before ede4953 as layer scale factor was always used to render features with `STYLEITEM`.

This commit introduces a workaround by lifting out code from the only place where scale factor is adjusted in a way that it can be reused also in the `STYLEITEM` case.

Note – this workaround does not fit for mapscript/style.i `updateFromString` as I couldn't see how to access `map` and `layer` there.


geographika:
@marisn - are you able to add a test for this change (sample Mapfile and image output) to the msautotest suite?

marisn:
> @marisn - are you able to add a test for this change (sample Mapfile and image output) to the msautotest suite?

Will this be good enough?

jmckenna:
@marisn looks good, thanks for adding a test

PR #7117, opened 2024-07-18
msautotest: completely remove msautotest/misc/sde_*.map.old files

rouault:
There's little value in keeping those files IMHO


jmckenna:
agreed, as the driver was removed completely for the 7.0.0 release

jmckenna:
thanks @rouault !

PR #7118, opened 2024-07-19
tostring() expression function: validate format string, and make sure buffer is large enough

rouault:


rouault:
Manually backport to 8.2 per https://github.com/MapServer/MapServer/pull/7123

PR #7130, opened 2024-08-06
Provide additional names in FindProtobufC.cmake

hobu:
We use different locations for protobufc in the Conda Forge builds, and this patch was needed to support look ups in alternative locations.

### DIFF
--- a/cmake/FindProtobufC.cmake
+++ b/cmake/FindProtobufC.cmake
@@ -90,13 +90,13 @@ find_library(PROTOBUFC_LIBRARY
 mark_as_advanced(PROTOBUFC_LIBRARY)
 
 find_path(PROTOBUFC_INCLUDE_DIR
-							NAMES "google/protobuf-c/protobuf-c.h"
+							NAMES "protobuf-c.h" "google/protobuf-c/protobuf-c.h"
 							PATHS "/usr" "/usr/local" "/opt" ENV PROTOBUFC_ROOTDIR
 							PATH_SUFFIXES "include")
 mark_as_advanced(PROTOBUFC_INCLUDE_DIR)
 
 find_program(PROTOBUFC_COMPILER
-							NAMES "protoc-c"
+							NAMES "protoc" "protoc-c"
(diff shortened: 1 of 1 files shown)


cmake/FindProtobufC.cmake
```diff
@@ -90,17 +90,21 @@ find_library(PROTOBUFC_LIBRARY
...
 mark_as_advanced(PROTOBUFC_INCLUDE_DIR)
 
 find_program(PROTOBUFC_COMPILER
-							NAMES "protoc-c"
+							NAMES "protoc" "protoc-c"
 							PATHS "/usr" "/usr/local" "/opt" ENV PROTOBUFC_ROOTDIR
 							PATH_SUFFIXES "bin")
 mark_as_advanced(PROTOBUFC_COMPILER)
 
+IF (DEFINED PROTOBUF_PROTOC_EXECUTABLE)
```
rouault, line 104: so PROTOBUF_PROTOC_EXECUTABLE is provided by the user ? Is it for an alternative name than protoc or protoc-c, or for probing a different location? If the later, doesn't conda-forge setting of CMAKE_PREFIX_PATH or other CMake variables enough to find it?

hobu: For conda-forge builds, we are cross-compiling on OSX, and we have to explicitly provide the host's `protoc` instead of letting it pick up the build's `protoc`. I needed a way to specify it or the wrong arch gets picked up and it errors.

rouault: As far as I can see, cmake -DPROTOBUFC_COMPILER=/path/to/proto-c should work without this patch, given that the find_program() output variable is a cache variable

jmckenna:
nudge here as this was tagged for the 8.2.2 release (see [Milestone](https://github.com/MapServer/MapServer/milestone/63)) which is set to be released this Friday

geographika:
For info - the Conda patches related to this PR are at: https://github.com/conda-forge/mapserver-feedstock/tree/main/recipe/patches

jmckenna:
thanks @hobu !

PR #7131, opened 2024-08-07
allow explicit override of MapScript output folder via CMake variable

hobu:
The assumptions about where MapScript should be build don't always hold true for every build system. This allows us to explicitly override it via CMake argument to be whatever is necessary.

### DIFF
--- a/src/mapscript/python/CMakeLists.txt
+++ b/src/mapscript/python/CMakeLists.txt
@@ -42,18 +42,20 @@ file(READ ${SETUP_PY_TEMP} SETUP_CONTENT)
 
 file(GENERATE OUTPUT $<TARGET_FILE_DIR:${SWIG_MODULE_pythonmapscript_REAL_NAME}>/setup.py INPUT ${SETUP_PY_TEMP})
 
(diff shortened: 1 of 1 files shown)


rouault:
> The assumptions about where MapScript should be build don't always hold true for every build system.

can you explain why this is necessary / why the current default doesn't work for you ?

geographika:
Seems fine to me to allow this to be configured. `MAPSCRIPT_WORKING_DIR` might be a bit too generic a variable name - this is for Python only. 
Do we have a place where these variables get documented?

hobu:
> can you explain why this is necessary / why the current default doesn't work for you 

conda-forge builds use `CMAKE_ARGS` to override a bunch of stuff, and the `CMAKE_BUILD_TYPE` thing that our windows builds do gets changed and causes the WORKING_DIRECTORY to be invalid for these custom commands.

More generally, I think it's probably time to refactor the Python MapScript builds to use scikit-build-core. The [PDAL Python bindings](https://github.com/PDAL/python/blob/main/pyproject.toml) could be inspirational on that topic.

PR #7135, opened 2024-08-14
Add CONNECTIONTYPE RASTERLABEL (RFC 139)

rouault:
Candidate implementation of [RFC 139](https://github.com/MapServer/MapServer-documentation/pull/960)

This adds a new CONNECTIONTYPE to be able to display labels based on raster content.

```
  LAYER
    PROJECTION
      "init=epsg:4326"
    END
    NAME "test"
    TYPE POINT
    CONNECTIONTYPE RASTERLABEL
    STATUS DEFAULT
    DATA data/temperature.tif
    # PROCESSING "BANDS=1"
    PROCESSING "LABEL_SPACING=40"
    # PROCESSING "RESAMPLE=BILINEAR"
    CLASS
        TEXT (tostring([value],"%.1f")+"°")
        LABEL
          TYPE TRUETYPE
          SIZE 7
        END # label
    END # class
  END
```

gives:

rasterlabel

The code is similar to UVRASTER.


jmckenna:
@rouault testing this PR with MS4W5 on Windows, some feedback:
- if I set an incorrect data path such as `DATA ttt/temperature.tif`, and with debugging on, I would have hoped to see a message informing me that the raster data path cannot be found, or something similar, instead I get:
  `map2img -m rasterlabel.map -o ttt.png -all_debug 5`
  ```msLoadMap(): 0.005s
     msDrawMap(): rendering using outputformat named png (AGG/PNG).
     msDrawMap(): WMS/WFS set-up and query, 0.000s
     Entering msRasterLabelLayerWhichShapes().
     msDrawMap(): Image handling error. Failed to draw layer named 'test'.
     msDrawMap(): Image handling error. Failed to draw layer named 'test'. <br>
     msFreeMap(): freeing map at 00C01CE8.
     freeLayer(): freeing layer at 02C29CC8.
  ```
Maybe this is not related to this PR, but instead would be a useful enhancement (?)

rouault:
@jmckenna Thanks for the feedback and reports. I've taken into account your 2 points.

1) Regarding wrong path in DATA, you will now get
```
$ map2img -conf ../etc/mapserv.conf -m rasterlabel.map -i png -o result/rasterlabel.png 
msDrawMap(): Image handling error. Failed to draw layer named 'test'. <br>
msRasterLabelLayerWhichShapes(): General error message. Cannot open DATA of layer test <br>

$ map2img -conf ../etc/mapserv.conf -all_debug 5 -m rasterlabel.map -i png -o result/rasterlabel.png 
msLoadMap(): 0.007s
msDrawMap(): rendering using outputformat named png (AGG/PNG).
msDrawMap(): WMS/WFS set-up and query, 0.000s
Entering msRasterLabelLayerWhichShapes().
msRasterLabelLayerWhichShapes(): cannot open DATA data/xtemperature.tif: msRasterLabelLayerWhichShapes(): General error message. Cannot open DATA of layer test
msDrawMap(): Image handling error. Failed to draw layer named 'test'.
msDrawMap(): Image handling error. Failed to draw layer named 'test'. <br>
msRasterLabelLayerWhichShapes(): General error message. Cannot open DATA of layer test <br>
msFreeMap(): freeing map at 0x561fde3661f0.
freeLayer(): freeing layer at 0x561fde498de0.
```

2) I've added a test case for the SLD fix (was initially hesitant to add one as the fix isn't perfect, but this is indeed worth testing, and I've added a comment near the test case to mention the current output isn't perfect). That commit is material for backport as it is not tied with this PR but was found while testing SLD support with rasterlabel layers following a question from @geographika

rouault:
hum I had to disable the new test because there's a memory leak during the call to yyparse() in msSLDGenerateTextSLD(). Despite my investigation, I can't manage to fix it. That said, the current fix is probably still better than crashing

jmckenna:
@rouault thanks, great improvement on the case of wrong path in DATA, tested here locally.  thanks!

PR #7121, opened 2024-07-20
msProcessProjection(): strip epsgaxis= MapServer specific parameter that confuses PROJ < 9.5.0

rouault:
Fixes #7019

Supersedes PR #7074

PR #7141, opened 2024-08-22
fix stripping epsgaxis= before importing OSR projection

### LINKED ISSUE #6966: no crs output in geojson when reprojecting to epsg with inverted axes

Expected a geojson output including a `crs` key. E.g.

```json
{
"type": "FeatureCollection",
"name": "Something",
"crs": { "type": "name", "properties": { "name": "urn:ogc:def:crs:EPSG::3035" } },
"features": [
{ "type": "Feature", "id": "something", "properties": {}, "bbox": [ 1,2,3,4 ], "geometry": { "type": "Polygon", "coordinates": [ [ [ ] ] ] } }
],
"bbox": [ 1,2,3,4 ]
}
```

Actually received that without the `crs` key.

Have a source in another srs than the request.

Ensure geojson `RFC7946=OFF` (default) 

Make a WFS request with an output srs that has "inverted axes" (not x/y but y/x cq lat/lon cq north/east). E.g. EPSG:3035 or EPSG:4326. (Not using 4326 cq WGS84 in the example as to not confuse the `crs` being hidden for other reasons.)

`QUERY_STRING=service=WFS&request=GetFeature&count=1&version=2.0.0&outputFormat=geojson&typeName=test&srsName=EPSG:3035`

## Analysis

When the [output projection is made into a OGC WKT](https://github.com/MapServer/MapServer/blob/rel-8-0-1/mapogroutput.cpp#L964) and it has "inverted axes", then [`+epsgaxis=ne` is added to the proj string](https://github.com/MapServer/MapServer/blob/rel-8-0-1/mapfile.c#L1271) which is inputted to GDAL's [`importFromProj4`](https://github.com/OSGeo/gdal/blob/master/ogr/ogrspatialreference.cpp#L11012). Resulting in a WKT named "unknown" and without authority name and code.

When that WKT is turned into an `OGRSpatialReference` and later on passed to GDAL to write the geojson, it has no URN (because no authority) and the `crs` key [is skipped](https://github.com/OSGeo/gdal/blob/release/3.6/ogr/ogrsf_frmts/geojson/ogrgeojsondatasource.cpp#L421).

The coordinate axes are swapped to x,y though, but why would I want that? edit1: ~~I don't see that requirement in the [(old) geojson spec](https://geojson.org/geojson-spec.html). I would expect the axes order to follow the epsg.~~ edit2: O, because geojson requires that:
> The order of elements must follow x, y, z order (easting, northing, altitude for coordinates in a projected coordinate reference system, or longitude, latitude, altitude for coordinates in a geographic coordinate reference system).

## Attached simple test case

[wfs-test.zip](https://github.com/MapServer/MapServer/files/13319528/wfs-test.zip)

```
# commands.txt
mapserv -nh "QUERY_STRING=map=wfs-test.map&service=WFS&version=2.0.0&request=GetFeature&typeName=test&outputFormat=geojson&srsName=EPSG:3035" # output srs axis x,y (y,x according to epsg, but x,y cause of geojson), no crs key though
mapserv -nh "QUERY_STRING=map=wfs-test.map&service=WFS&version=2.0.0&request=GetFeature&typeName=test&outputFormat=geojson&srsName=EPSG:3857" # output srs axis x,y (correct according to epsg), yes crs key
mapserv -nh "QUERY_STRING=map=wfs-test.map&service=WFS&version=2.0.0&request=GetFeature&typeName=test&outputFormat=gml3&srsName=EPSG:3035" # output srs axis y,x (correct according to epsg), gml3, yes srsName
```

## Operating system

ubuntu 20.04

mapserver 8.0.1 (from source)
proj 9.3 (from source)
gdal 3.6.2 (from source, also with proj 9.3)

roelarents:
like #7121
fixes #6966


src/mapgdal.cpp
```diff
@@ -634,7 +634,8 @@ char *msProjectionObj2OGCWKT(projectionObj *projection)
   /* -------------------------------------------------------------------- */
   /*      Look for an EPSG-like projection argument                       */
   /* -------------------------------------------------------------------- */
-  if (projection->numargs == 1 &&
+  if ((projection->numargs == 1 ||
+      (projection->numargs == 2 && strstr(projection->args[1], "epsgaxis=") != NULL)) &&
```
rouault, line 638:
```suggestion
      (projection->numargs == 2 &&
       strstr(projection->args[1], "epsgaxis=") != NULL)) &&
```

src/mapgdal.cpp
```diff
@@ -634,7 +634,9 @@ char *msProjectionObj2OGCWKT(projectionObj *projection)
   /* -------------------------------------------------------------------- */
   /*      Look for an EPSG-like projection argument                       */
   /* -------------------------------------------------------------------- */
-  if (projection->numargs == 1 &&
+  if ((projection->numargs == 1 ||
+      (projection->numargs == 2 &&
+       strstr(projection->args[1], "epsgaxis=") != NULL)) &&
```
rouault, line 639:
```suggestion
       (projection->numargs == 2 &&
        strstr(projection->args[1], "epsgaxis=") != NULL)) &&
```

PR #7159, opened 2024-09-12
Set layer extent to requested wms bbox

PatrikSylve:
When constructing the WMS URL for a layer with a WMS connection, the layerObj.projection is assigned the requested SRS if allowed. However, in some cases, reprojecting the layer's extent into the requested SRS can result in invalid values (e.g., when the layer covers the entire world in EPSG:4326, and the requested SRS is a UTM zone), causing the layer to fail the msLayerIsVisible check and not render.  To avoid this issue, the WMS bounding box extent is assigned directly rather than attempting to reproject the layer extent.


jratike80:
Does it mean that after this change the LAYER-EXTENT and wms_extent won't have any meaning, but the BBOX is always accepted and Mapserver tries to render the WMS layer, even when the BBOX really is outside the extent?

PatrikSylve:
> Does it mean that after this change the LAYER-EXTENT and wms_extent won't have any meaning, but the BBOX is always accepted and Mapserver tries to render the WMS layer, even when the BBOX really is outside the extent?

At this point in the code it has already been asserted that the layer is visible (https://github.com/MapServer/MapServer/blob/main/src/mapdraw.c#L402), so it should not render it if the original layer extent is outside. When mapserver mapserver performs the WMS request the layer is treated as a Raster, so in that context the layer-extent has meaning.

jmckenna:
@PatrikSylve can you include an [msautotest ](https://mapserver.org/development/tests/autotest.html) test for the scenario that you mention here, as part of this pull request?  thanks.

PatrikSylve:
@jmckenna 
I updated wms_client test and added that the wms client layer supports  ESPG:32663. Since the remote https://demo.mapserver.org does not support this CRS the test results in an 'Invalid CRS'-exception (which is expected). Without this fix the response is just an empty image since the layer was evaluated to be outside map extent. 

For a cleaner test, would it be possible to add EPSG:32663 as an accepted SRS for a service on demo.mapserver?

jmckenna:
@PatrikSylve I've added EPSG:32633 to that service (see [GetCapabilities](https://demo.mapserver.org/cgi-bin/wms?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetCapabilities)).  Can you quickly try this again?  (I'd like to include your change in today's MapServer 8.4.0 beta release)  Please let me know.

PatrikSylve:
Thanks! 

My previous example request pointed to no data so I updated the BBOX in test to get a better output. I tested this request with a build of the previous version of the code as well, and that one resulted in an empty image. So looks like it works as expected :+1:

jmckenna:
Cool, thanks for the fast changes @PatrikSylve .  Merging!  (willdo the beta release tomorrow morning)

PR #7089, opened 2024-06-22
msProjectionInheritContextFrom(): do not reuse source PROJ context if…

rouault:
… called from different thread

Fixes #7079

PR #7161, opened 2024-09-18
Add a per-thread PROJ context pool

rouault:
Should improve performance issues of #7089


rouault:
Put as draft awaiting for @szekerest feedback

rouault:
Closing as https://github.com/MapServer/MapServer/pull/7162 does the job in a simpler way

PR #7162, opened 2024-09-18
msProjectionContextGetFromPool(): set thread_id to current thread

rouault:
Should improve performance issues of https://github.com/MapServer/MapServer/pull/7089


rouault:
Put as draft awaiting for @szekerest feedback

szekerest:
@rouault Got a feedback from the client that the fix solved the performance problem. We can go ahead with this.

Issue #5894, opened 2019-10-12
Comments with brackets can cause EXPRESSION errors


geographika:
If there is a comment on the same line as an expression and it contains a bracket then the expression parsing breaks. For example:

```
  LAYER
    NAME 'expressions_test099'
    FILTER ('[cty_name]' = 'Aitkin') # Extra comment and bracket )
    INCLUDE 'include/bdry_counpy2_shapefile.map'
    METADATA
        "gml_include_items" "cty_name,cty_fips"
    END
  END
```

Causes:
```
msQueryByRect(): Search returned no results. No matching record(s) found.
msEvalExpression: Expression parser error. Failed to parse expression: &#39;[cty_name]&#39; = &#39;Aitkin&#39;) # Extra comment and bracket 
yyparse(): Expression parser error. syntax error
```

The parser seems to accept all input up to the final bracket in the line. If there is a bracket in a comment with text after there is a different error:

```
  LAYER
    NAME 'expressions_test099'
    FILTER ('[cty_name]' = 'Aitkin') # Extra comment and) bracket
    INCLUDE 'include/bdry_counpy2_shapefile.map'
    METADATA
        "gml_include_items" "cty_name,cty_fips"
    END
  END
```

Causes:

`loadLayer(): Unknown identifier. Parsing error near (bracket):(line 80)`

PR #7178, opened 2024-10-22
SLD Filters missing due to logicial keywords in expressions

geographika:
When using WMS GetStyles requests to generate SLD I noticed some filters were missing. On debugging it became apparent that any string containing a logical operator was failing to generate a filter:

```
EXPRESSION ("[Property1]" = "Not Applicable")
```

Was missing the expected filter:

```xml
<ogc:Filter><ogc:PropertyIsEqualTo><ogc:PropertyName>Property1</ogc:PropertyName><ogc:Literal>Not Applicable</ogc:Literal></ogc:PropertyIsEqualTo></ogc:Filter>
```

`msSLDNumberOfLogicalOperators` had a simple search for `AND`, `NOT`, and `OR` that was finding these words in strings causing errors. This fix attempts to tokenize the string, leaving quoted strings intact to avoid this. It is by no means perfect, and I also had to change previous test cases to have spaces around keywords e.g. `(([FID] > 1)And([FID] <= 10))` so that they pass. 

MapServer seems to be missing a parsing engine for expressions (or at least I can't find anything in the codebase). Without this any SLD expression building is going to be flaky. See also #5894.

### DIFF
--- a/src/mapogcsld.cpp
+++ b/src/mapogcsld.cpp
@@ -4815,27 +4815,18 @@ static int msSLDNumberOfLogicalOperators(const char *pszExpression) {
       continue;
     }
 
-    if (strncasecmp(papszArgs[i], "AND", 3) == 0) {
+    if ((strncasecmp(papszArgs[i], "AND", 3) == 0) ||
+        (strcasestr(papszArgs[i], "AND("))) {
       nAndCount += 1;
     }
 
-    if (strcasestr(papszArgs[i], "AND(")) {
-      nAndCount += 1;
-    }
-
-    if (strncasecmp(papszArgs[i], "OR", 2) == 0) {
+    if ((strncasecmp(papszArgs[i], "OR", 3) == 0) ||
+        (strcasestr(papszArgs[i], "OR("))) {
       nOrCount += 1;
     }
 
-    if (strcasestr(papszArgs[i], "OR(")) {
-      nOrCount += 1;
-    }
-
-    if (strncasecmp(papszArgs[i], "NOT", 3) == 0) {
-      nNotCount += 1;
-    }
-
-    if (strcasestr(papszArgs[i], "NOT(")) {
+    if ((strncasecmp(papszArgs[i], "NOT", 3) == 0) ||
+        (strcasestr(papszArgs[i], "NOT("))) {
       nNotCount += 1;
     }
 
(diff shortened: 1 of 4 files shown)


geographika:
CI currently failing due to #7179

src/mapogcsld.cpp
```diff
@@ -4793,49 +4793,54 @@ static char *msSLDGetLeftExpressionOfOperator(const char *pszExpression) {
...
-  if (!pszNot) {
-    pszNot = strcasestr(pszExpression, "NOT(");
-  }
+    if (strncasecmp(papszArgs[i], "AND", 3) == 0) {
+      nAndCount += 1;
+    }
 
-  if (!pszAnd && !pszOr && !pszNot)
-    return 0;
+    if (strncasecmp(papszArgs[i], "OR", 3) == 0) {
```
rouault, line 4822:
```suggestion
    if (strncasecmp(papszArgs[i], "OR", 2) == 0) {
```

rouault:
This indeed feels like band-aid fixing, but better than what we have, so why not. Perhaps a TODO/FIXME comment in the code itself would be welcome. msSLDGetLogicalOperator() would also benefit from a similar logic.
Ideally we'd perhaps need to use yylex() and a new SLD parsing mode in mapparser.y

geographika:
> This indeed feels like band-aid fixing, but better than what we have, so why not. Perhaps a TODO/FIXME comment in the code itself would be welcome. msSLDGetLogicalOperator() would also benefit from a similar logic. Ideally we'd perhaps need to use yylex() and a new SLD parsing mode in mapparser.y

Yes, these changes are very much trying to make an error-prone approach slightly less buggy. With further tests it appears both `msSLDGetLogicalOperator`, `msSLDGetRightExpressionOfOperator`, and `msSLDGetLeftExpressionOfOperator` have the same issues - see the test where `"An And value"` is converted to `"An"`. 

@sdlime - could any of the changes in https://mapserver.org/development/rfc/ms-rfc-64.html be used here instead of the current string searching approach?

I've a copy of the Lex Yacc book but never got much past the first chapter..

geographika:
I've updated the function so the previous cases such as `AND(` and `OR(` still pass, so anything that previously worked should continue to do so. I keep thinking that expression parsing is implemented somewhere - `msEvalTextExpressionInternal` seems close but returns a true/false rather than a parse tree that could be used for converting expressions from Mapfile to SLD.

Issue #6950, opened 2023-10-19
Expressions containing <> do not throw parsing errors


geographika:
I ran into this issue when checking SLD output from LAYER CLASS EXPRESSIONS. 

I'd not thoroughly read the [logical expression docs](https://www.mapserver.org/mapfile/expressions.html#arithmetic-expressions-that-return-a-logical-value) and so had been unaware that `NOT` expressions in MapServer can be represented by `!=` or `ne`, but not `<>`. 

The following layer parses fine in MapServer, but the `<>` gets parsed as `MS_TOKEN_COMPARISON_LT` and `MS_TOKEN_COMPARISON_GT` in https://github.com/MapServer/MapServer/blob/24d36f842c243f96eba88ff896db1bf34d8a687a/src/maplayer.c#L679

```
    LAYER
      NAME "Test18"    
      TYPE POINT
      CLASS
        EXPRESSION ( [FID] <> 1 AND [FID] <> 10 )
        STYLE
            COLOR 220 0 0
            WIDTH 1
        END
      END
    END
```

This probably causes strange issues when rendering data, but is more noticeable when generating SLD as it returns:

```xml
  <ogc:Filter>
    <ogc:And>
      <ogc:PropertyIsLessThan>
        <ogc:PropertyName>FID</ogc:PropertyName>
        <ogc:Literal>> 1</ogc:Literal>
      </ogc:PropertyIsLessThan>
      <ogc:PropertyIsLessThan>
        <ogc:PropertyName>FID</ogc:PropertyName>
        <ogc:Literal>> 10 </ogc:Literal>
      </ogc:PropertyIsLessThan>
    </ogc:And>
  </ogc:Filter>
```

Ideally I think the MapServer expression parser should throw an error when reading the Mapfile (or we attempt to support `<>` in expressions).

sdlime:
Maybe easiest to support <>…

PR #7182, opened 2024-11-02
Allow <> to be used as a logical operator in expressions

geographika:
This pull request allows `<>` to be used for "not equal to" logical expressions, along with the already supported `ne` and `!=`. Fix for #6950. 

When creating SLD filters expressions are created by checking values in strings, so modifications have been added to `mapogcsld.cpp` have been made to support these. There are still issues around this approach for creating SLD filters as noted in #7178. 

`<>` has been added to `maplexer.l` which causes `mapparser.c` to be regenerated. This was created using the MapServer vagrant image. Note the version of Bison has changed from:

```
/* A Bison parser, made by GNU Bison 3.0.4.  */
/* A Bison parser, made by GNU Bison 3.5.1.  */
```

However I've not checked in `mapparser.c` and `mapparser.h` as part of this pull request (it causes cppcheck errors).


geographika:
The newly created `mapparser.c` is causing issues with Cppcheck. 

```
 Cppcheck 1.90
Running cppcheck on /home/runner/work/mapserver/mapserver/./scripts/../src/. ... (can be long)
src/mapparser.y:974,error,memleak,Memory leak: yyval.strval
memleak check failed
src/mapparser.c:1390,warning,uselessAssignmentPtrArg,Assignment of function parameter has no effect outside the function. Did you forget dereferencing it?
uselessAssignmentPtrArg check failed
src/mapparser.c:1390,style,unreadVariable,Variable 'yymsg' is assigned a value that is never used.
unreadVariable check failed
```

geographika:
There are also modifications to logic in `mapogcsld.cpp`:

```
      if ((pszExpression[i] != szCompare2[0] ||
           pszExpression[i] != toupper(szCompare2[0])
...
```

Changed to the code below (`||` to `&&`), as I think this should be checking if a character doesn't match lower and upper case of the comparison. However as earlier in the code only values such as `ne` `eq` are checked for in lower case I don't think the upper case has any effect anyway. 


```
      if ((pszExpression[i] != szCompare2[0] &&
         (pszExpression[i] != toupper(szCompare2[0]
...
```

src/mapogcsld.cpp
```diff
@@ -4963,19 +4970,23 @@ static char *msSLDGetAttributeNameOrValue(const char *pszExpression,
...
 
       {
         pszAttributeName[iValue++] = pszExpression[i];
       } else {
         if (((pszExpression[i + 1] == szCompare[1] ||
               pszExpression[i + 1] == toupper(szCompare[1])) ||
              (pszExpression[i + 1] == szCompare2[1] ||
-              pszExpression[i + 1] == toupper(szCompare2[1]))) &&
-            (pszExpression[i + 2] == ' ')) {
+              pszExpression[i + 1] == toupper(szCompare2[1])) ||
```
rouault, line 4986: This (existing) logic seems flowed to me. It would possibly match a string that would be szCompare[0], szCompare2[1], szCompare3[2] (or similar permutations). So for example it could match "!>" (a mix of "!=" and "<>") Once we have selected which [0] character of szCompare/szCompare2/szCompare3, we should keep it for following comparisons.

geographika:
> Is this ready to be merged?

No - I still need to add in the suggested review changes.  Would be easy in Python!

geographika:
Debugging this code further revealed several other discrepancies in the generation of SLD Filters. 

`msSLDGetComparisonValue` doesn't check for whitepace around operators, e.g. `strstr(pszExpression, "!=")` will return `PropertyIsNotEqualTo`, but then the previous approach in `msSLDGetAttributeNameOrValue` relied on whitespace to extract an attribute value:

```c
if (((pszExpression[i + 1] == szCompare[1] ||
      pszExpression[i + 1] == toupper(szCompare[1])) ||
     (pszExpression[i + 1] == szCompare2[1] ||
      pszExpression[i + 1] == toupper(szCompare2[1]))) &&
    (pszExpression[i + 2] == ' ')) 
```

I've updated `msSLDGetAttributeNameOrValue` so whitespace isn't required to generate SLD filters. The alternative would be to require whitespace around all operators in all expressions (at least for SLD output), but this would likely break existing Mapfiles. 

Invalid operators will still be matched in `msSLDGetComparisonValue` e.g. `>!` will match `>` and  `=!` will match `=`. Filters will then use the remaining string as part of the attribute name or value, but I think in the absence of a proper expression parser, it is the responsibility of the Mapfile author to spot this. I've added tests to at least document this issue. 

@rouault - hopefully this is a slight improvement on the current SLD generation, albeit not ideal.

src/mapogcsld.cpp
```diff
@@ -4962,25 +4971,18 @@ static char *msSLDGetAttributeNameOrValue(const char *pszExpression,
...
-          (pszExpression[i] != szCompare2[0] ||
-           pszExpression[i] != toupper(szCompare2[0])))
-
-      {
-        pszAttributeName[iValue++] = pszExpression[i];
+    for (int i = 0; i < nLength; i++) {
+      // Check for comparison operator
+      if ((i + 1 < nLength && strncmp(&pszExpression[i], szCompare, 2) == 0) ||
+          (i + 1 < nLength && strncmp(&pszExpression[i], szCompare2, 2) == 0) ||
+          (i + 1 < nLength && strncmp(&pszExpression[i], szCompare3, 2) == 0)) {
```
rouault, line 4978: to preserve case insensitive comparisons:
```suggestion
      if ((i + 1 < nLength && EQUALN(&pszExpression[i], szCompare, 2)) ||
          (i + 1 < nLength && EQUALN(&pszExpression[i], szCompare2, 2)) ||
          (i + 1 < nLength && EQUALN(&pszExpression[i], szCompare3, 2))) {
```

PR #7184, opened 2024-11-10
Make interpolation and countour compatible of GDAL 3.10

rouault:
GDAL 3.10 has deprecated the usage of the MEM::: open syntax

The workaround without this fix is to use:

MAP
   CONFIG "GDAL_MEM_ENABLE_OPEN" "YES"
   [ ... layer definitions ... ]
END

### DIFF
--- a/src/interpolation.c
+++ b/src/interpolation.c
@@ -257,22 +257,38 @@ int msInterpolationDataset(mapObj *map, imageObj *image,
   free(values);
   free(xyz_values);
 
+  GDALDriverH hMemDRV = GDALGetDriverByName("MEM");
+  if (!hMemDRV) {
+    msSetError(MS_IOERR, "GDAL MEM driver not available",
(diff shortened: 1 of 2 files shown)


PostholerCom:
Also, earlier, before your map file fix:
 
CONFIG "GDAL_MEM_ENABLE_OPEN" "YES" 

 I built 3.10 from source with: 

-DGDAL_MEM_ENABLE_OPEN=ON

Rebulit mapserver 8.2.2 (probably didn't need to do that) and contours still didn't work. So, there may be something on the GDAL side, too.

rouault:
> I built 3.10 from source with:
> 
> -DGDAL_MEM_ENABLE_OPEN=ON

This isn't a CMake option, but a C++ define. That should be passed through ``-DCMAKE_CXX_FLAGS="-DGDAL_MEM_ENABLE_OPEN"``

PR #7186, opened 2024-11-12
mapuvraster: fix hDS file descriptor leak in msUVRASTERLayerGetExtent

gogglesguy:
mapuvraster is very leaky in 7.6.x and beyond. 7.4.x series is fine.


rouault:
good catch. thanks

Issue #7189, opened 2024-11-26
WMS Capabilities: inefficient calculation of BoundingBox for supported CRS


enatdvmv:
With **MapServer 8**, the calculation of the bounding boxes for supported CRS has become quite inefficient (WMS GetCapabilities Request).

We supported a total of 60 EPSGs in MapServer 7. With MapServer 8, the response takes 4 to 5 seconds, much too long.

We have temporarily reduced the supported EPSGs to 20 [Sample Request](https://www.geodaten-mv.de/dienste/adv_dop). But the customer wants more and he wants bbox for all suported EPSGs in the wms capabilities (_WMS_BBOX_EXTENDED "true"_).

Tested on Debian bookworm with MapServer 8.0.0-3 and Debian trixie with MapServer 8.2.2-1.

jmckenna:
@enatdvmv are you already setting `ows_extent` at the layer-level?

enatdvmv:
@jmckenna yes, the parameter WMS_EXTENT is set at web and layer level. It makes no difference whether it is set or not. Why does the request cost resources and time with MapServer 8?

geographika:
It could be related to switching to a new version of PROJ and having to open the proj.db file multiple times. @enatdvmv - has your PROJ version also changed compared to your previous setup?

jratike80:
> Sorry, you can close the issue.

But it is still an issue. Maybe it is possible to deal with the new Proj faster somehow.

geographika:
@enatdvmv - I presume you are using `wms_bbox_extended` in your Mapfile. Could you share your MAP METADATA section?

It looks like the reprojection will be made 60 times for each layer if you have 60 EPSG codes:

https://github.com/MapServer/MapServer/blob/3db4c6b1b0a6eb72e72824334a3b351ce78529e7/src/mapows.c#L2482-L2492

This code has already had some optimisation in https://github.com/MapServer/MapServer/commit/d2dda2ce62269ca661577a484929b16af9539e6f by @rouault - `PROJ6: improve performance of msOWSPrintBoundingBox() with context sharing` so not sure if it can get faster (without a large refactor of code). 

A couple of options would be to move `wms_bbox_extended` to only the layers that require it, or if your Mapfile is static you could cache the GetCapabilities requests.

rouault:
Current state of things. The following GDAL Python script with PROJ master / 9.5.1 runs in 400 ms on my Linux system
```python
from osgeo import osr

[...]

PR #7190, opened 2024-11-28
msProjectRect(): cache whether source projection is polar when repeatdly projecting from it

rouault:
Helps marginally on the belows scenario of
https://github.com/MapServer/MapServer/issues/7189 where the execution time goes from ~440 ms to ~400 ms

```c++
#include "mapserver.h"
#include <stdio.h>
#include <string>

int main() {
    projectionObj src;
    msInitProjection(&src);
    msLoadProjectionStringEPSG(&src, "EPSG:25833");
    const int epsg_list[] = { 25832,
            3034,
            3035,
            3044,
            3045,
            31468,
            31469,
            3857,
            4258,
            4326,
            4647,
            5650,
            5652,
            5674,
            5675,
            5677,
            5678,
            5679,
            4937,
            7409,
            7423,
            9422,
            4979 };
    double minx=200000;
    double miny=5.886e+06;
    double maxx=465000;
    double maxy=6.075e+06;
    for( int epsg: epsg_list )
    {
        projectionObj dst;
        msInitProjection(&dst);
        msProjectionInheritContextFrom(&dst, &src);
        msLoadProjectionStringEPSG(&dst, (std::string("EPSG:") + std::to_string(epsg)).c_str());
        rectObj ext;
        ext.minx = minx;
        msProjectRect(&src, &dst, &ext);
        msFreeProjection(&dst);
    }
    msFreeProjection(&src);
    return 0;
}
```

### DIFF
--- a/src/mapproject.c
+++ b/src/mapproject.c
@@ -490,6 +490,7 @@ int msProjectTransformPoints(reprojectionObj *reprojector, int npoints,
 */
 int msInitProjection(projectionObj *p) {
   p->gt.need_geotransform = MS_FALSE;
+  p->is_polar = -1;
   p->numargs = 0;
   p->args = NULL;
   p->wellknownprojection = wkp_none;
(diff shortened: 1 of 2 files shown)


jmckenna:
thanks for this @rouault

PR #7191, opened 2024-12-04
Add HTTP status code for WMS requests, if MS_WMS_ERROR_STATUS_CODE config option/env variable are set to ON

rouault:
Currently when one issues an HTTP request that ends up in error, MapServer doesn't emit a HTTP Status header response. Hence the webserver will default to generate a 200 OK one.

This is an historical behavior used by other WMS server implementations (GeoServer, Degree) due to the WMS specification being unclear on what do do. Given that WMS is in a retirement phase, it would be risky to change that behavior by default.

That said there are WMS clients that would enjoy having a proper HTTP 400 or 500 status code. Hence if the MS_WMS_ERROR_STATUS_CODE environment variable / config option is set to ON (typically in the ENV section of the mapserver configuration file), MapServer will emit those status codes.

This relies on code to call the msSetErrorWithStatus() new function. A pass has been done in mapwms.cpp to call it. And a more moderate one in a few places of mapfile.c and mapservutil.c related to generic mapfile handling.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,10 +29,10 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 message(STATUS "Requiring C++${CMAKE_CXX_STANDARD} - done")
 
-# Set C99 version
+# Set C version
 # Make CMAKE_C_STANDARD available as cache option overridable by user
-set(CMAKE_C_STANDARD 99
-  CACHE STRING "C standard version to use (default is 99)")
+set(CMAKE_C_STANDARD 11
+  CACHE STRING "C standard version to use (default is 11)")
 message(STATUS "Requiring C${CMAKE_C_STANDARD}")
 set(CMAKE_C_STANDARD_REQUIRED ON)
 set(CMAKE_C_EXTENSIONS OFF)
--- a/msautotest/wxs/expected/wms_invalid_request_with_400_status.txt
+++ b/msautotest/wxs/expected/wms_invalid_request_with_400_status.txt
@@ -1,0 +1,9 @@
+Status: 400 Bad Request
+Content-Type: text/xml; charset=UTF-8
+
+<?xml version='1.0' encoding="UTF-8" standalone="no" ?>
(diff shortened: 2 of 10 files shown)


jratike80:
What if the request was made with `&exceptions=application/vnd.ogc.se_inimage` or `&application/application/vnd.ogc.se_blank`?

rouault:
> What if the request was made with `&exceptions=application/vnd.ogc.se_inimage` or `&application/application/vnd.ogc.se_blank`?

Current state of code is that Status 400 is emitted and an image returned. I guess this is logical

geographika:
Nice addition. One general question/comment - do we have criteria for when a setting is a `CONFIG` option (as in this case), and when it is part of `WEB METADATA`? For example, `wms_allow_getmap_without_styles` is a similar option for WMS but is part of `WEB METADATA`. Maybe the error codes are more of a server setting, whereas `wms_allow_getmap_without_styles` is at the `MAP` level?

Some other notes (for reference):

I had a look at how OWSLib handles failing WMS requests - looks like the status code is first checked and then the XML. 

```python
    if req.status_code in [400, 401]:
        raise ServiceException(req.text)

    if req.status_code in [404, 500, 502, 503, 504]:    # add more if needed
        req.raise_for_status()

    # check for service exceptions without the http header set
    if 'Content-Type' in req.headers and \
            req.headers['Content-Type'] in ['text/xml', 'application/xml', 'application/vnd.ogc.se_xml']:
        # just in case 400 headers were not set, going to have to read the xml to see if it's an exception report.
        se_tree = etree.fromstring(req.content)

        # to handle the variety of namespaces and terms across services
        # and versions, especially for "legacy" responses like WMS 1.3.0
        possible_errors = [
            '{http://www.opengis.net/ows}Exception',
            '{http://www.opengis.net/ows/1.1}Exception',
            '{http://www.opengis.net/ogc}ServiceException',
            'ServiceException'
        ]

        for possible_error in possible_errors:
            serviceException = se_tree.find(possible_error)
            if serviceException is not None:
                # and we need to deal with some message nesting
                raise ServiceException('\n'.join([t.strip() for t in serviceException.itertext() if t.strip()]))

```

I didn't know you could set the config options on the command line in the msautotests  - very handy! `RUN_PARMS: wms_invalid_request_with_400_status.txt MS_WMS_ERROR_STATUS_CODE=ON [MAPSERV] QUERY_STRING="map=[MAPFILE]&SERVICE=WMS`

geographika (approved):

rouault:
@jmckenna Is that something that can be merged now or should we wait after the 8.4.0 release?

jmckenna:
Thanks @rouault I believe we should merge this now, so I draft a release this afternoon.

Issue #5994, opened 2020-02-20
Allow wfs_use_default_extent_for_getfeature for MSSQL Driver 


geographika:
I've been looking to optimise WFS 2.0 MapServer queries using the MSSQL driver and one change would greatly improve query speeds would be to disable the default spatial query. This probably applies to other database drivers such as Postgres and Oracle. 

The `"wfs_use_default_extent_for_getfeature"` was implemented with https://github.com/mapserver/mapserver/commit/45037eb0601e2f48ae548e2cd28b29b429644870 for OGR data sources as documented at https://mapserver.org/el/ogc/wfs_server.html#layer-object

If no spatial filter is passed to a WFS GetFeature the bbox of the map is added to the WHERE clause in SQL Server:

`GEOM.STIntersects(geometry::STGeomFromText('POLYGON((-826053.00001 7143520.99999,-719272.99999 7143520.99999,-719272.99999 7257149.00001,-826053.00001 7257149.00001,-826053.00001 7143520.99999))',3857)) = 1 ) `

This can greatly slow down the query, and as the bbox is set to the extent of all data in the MAP serves no purpose (in this case). 

In addition, if `"wfs_compute_number_matched"` is set to `"true" ` to get the feature count the default spatial bbox is added to get the count, which has a further impact on performance. Queries such as the following are created:

`SELECT count(*) FROM (select * from cities) as foo WHERE ogr_geometry.MakeValid().STIntersects(geometry::STGeomFromText('POLYGON((-826053.00001 7143520.99999,-719272.99999 7143520.99999,-719272.99999 7257149.00001,-826053.00001 7257149.00001,-826053.00001 7143520.99999))',3857)) = 1 ) `

This SQL is generated in mapmssql2008.c in the msMSSQL2008LayerGetShapeCount function. 

Again the spatial filter will not filter out any records, but greatly increases the running time of a query.

jratike80:
I think that `"wfs_use_default_extent_for_getfeature"` could well affect all data sources and not only OGR sources. In addition it could be `"false"` by default .  Or can anybody say when it makes something good with WFS? If WFS user does not include BBOX or any other spatial filter then they want to get data from the whole table so adding a BBOX filter that selects everything feels stupid.

[...]

PR #7197, opened 2024-12-06
Allow use_default_extent_for_getfeature to be used for OGC Features API and PostGIS

geographika:
This pull request adds two new options to MapServer.

1. Allows `wfs_use_default_extent_for_getfeature` to be used for LAYERs using PostGIS connections. This was originally added for OGR layers in 45037eb and then to MSSQL layers (see issue #5994 and implemented by @rouault in #5998). 
Users can now add the following block to avoid adding a default spatial query:

```
METADATA
    ....
    "wfs_use_default_extent_for_getfeature" "false"
    # "ows_use_default_extent_for_getfeature" "false" # ows can also be used following this update to apply to both WFS and OGCAPI requests
END
```

This avoids adding the `ST_Intersects` clause based on the `MAP` extent (see example below), which can greatly speed up attribute or ID based queries. 

```sql
SELECT COUNT(*) FROM (SELECT "id"::text FROM (
                 SELECT id
                   FROM public.table
             ) as subquery WHERE ((ST_Intersects("shape", ST_GeomFromText('POLYGON((-20037508.3427892 -20037508.3427892,-20037508.3427892 20037508.3427892,20037508.3427892 20037508.3427892,20037508.3427892 -20037508.3427892,-20037508.3427892 -20037508.3427892))',3857))) AND 
ST_Intersects(ST_Transform(shape,4326),ST_GeomFromText('POLYGON((-180 -85.0511287798066,-180 85.0511287798066,180 85.0511287798066,180 -85.0511287798066,-180 -85.0511287798066))',4326))))

-- replaced with
SELECT COUNT(*) FROM (SELECT "id"::text FROM (
                 SELECT id
                   FROM public.table
             ) as subquery WHERE (true))

```

Thanks to @jandic for the update and testing on PostGIS. 

2. Allows `ows_use_default_extent_for_getfeature` or `oga_use_default_extent_for_getfeature` to avoid adding a default spatial clause when using the OGC Features API, for example a request to http://localhost/mymap/ogcapi/collections/mn_population_centers/items/35` - without a `bbox` querystring parameter. Again, this can greatly help with database query performance. 

Note `ows_use_default_extent_for_getfeature` can now be used to apply this setting to both WFS and OGC Features API requests.


geographika:
This currently isn't working as expected, due to some strange issue with `MS_INIT_INVALID_RECT`.
In `msPostGISBuildSQLWhere` the following code `const rectObj rectInvalid = MS_INIT_INVALID_RECT;` produces a `rectObj` with `miny` set to a Denormalized number and `minx` set to 0:

```
    minx: 0.0000000000000000
    miny: 1.367854459314e-311#DEN
    maxx: 1.0000000000000001e+300
    maxy: 1.0000000000000001e+300
```

The comparison with the invalid rect then fails as it doesn't match. 
In the same file in `msPostGISLayerGetShapeCount` the exact same code works:

```cpp
  const rectObj rectInvalid = MS_INIT_INVALID_RECT;
  bool bIsValidRect = memcmp(&rect, &rectInvalid, sizeof(rect)) != 0;
```

`rectInvalid` equals:

```
  minx: -1.0000000000000001e+300
  miny: -1.0000000000000001e+300
  maxx: 1.0000000000000001e+300
  maxy: 1.0000000000000001e+300
```
No idea what is happening here. I thought it may only occur when debugging in Visual Studio, but the comparison fails in a Release build.

src/mappostgis.cpp
```diff
@@ -1911,121 +1911,130 @@ static std::string msPostGISBuildSQLWhere(layerObj *layer, const rectObj *rect,
     return std::string();
   }
 
+  const rectObj rectInvalid = MS_INIT_INVALID_RECT;
+  bool bIsValidRect = memcmp(&rect, &rectInvalid, sizeof(rect)) != 0;
```
rouault, line 1915: you are comparing the address of the pointer rect, instead of the pointed value, so:
```suggestion
  bool bIsValidRect = memcmp(rect, &rectInvalid, sizeof(rectInvalid)) != 0;
```

src/mappostgis.cpp
```diff
@@ -2771,6 +2780,8 @@ static int msPostGISLayerGetShapeCount(layerObj *layer, rectObj rect,
 #ifdef USE_POSTGIS
   int rectSRID = -1;
   rectObj searchrectInLayerProj = rect;
+  const rectObj rectInvalid = MS_INIT_INVALID_RECT;
+  bool bIsValidRect = memcmp(&rect, &rectInvalid, sizeof(rect)) != 0;
```
rouault, line 2784: That's correct here, as rectObj is passed by value, and not as a pointer

src/mappostgis.cpp
geographika: Thanks so much @rouault ! Changing that line fixed the debug output of the line above before reaching it..

geographika:
`query_postgis_test012` currently failing - will resolve this.

jmckenna:
is this one ready to be merged?

PR #7198, opened 2024-12-09
Avoid casting unique ID to string if the field is numeric in PostGIS Driver

geographika:
Following on from #7197 - another performance improvement related to the PostGIS driver.

The PostGIS driver currently adds a WHERE clause to the SQL to get a single feature from the database, but converts both the field and value to strings - `WHERE "id"::text = '5'`. This means no index is used to find the feature. Thanks to @jandic for finding this issue. 

This pull request avoids casting the field to a string, which in our use case reduced query times from 2730ms to 144 ms - 19 times faster (the table has ~2 million records). 

OGC Feature API queries such as `mymap/ogcapi/collections/MyLayer/items/1?f=json` are affected as are any `MS_STRING` filter expression used for a PostGIS layer. 

Note a similar issue was fixed in `mapogr.cpp` with https://github.com/MapServer/MapServer/commit/22dc081f1060ca40e0b209e7bb3e9aa0ca8b5ac3 (by @rouault) - and a similar approach was taken in this pull request:

```
// Do not cast if the variable is of the appropriate type as it can
// prevent using database indexes, such as for SQlite
```

The pull request relies on the `"gml_%s_type"` being set for the Id field for the layer (or any field used in a query), which can be added manually, or by setting `"gml_types" "auto"`. 

Not all database drivers are affected, for example MSSQL uses `[id] = '2'` but as the field is not converted the index is still used even though the id is a string and the fields is an integer.

### DIFF
--- a/src/mappostgis.cpp
+++ b/src/mappostgis.cpp
@@ -1580,6 +1580,25 @@ static int msPostGISBase64Decode(unsigned char *dest, const char *src,
 #endif
 
 /*
+** isPropertyNumeric()
+**
+** Check if a field is numeric
+*/
+static bool isPropertyNumeric(layerObj *layer, const char *property) {
+
+  if (!property)
+    return false;
+
+  char md_item_name[256];
+  snprintf(md_item_name, sizeof(md_item_name), "gml_%s_type", property);
+
+  const char *type = msLookupHashTable(&(layer->metadata), md_item_name);
+
+  return (type != nullptr && (EQUAL(type, "Integer") || EQUAL(type, "Long") ||
+                              EQUAL(type, "Real")));
+}
+
+/*
 ** msPostGISBuildSQLBox()
 **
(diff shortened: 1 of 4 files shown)


src/mappostgis.cpp
```diff
@@ -1579,6 +1579,25 @@ static int msPostGISBase64Decode(unsigned char *dest, const char *src,
 
 #endif
 
+/*
+** isPropertyNumeric()
+**
+** Check if a field is numeric
+*/
+static bool isPropertyNumeric(layerObj *layer, const char *property) {
```
rouault, line 1587: would you mind moving that into a msIsPropertyIsNumeric() method (in maplayer.c maybe?), so it can be used here, and by mapogr.cpp ?

jmckenna:
guys is this one ready to be merged?

PR #7203, opened 2024-12-13
Emit explicit error message when proj.db cannot be found

### LINKED ISSUE #7202: Improve error message when proj.db not found
Using PROJ < 6, when the `epsg` file was not found, MapServer would return an error message that is meaningful, such as `msProcessProjection(): No such file or directory`.  That wording at least puts the user onto checking the path of the `epsg` file.

However, when using PROJ >=6, if the `proj.db` file is not found then MapServer returns a cryptic error message (that appears to come direct from the PROJ library:
```
msProcessProjection(): Projection library error. proj error "Invalid PROJ string syntax" for "init=epsg:3857" <br>
```
That wording hints that some syntax mistake was made inside the mapfile (most users will assume that), but in fact here the `proj.db` is not found, causing that error.

This ticket is filed in the hopes that that message can somehow be improved for MapServer users.

rouault:
Fixes #7202

```
$ PROJ_DATA=invalid mapserv -conf ../etc/mapserv.conf QUERY_STRING="map=wms_layer_groups.map&SERVICE=WMS&VERSION=1.1.1&REQUEST=GetCapabilities"
Content-Type: text/html

<HTML>
<BODY BGCOLOR="#FFFFFF">
msProcessProjection(): Projection library error. PROJ error &quot;Cannot find proj.db&quot; when instantiating &quot;init=epsg:4326&quot;
```


jmckenna:
thanks @rouault, great!

jmckenna:
tested on Windows, received:
```
msProcessProjection(): Projection library error. PROJ error "Cannot find proj.db" when instantiating "init=epsg:3857" <br>
```
thanks.

PR #7215, opened 2025-01-30
msDrawRasterBuildRasterPath(): micro-optimization: do not attempt to open non-exiting file

rouault:
For a mapfile *without* SHAPEPATH and with a raster layer pointing to a tileindex layer (e.g. PostGIS), we currently try to access an inexisting file with the name of that tileindex layer.

Really a micro-optimization. Nothing critical. Found when investigated other useless file opening attempts.


jmckenna:
thanks @rouault !

PR #7218, opened 2025-02-06
Update the MapServer OpenLayers Viewer to use OpenLayers 10.4

geographika:
The current built-in [OpenLayers Viewer](http://localhost/mapserver_docs/cgi/openlayers.html) uses [OpenLayers 2](https://openlayers.org/two/) - last updated in 2014. This pull request updates the templates used by the inbuilt viewer to use OpenLayers 10.4 - the current latest version. 

To enable this update a new OpenLayers source was added to the OpenLayers project, to enable map layers to be created using the MapServer CGI parameters, as well as the standard WMS interface. See https://github.com/openlayers/openlayers/pull/16591 for details, and the online example (using the Itasca demo) at https://openlayers.org/en/latest/examples/mapserver-cgi.html. Thanks to @ahocevar for the JS code reviews and accepting the pull request. 

A cut-down version of the OL library has been built - see the build steps in [this repo](https://github.com/geographika/ol-mapserver) that will be used as the default JS library used if `MS_OPENLAYERS_JS_URL` is not set. This will be uploaded to mapserver.org in the pull request at https://github.com/MapServer/MapServer-documentation/pull/1005. 

OpenLayers 3+ has its own CSS file, which isn't automatically picked up by the JS file (as it was in OpenLayers2). For this reason a new `MS_OPENLAYERS_CSS_URL` environment variable has been added as part of this pull request, and the MapServer docs will be updated. By default, it will point to `//mapserver.org/lib/10.4.0/ol-mapserver.css`.

Along with changing the HTML generated by MapServer for the templates, the following changes have been made:

- an optimisataion when used for viewing WMS - previously a GetMap image was created by the call, but then the code went on to generate and return the HTML OpenLayers page. The creation of this image is now skipped, which speeds up initial load. 
- a redundant/confusing error message was removed - `At least one layer name required in LAYERS` (it seems to be copied from somewhere else as it had the wrong function name `msWMSLoadGetMapParams`)
- tests added to check the HTML output via CGI and WMS. There were no tests previously.

### DIFF
--- a/src/maptemplate.c
+++ b/src/maptemplate.c
@@ -4997,8 +4997,6 @@ int msReturnOpenLayersPage(mapservObj *mapserv) {
     openlayersUrl = (char *)tmpUrl;
 
   if (mapserv->Mode == BROWSE) {
-    msSetError(MS_WMSERR, "At least one layer name required in LAYERS.",
-               "msWMSLoadGetMapParams()");
     layer = processLine(mapserv, olLayerMapServerTag, NULL, BROWSE);
   } else
     layer = processLine(mapserv, olLayerWMSTag, NULL, BROWSE);
--- a/src/maptemplate.c
+++ b/src/maptemplate.c
@@ -51,49 +51,73 @@ static inline void IGUR_voidp(void *ignored) {
   (void)ignored;
 } /* Ignore GCC Unused Result */
 
-static const char *const olUrl = "//www.mapserver.org/lib/OpenLayers-ms60.js";
+static const char *const olUrl = "//cdn.jsdelivr.net/npm/ol@v10.4.0/dist/ol.js";
+static const char *const olCssUrl = "//cdn.jsdelivr.net/npm/ol@v10.4.0/ol.css";
+
 static const char *const olTemplate =
-    "<html>\n"
+    "<!DOCTYPE html>\n"
+    "<html lang=\"en\">\n"
     "<head>\n"
-    "<meta content=\"text/html;charset=utf-8\" http-equiv=\"Content-Type\">\n"
-    "  <title>MapServer Simple Viewer</title>\n"
-    "    <script type=\"text/javascript\" "
-    "src=\"[openlayers_js_url]\"></script>\n"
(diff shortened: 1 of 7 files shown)


geographika:
CI is failing when building the Fuzzers (not running them) with lots of `undefined reference to `__extendhfsf2` errors. 

See logs at https://github.com/MapServer/MapServer/actions/runs/13178105581/job/36798663506#step:4:8098

```
gdalrasterize.cpp:(.text._ZL14gvBurnScanlineP17GDALRasterizeInfoiiid[_ZL14gvBurnScanlineP17GDALRasterizeInfoiiid]+0x556e): undefined reference to `__extendhfsf2'
```

src/maptemplate.c
```diff
@@ -4996,22 +5025,29 @@ int msReturnOpenLayersPage(mapservObj *mapserv) {
   if (tmpUrl)
     openlayersUrl = (char *)tmpUrl;
 
+  /* now do the same for the MS_OPENLAYERS_CSS_URL */
+  tmpUrl = msGetConfigOption(mapserv->map, "MS_OPENLAYERS_CSS_URL");
+  if (tmpUrl == NULL)
+    tmpUrl = CPLGetConfigOption("MS_OPENLAYERS_CSS_URL", NULL);
+
+  if (tmpUrl)
+    openlayersCssUrl = (char *)tmpUrl;
```
rouault, line 5034:
```suggestion
    openlayersCssUrl = tmpUrl;
```

geographika: This was copied from the JS setting, so I presume that should also be updated?

```
  if (tmpUrl)
    openlayersUrl = (char *)tmpUrl;
```

rouault: > This was copied from the JS setting, so I presume that should also be updated?
> 
> ```
> 
> ```

yes, you can always (I shouldn't say "always" as I'm pretty sure someone could invent on purpose a corner use case in C++ where that wouldn't be true) safely remove such const related casts as soon as the compiler doesn't warn when they're removed

rouault (approved):

PR #7225, opened 2025-02-12
CI: Run coverity-scan using Docker

geographika:
As with #7224 run the coverity-scan workflow using Docker.
As this is only run once a week, it may need to be merged to see if it works.


geographika:
Merging and will check it runs successfully tomorrow.

Issue #7228, opened 2025-02-18
CI Failing due to updated GDAL Version


geographika:
CI is currently failing as `ubuntugis/ubuntugis-unstable` updated its version of GDAL for Focal (in the middle of my updates to migrate to Docker).

As the msautotests will all need to be fixed anyway for this change, it seemed like a good opportunity to update the build scripts from `ubuntu:20.04` to `ubuntu:24.04`. I currently have a WIP branch at https://github.com/geographika/mapserver/commits/ci-ubuntu-update/

I'll open up a pull request at some point this week, although there are a few issues I'll need help with on SVG changes and cppcheck issues.

geographika:
Closed with #7235 (staying with ubuntu:20.04 for now).

PR #7229, opened 2025-02-18
CI: Update builds to Ubuntu 24.04 (LTS Noble Numbat)

geographika:
## Background

Recent pull requests #7227 #7225 and #7224 moved as much of the CI build logic from GitHub Actions to Docker containers running shell scripts. However, due to some unfortunate timing (!) these were passing when the pull requests were opened but failed when merged. This was due to `ubuntugis/ubuntugis-unstable` updating the GDAL package for `ubuntu:20.04` (see #7228) from GDAL 3.4.3 to GDAL 3.9.3. The GDAL updates cause many differences in msautotest outputs.

In hindsight, I should have just fixed these outputs, but as at some point we'd have to move to a new version of Ubuntu I thought it would be a good time to update the Ubuntu base images and GDAL fixes..

After checking the history of the CI the last time this was done was in "Vagrant and Travis-CI: update to Ubuntu Bionic" #6111 in 2020 by @rouault, and involved 313 file changes (see also https://github.com/rouault/mapserver/commit/c71a4f3cc10c7c659830165b3e4c94faeb1e38cf updating tests from 6.4 to 7.0 where 1,571 files were changed). 

## Updates

I tried to keep a clean commit history with relevant comments, but due to lots of trial and error this isn't always the case. 

Key updates are:

- Updated Bison 3.0.4 to Bison 3.8.2 and regenerating the parser files
- Moved all CI workflows to Docker, including the PHP MapScript builds
- Updated PostgreSQL version
- Simplified the MapScript CI builds so they don't share the full build.sh script
- Updated various expected msautotest outputs due to changes in GDAL

## Notes

- Is FindGDAL still required? It looks like this is now a part of CMake: https://cmake.org/cmake/help/latest/module/FindGDAL.html
- Is vagrant still needed if we can use Docker to build MapServer and run msautotest, and the results stay consistent with the CI? Vagrant will be handled in a separate pull request. 
- I have a new Python script for checking perceptual differences in image files and PDFs which helped greatly for finding image outputs that remained visually the same. I will add this to /scripts in a separate pull request.

## TODOs

- the majority of failing tests are SVG outputs which do look different (sizes, background colours). For background Ubuntu 20.04 LTS (Focal Fossa, used by the CI up to now) shipped with librsvg 2.48, the new Ubuntu image uses 2.58. @mapcentia  - maybe you have more knowledge about the SVG changes, and there could be a link to #7054?
- Fix the remaining msautotests
- Fix the cppchecks. This was updated from 1.90 to 2.7 when moving to the new Ubuntu image. The log can be found in the GitHub Actions log.


rouault:
> Fix the cppchecks. This was updated from 1.90 to 2.7 when moving to the new Ubuntu image. The log can be found in the GitHub Actions log.

All/most are false positives. We should either add suppressions in the code, or filter them out in the scripts/cppcheck.sh script

geographika:
Replaced with #7242.

PR #7233, opened 2025-02-20
Insert a wisdom of the day in WMS, WCS and WFS GetCapabilities response

rouault:
In the same vain as https://github.com/OSGeo/gdal/pull/11865 and https://github.com/OSGeo/PROJ/pull/4406

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1140,6 +1140,8 @@ mark_as_advanced(
 add_definitions(-DMAPSERVER_BUILD)
 add_definitions(-DMAPSERVER_CONFIG_FILE="${MAPSERVER_CONFIG_FILE}")
 
+add_definitions(-DMAPSERVER_WISDOM_OF_THE_DAY="Nothing better than climbing Denali to enlighten your mind.")
+
 install(
   FILES ${PROJECT_SOURCE_DIR}/share/ogcapi/templates/html-plain/collection.html
         ${PROJECT_SOURCE_DIR}/share/ogcapi/templates/html-plain/collection-item.html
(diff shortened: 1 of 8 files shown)


jmckenna:
Respectfully, MapServer is not about politics, and should remain free of politics.  I am a strong -1 to bring politics into MapServer (taking a stance on one side or another, in any global conflict).  This should be discussed as a PSC.  I do see a few projects taking stances one way or another, but in the case of MapServer I hope we can not go into those weeds, instead, let's continue to make open & fast mapping software, and put our free time & effort into that, instead of wasting time picking sides & debating wisdom quips.  Who really has time for that.

jmckenna:
See similar reaction in the PROJ pull request, by the PROJ maintainer: https://github.com/OSGeo/PROJ/pull/4406#issuecomment-2672368389

rouault:
> instead of wasting time picking sides

We're talking about freedom and world stability. That's not a waste of time

eldang:
While I am very much in favour of taking sides, doing it in this way carries an unacceptable risk of collateral damage.  Right now in the United States, the very people you're [justifiably] asking us to take sides against are using the presence of innocuous words like "gender" as a pretext to censor entire datasets that are of value to the entire world.  Getting our slogans into datasets that they are unrelated to will only make it harder to resist this line of attack.

I appreciate the sentiment here, and want to make very clear that we are approaching this from the same side.  But please find a better way to lodge the protest.

[this also applies to https://github.com/OSGeo/gdal/pull/11875 & https://github.com/OSGeo/gdal/pull/11876 ]

strk:
fortune files would be a great place for those words of wisdom, how about starting a project to publish such fortunes ?

rouault:
closing

Issue #7226, opened 2025-02-13
SCALETOKEN not working in FILTER


jkempe:
It looks like the use of a SCALETOKEN is not working as expected when used in a FILTER.

If I define a `SCALETOKEN` like this:
```
SCALETOKEN
    NAME "%ne_scalerank%"
    VALUES
        "0" "10"
        "1000" "8"
        "10000000" "5"
    END # VALUES
END # SCALETOKEN
```

and try to use it in a `FILTER` like this:
`FILTER ([scalerank] < %ne_scalerank% )`

all features are dropped and the logs shows:
```
[Wed Feb 12 13:50:20 2025].23938 replacing scaletoken (%ne_scalerank%) with (5) in layer->filter (ne_10m_lakes) for scale=221471921.250000
[Wed Feb 12 13:50:20 2025].24460 msEvalExpression(): General error message. Cannot evaluate expression, no item index defined.
[Wed Feb 12 13:50:20 2025].24474 msEvalExpression(): General error message. Cannot evaluate expression, no item index defined.
[Wed Feb 12 13:50:20 2025].24482 msEvalExpression(): General error message. Cannot evaluate expression, no item index defined. 
```

Expected behavior is that features from the shapefile with a scalerank < 5 are being shown on the map.

Mapfile to reproduce:
```
MAP
    CONFIG "MS_ERRORFILE" "ms_error.txt"

    STATUS ON

    PROJECTION
        "init=epsg:4326"
    END

    EXTENT -180 -90 180 90

    WEB
        METADATA
            "wms_title"           "WMS Demo Server"
            "wms_srs"             "EPSG:4326"
            "wms_enable_request"  "*"
        END
    END

    LAYER
        DEBUG 5
        NAME 'ne_10m_lakes'

        TYPE POLYGON
        PROJECTION
            "init=epsg:4326"
        END
        EXTENT -180 -90 180 90
        DATA "ne_10m_lakes.shp"

        SCALETOKEN
            NAME "%ne_scalerank%"
            VALUES
                "0" "10"
                "1000" "8"
                "10000000" "5"
            END # VALUES
        END # SCALETOKEN

        FILTER ([scalerank] < %ne_scalerank% )

        CLASS
            STYLE
                OUTLINECOLOR "#647078"
                WIDTH 1
                COLOR "#EBEEF0"
            END
        END

    END

END
```

[...]

PR #7237, opened 2025-02-21
Allow FILTER expressions with SCALETOKENs

geographika:
Fix for #7226. 

The issue was that once the original filter was parsed, the parentheses were removed, and it became a string e.g. `([EAS_ID] = %filter% )` became `"[EAS_ID] = %filter%"` so features were not filtered correctly. 

This fix adds back the parentheses - but only when the FILTER was originally an expression, so string filters are preserved. 

@jkempe - can you check this for your test case?


src/maplayer.c
```diff
@@ -105,7 +105,15 @@ int msLayerRestoreFromScaletokens(layerObj *layer) {
     layer->tileitem = layer->orig_st->tileitem;
   }
   if (layer->orig_st->filter) {
-    msLoadExpressionString(&(layer->filter), layer->orig_st->filter);
+    if (layer->filter.type == MS_EXPRESSION) {
+      char *tmpval = (char *)msSmallMalloc(sizeof(char) *
```
rouault, line 109: Although it is fine here, for the sake of the relief of our souls, I would store the size in a variable and then use it with snprintf()

something like
```
const size_t tmpval_size = strlen(layer->filter.string) + 3;    // you can drop sizeof(char). It is always 1
char *tmpval = (char *)msSmallMalloc(tmpval_size);
snprintf(tmpval, tmpval_size, "(%s)", layer->filter.string);
```
same below

src/maplayer.c
```diff
@@ -105,7 +105,15 @@ int msLayerRestoreFromScaletokens(layerObj *layer) {
     layer->tileitem = layer->orig_st->tileitem;
   }
   if (layer->orig_st->filter) {
-    msLoadExpressionString(&(layer->filter), layer->orig_st->filter);
+    if (layer->filter.type == MS_EXPRESSION) {
+      const size_t tmpval_size = strlen(layer->filter.string) + 3;
+      char *tmpval = (char *)msSmallMalloc(tmpval_size);
+      snprintf(tmpval, "(%s)", tmpval_size, layer->filter.string);
```
rouault, line 111:
```suggestion
      snprintf(tmpval, tmpval_size, "(%s)", layer->filter.string);
```

geographika: Yes sorry ignore that error!

geographika:
Doesn't look like feedback will be forthcoming, but issue confirmed and fixed with test.

PR #7242, opened 2025-02-27
CI: Update all build scripts and msautotest outputs to Ubuntu 24.04

geographika:
## Background

Replaces #7229. This updates all CI builds and tests to use Ubuntu 24.04 (LTS Noble Numbat). The last time this was done was in "Vagrant and Travis-CI: update to Ubuntu Bionic" #6111 in 2020, and involved 313 file changes (see also https://github.com/rouault/mapserver/commit/c71a4f3cc10c7c659830165b3e4c94faeb1e38cf updating tests from 6.4 to 7.0 where 1,571 files were changed). 

## Key Updates

- Updated various expected msautotest outputs due to changes in various dependencies when moving from Ubuntu 20.04 to 24.04. Now tested with PROJ version 9.4 and GDAL version 3.9. 
- Updated Bison 3.0.4 to Bison 3.8.2 and regenerating the parser files
- Updated PostgreSQL version
- Updated Vagrant scripts to Ubuntu 24.04 (used to generate new msautotest images and get tests passing locally)

## Other Notes

- Majority of differences were due to librsvg update from 2.48 to 2.58. Since then librsvg has switched to Rust, and previously added explicit pt (points) units to width and height when generating SVG output (e.g. `width="400pt" height="300pt"` is now `width="400" height="300"`). A `MAP SIZE` of `SIZE 200 150` now prodcues an image of 200x150 pixels (rather than 266x200) which seems to make more sense. Also older versions defaulted to a white background when rendering transparent SVGs, whereas the newer versions (2.50+) default to transparency instead of white. 
- I've set the cppchecks CI to pass even though there are errors. I'll try and address these in separate pull requests, as I'm worried attempting to fix some of them may cause more harm than good.
- The initial `vagrant up` runs successfully but appears to build MapServer twice. This may be due to discrepancies between the Vagrant VM and the local file system (Windows). 
- There are a few msautotests that I couldn't fix. I'll open a separate issue for these. 
- MapScript test fixes including Perl MULTIPOINT, and Mono garbage collection changes
- Other changes causing reference test data to require updates include the following (I tried to group these in commits):
  - GDAL adding `"xy_coordinate_resolution"` property to GeoJSON output
  - GDAL adds a single whitespace to the end of DAT file outputs
  - GDAL outputting numeric fields as numbers rather than strings e.g. `1` rather than `"1"`
  - PG16 no longer requires aliases for subqueries
  - Small precision differences due to a new version of PROJ
  - NetCDF reads STATISTICS metadata from auxiliary .XML files which have been added to the reference test data


jratike80:
Lot of work, thank you!

geographika:
Test differences listed in #7243 and cppcheck warnings in #7244. 

I'd like to merge this and address those issues separately. I'm unsure whether to keep the individual commits (the majority have useful commit messages), or to do a bug squash and merge, losing the commit history. Any thoughts?

rouault:
Excellent work ! I would probably go for a commit squash to get "atomic" history (i.e. each revision passing tests)

Issue #7244, opened 2025-02-27
Issues following update to Cppcheck 2.13


geographika:
As part of #7242 Cppcheck was updated from 2.13.0 (from 1.9.0). This raised the issues below that need to be addressed. The checks are currently set to pass in the CI until they can be fixed in https://github.com/MapServer/MapServer/pull/7242/commits/00ef1a51564d87f3ed7743c3ebfc713b736d2434

```
2025-02-27T15:36:19.5194892Z /home/runner/work/mapserver/mapserver/src/maplabel.c:1185,warning,uninitvar,Uninitialized variables: pnts1.x, pnts1.y, pnts1.z, pnts1.m
2025-02-27T15:36:19.5196086Z /home/runner/work/mapserver/mapserver/src/maplabel.c:1195,warning,uninitvar,Uninitialized variables: pnts2.x, pnts2.y, pnts2.z, pnts2.m
2025-02-27T15:36:19.5197039Z /home/runner/work/mapserver/mapserver/src/maputil.c:1998,warning,uninitvar,Uninitialized variables: old_diffdir.z, old_diffdir.m
2025-02-27T15:36:19.5197948Z /home/runner/work/mapserver/mapserver/src/maputil.c:1999,warning,uninitvar,Uninitialized variables: old_offdir.z, old_offdir.m
2025-02-27T15:36:19.5198497Z uninitvar check failed
2025-02-27T15:36:19.5436177Z /home/runner/work/mapserver/mapserver/src/mappostgis.cpp:2654,performance,passedByValueCallback,Function parameter 'rect' should be passed by const reference. However it seems that 'msPostGISLayerWhichShapes' is a callback function.
2025-02-27T15:36:19.5441122Z /home/runner/work/mapserver/mapserver/src/mappostgis.cpp:2783,performance,passedByValueCallback,Function parameter 'rect' should be passed by const reference. However it seems that 'msPostGISLayerGetShapeCount' is a callback function.
2025-02-27T15:36:19.5444606Z /home/runner/work/mapserver/mapserver/src/mapunion.cpp:365,performance,passedByValueCallback,Function parameter 'rect' should be passed by const reference. However it seems that 'msUnionLayerWhichShapes' is a callback function.
2025-02-27T15:36:19.5446757Z passedByValue check failed
[...]

PR #7247, opened 2025-03-04
Fix knownConditionTrueFalse cppcheck warnings

geographika:
Follow-up to #7244. Return failure if `msInsertHashTable` returns NULL to fix one of these warnings.
The majority of the other warnings are checking functions that call a few macros (defined in https://github.com/MapServer/MapServer/blob/main/src/mapcopy.h) e.g.

```
  MS_COPYSTRING(dst->string, src->string);
  MS_COPYSTELEM(type);
  MS_COPYSTELEM(flags);

  return MS_SUCCESS;
```

I added checks to see that `dst->string` isn't NULL, although not sure if this should be done for all calls to `MS_COPYSTRING`. 
An alternative is to remove any checks for MS_SUCCESS in the parent functions. 
If the macros fail to run, I assume MapServer will crash without ever reaching `return MS_SUCCESS`.

### DIFF
--- a/src/mapcopy.c
+++ b/src/mapcopy.c
@@ -162,8 +162,9 @@ int msCopyHashTable(hashTableObj *dst, const hashTableObj *src) {
     key = msNextKeyFromHashTable(src, key);
     if (!key)
       break;
-    else
-      msInsertHashTable(dst, key, msLookupHashTable(src, key));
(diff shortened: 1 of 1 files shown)


jmckenna:
(adding to 8.6.0 milestone)

geographika:
Closing for now as low priority, and I won't get a chance to look at this anytime soon.

PR #7252, opened 2025-03-11
Make sure 130 tests make request for WMS version 1.3.0

alistaireverett:
I came across some tests which from the name should have been testing WMS server 1.3.0, but were making requests for version 1.1.1. I guess these have just been missed sometime in the past so I just went ahead and updated them.


jratike80:
It may be good to check carefully what these tests are testing. They are about WMS Client, so Mapserver sends requests to a remote WMS. What if user sends WMS 1.1.0 request to Mapserver, and Mapserver must understand to convert the request into WMS 1.3.0 for making the remote WMS happy, or vice versa?

It may well be that your fixes are correct, just thinking.

alistaireverett:
You could have point here. One of the reasons I looked at this was because of axis order changes between WMS 1.1.1 and 1.3.0 and checking these were handled correctly. In `wms_client_130_axis_fail.map` for example, I think `"wms_strict_axis_order" "0"` would be expected to have no effect when the request is made for WMS 1.1.1, because the axis order is also the same, so then it made sense to me that it should make a request for 1.3.0 instead. But I'm not 100% certain this was the intention of the tests.

Another option would be to keep the existing tests, but add new tests with requests for version 1.3.0. But I think you are correct it is probably worth waiting to see if someone else who has more knowledge of these tests is in agreement before this is merged in case I haven't fully understood things.

ejn:
It's a long time ago, but iirc my intention for these tests was that they are made using WMS 1.1.1 and Mapserver should then make the appropriate requests to the backend using WMS 1.3.0, including axis order mangling as configured and/or required (and in the `fail`-case the configuration should be inappropriate, resulting in the fail). So Mapserver as WMS 1.1.1 server but 1.3.0 client.

So my recommendation would be to keep the current tests for (client) -> WMS 1.1.1 -> (mapserver) -> WMS 1.3.0 -> (backend) and add new tests for (client) -> WMS 1.3.0 -> (mapserver) -> WMS 1.3.0 -> (backend) as necessary.

jmckenna:
@ejn I fully agree with your suggested plan to have both tests

alistaireverett:
Great, thanks for the input. I'll update this pull request to keep the existing tests using 1.1.1 and add test using requests for 1.3.0 later this afternoon.

jmckenna:
thanks @alistaireverett !

PR #7255, opened 2025-03-20
Fix opacity for gradient legend symbol

PatrikSylve:
The AGG renderer did not handle alpha values when rendering pixmap symbol. So if OPACITY was used together with COLORRANGE the colors would be off. 

The AGG method did accept an alpha parameter which seems to work better.   

In the provided test, this is the result I got before:  
wms_gradient_legend3

And with this fix, it looks like this: 
wms_gradient_legend3


jmckenna:
Thanks @PatrikSylve, and thanks for including a test.

PR #7259, opened 2025-03-26
 Bump C/C++ standard to 17 for ICU 76.

sebastic:
As reported in [Debian Bug #1101239](https://bugs.debian.org/1101239), MapServer fails to build with ICU 76 (pulled in via libxml2) which now requires C++17.

### DIFF
--- a/.github/workflows/start.sh
+++ b/.github/workflows/start.sh
@@ -40,7 +40,7 @@ ci/ubuntu/build.sh
 
 # Validate openapi document
 pip install check-jsonschema --break-system-packages
-wget https://raw.githubusercontent.com/OAI/OpenAPI-Specification/main/schemas/v3.0/schema.yaml -O openapi_schema.yaml
+wget https://raw.githubusercontent.com/OAI/OpenAPI-Specification/refs/heads/main/_archive_/schemas/v3.0/schema.yaml -O openapi_schema.yaml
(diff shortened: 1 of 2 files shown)


sebastic:
The two test failures are caused by changes to the OAI repo, the schemas moved to the `_archive_` subdirectory.

New URL:

 https://raw.githubusercontent.com/OAI/OpenAPI-Specification/refs/heads/main/_archive_/schemas/v3.0/schema.yaml

rouault:
Given that libxml2 is a library with a C API, couldn't it be argued that it's a bug of its public headers to require a C++ compiler ?
That said, given that both recent GDAL and PROJ which are mandatory Mapserver dependencies require C++17, I don't see any harm in making that a requirement for Mapserver too. CC @sdlime @geographika

jmckenna:
In terms of Windows builds, this would mean requiring MSVC 2019 at the minimum (MS4W 5+ is built off of MSVC 2022).  

Important discussion, pinging `-dev` list now also...

rouault:
> Are most OSes capable of running a compiler with C++17 support?

yes, see https://en.wikipedia.org/wiki/C%2B%2B17#Compiler_support

geographika:
+1 on bumping. In general, staying in sync with GDAL seems to be a good approach.

jmckenna:
Thanks @sebastic and thanks everyone for your fast feedback.

PR #7263, opened 2025-04-02
Add DEFAULTVALUE to be used in legend when STYLE SYMBOL is an attribute.

TAlonglong:
When reading the map file, initStyle symbol is assign to the number 0.

When STYLE SYMBOL is an attribute, and the request is GetLegendGraphics, bindings are not replaceing the attribute and symbol number 0 is used. And the SYMBOL 0 is a dummy symbol added by default by mapserver.

By adding DEFAULTSYMBOL as in this PR, this can be set to a valid number of the symbols available to be used in this case.

eg:
```
CLASS
  NAME "Name of class"
  GROUP "groupname"
  EXPRESSION ('[var]' >= '0' and '[var]' < '33')
  STYLE
      SYMBOL [symbol_name]
      COLOR "some color"
      DEFAULTSYMBOL 1
  END
END
CLASS
....
END
.....
```

There are issues with this. The user can supply a DEFAULTSYMBOL with a not valid symbol number, with unknown results. ANd maybe others.

It solve my problem, but I don't know how many other problems I introduce.

Any views on this?

### DIFF
--- a/src/mapfile.c
+++ b/src/mapfile.c
@@ -3009,9 +3009,9 @@ int loadStyle(styleObj *style) {
       if (symbol == MS_NUMBER) {
         if (msCheckNumber(msyynumber, MS_NUM_CHECK_GTE, 0, -1) == MS_FAILURE) {
           msSetError(MS_MISCERR,
-                    "Invalid DEFAULTSYMBOL id, must be greater than or equal "
-                    "to 0 (line %d)",
-                    "loadStyle()", msyylineno);
+                     "Invalid DEFAULTSYMBOL id, must be greater than or equal "
+                     "to 0 (line %d)",
+                     "loadStyle()", msyylineno);
           return (MS_FAILURE);
         }
         style->symbol = (int)msyynumber;


jmckenna:
@TAlonglong please also include a test for this in msautotest, as part of your pull request (this will help you understand how your proposed change impacts other parts of MapServer).  Follow: https://mapserver.org/development/tests/autotest.html

TAlonglong:
Thanks for your suggestion Jeff. That was my intention also, but delay writing test as long as possible...

One thing. I try to run the tests in main. But still it seems that a lot of tests are failing. Is this expected?

jmckenna:
Thanks for your work on this @TAlonglong 

What happens if you pass a name of a symbol to DEFAULTSYMBOL?

jmckenna:
If it is not supported, then that is why documenting this is important for the users - are you able to add this new parameter into the docs, as a pull request to the [MapServer-documentation](https://github.com/MapServer/MapServer-documentation) repo?

thanks again.

TAlonglong:
> Thanks for your work on this @TAlonglong
> 
> What happens if you pass a name of a symbol to DEFAULTSYMBOL?

https://github.com/MapServer/MapServer/pull/7263/commits/ddb5fbaf1291c793c9f3c96fbe63fc608b591e73#diff-d78ab669aaeeadfaa7ea06cfeea04f3d6bddae3ccf90c48b17495115baac112eR3006-R3008

It will return a MS_FAILURE

But yes, I can add this to the documentation. But I will have a few days of now, so maybe Wednesday in almost 2 weeks time

jmckenna:
I had meant what error message does the user see? (when a user specifies a symbol name, does the error say something about "symbol name is not supported, please specify an integer for the id of the symbol" ?).  Anyway, that is minor, but I was just wondering.   Thanks again.

geographika:
@TAlonglong couldn't this be solved by adding a new attribute in a DATA clause that set a default if another attribute was empty?
E.g. if the source was PostGIS you could use `COALESCE(name, 'Point') AS name_with_default`. This could be added to most data sources. 

An alternative, which could be useful in lots of cases, would be a new [string](https://mapserver.org/mapfile/expressions.html#functions-that-return-a-string) expression that allowed a default if empty. E.g. `ifempty(n1, "default_value")`. 
The STYLE SYMBOL property would also then need updating to accept an expression.

TAlonglong:
> @TAlonglong couldn't this be solved by adding a new attribute in a DATA clause that set a default if another attribute was empty? E.g. if the source was PostGIS you could use `COALESCE(name, 'Point') AS name_with_default`. This could be added to most data sources.

Not sure if I follow you. Problem is anyway the the GetLegendGraphics request does not run any postgis call as far as I could figure out.

> 
> An alternative, which could be useful in lots of cases, would be a new [string](https://mapserver.org/mapfile/expressions.html#functions-that-return-a-string) expression that allowed a default if empty. E.g. `ifempty(n1, "default_value")`. The STYLE SYMBOL property would also then need updating to accept an expression.

Hm, yes this could be a solution. This needs to be handled before the attribute handeling.

geographika:
@TAlonglong - wouldn't KEYIMAGE work better for this case? See https://mapserver.org/mapfile/class.html#mapfile-class-keyimage
The STYLE could also use attributes for COLOR etc. as well as the SYMBOL so using a KEYIMAGE would allow the Mapfile creator to decide exactly what should be shown in the legend output.

TAlonglong:
Sorry for the long delay, but I have finally taken some time to look at this again.

Thanks for the suggestion of KEYIMAGE. Very usefull, it will might fit my needs. I would need an image for each style, potential leading to 100 ish similar images with different colors pending the colorscale.

I will need to test a bit before I make a decision.

TAlonglong:
My inital thought was to have a specific wind barb (hence default symbol) in the legend. But as I only need to illustrate which color is matching the height of the wind, this might be a better solution using KEYIMAGE. I will need to ask my users what they think.
image

jmckenna:
@TAlonglong polite nudge (as we'll likely be starting a release cycle for 8.6.0 - just letting you know, in case you want to get this included)

geographika:
If KEYIMAGE solves this use-case, I'd be in favour of closing this. It could be useful, but then default color, size, etc. couldn't be set, whereas KEYIMAGE should provide this flexibility.

PR #7274, opened 2025-04-28
SHP/DBF: Attempt to read encoding from CPG file  

PatrikSylve:
When a tile index references file paths containing special characters, e.g. `ø`, mapserver will not properly encode the attributes, leading to errors when attempting to open the files:  

> msDrawRasterLayerLow(): Unable to access file. Corrupt, empty or missing file...

Setting the ENCODING on the layer and copy this encoding to the temporary layer that is created in `msDrawRasterSetupTileLayer`, ensures that MapServer correctly encodes file paths containing special characters

```
LAYER
...
   ENCODING "ISO-8859-1"
```


PatrikSylve:
> Do you have tileindex in shapefile? Would having a .cpg file be an option? Does GeoPackage tileindex require encoding as well, even it is always usign UTF-8?


Yes, my tileindex is a shapefile. I am using [`gdaltindex`](https://gdal.org/en/stable/programs/gdaltindex.html) to generate it and it, and unfortunately it does not output a CPG-file.  

I think in this case, using GeoPackage instead would likely work out of the box. So that might be a workaround.

jratike80:
> Yes, my tileindex is a shapefile. I am using [gdaltindex](https://gdal.org/en/stable/programs/gdaltindex.html) to generate it and it, and unfortunately it does not output a CPG-file.

Have you tried to write it yourself? Create a text file with .cpg extension, and I think that writing this line into it should be correct:

`ISO 88591`

PatrikSylve:
> Have you tried to write it yourself? Create a text file with .cpg extension, and I think that writing this line into it should be correct:
> 
> `ISO 88591`

Hm this does not seem to work. I tested with  `ISO 88591`, `ISO-8859-1`, `ISO 8859 1`, but got the same issue.

jratike80:
Ok, maybe Mapserver is reading shapefiles with its own methods and the GDAL documentation does not apply https://gdal.org/en/stable/drivers/vector/shapefile.html#encoding 
If you are curious you can try how QGIS shows the attributes with different .cpg files.

I believe I have managed to avoid scandinavian letters in the paths and therefore have never faced that trouble.

PatrikSylve:
Yes looks like custom handling in MapServer in `mapshape.c`. 

Found this in QGIS, https://github.com/qgis/QGIS/blob/master/src/analysis/processing/qgsalgorithmshpencodinginfo.cpp. I'll investigate a bit and see if something similar could be implemented in Mapserver.

jmckenna:
thanks @PatrikSylve nice addition.  Please also include a test for this, as part of your pull request, in /[msautotest ](https://mapserver.org/development/tests/autotest.html) (I think we need more tests for umlauts/special characters, in msautotest, especially as so many users still rely on tileindex + shp with MapServer)

jmckenna:
thanks again @PatrikSylve

jratike80:
I tried to find some reference about what values are valid in the .cpg file.
GDAL does the codepage things around here https://github.com/OSGeo/gdal/blob/master/ogr/ogrsf_frmts/shape/ogrshapelayer.cpp#L296
and it has taken the reference from
https://support.esri.com/en-us/knowledge-base/read-and-write-shapefile-and-dbase-files-encoded-in-var-000013192

PatrikSylve:
> I tried to find some reference about what values are valid in the .cpg file. GDAL does the codepage things around here https://github.com/OSGeo/gdal/blob/master/ogr/ogrsf_frmts/shape/ogrshapelayer.cpp#L296 and it has taken the reference from https://support.esri.com/en-us/knowledge-base/read-and-write-shapefile-and-dbase-files-encoded-in-var-000013192

Good reference. I see it would fail now if the CPG contained just the code `88591` for an example, GDAL capture this fine. I'll review the code again in a new PR and make sure to capture the codes listed in the ESRI reference.

Issue #7279, opened 2025-05-12
CI currently failing


geographika:
Likely due to a switch from GDAL 3.9.3 to 3.10.3 (on 5/5/25) causing differences in image outputs. 
See https://github.com/MapServer/MapServer/issues/7228 for previous changes required to handle a new version of GDAL.

PR #7280, opened 2025-05-15
WMS_SERVER: Add sanity check for TRANSPARENT parameter value in GetMap

tschorr:
According to the [OGC WMS specification](https://portal.ogc.org/files/?artifact_id=14416), the only allowed values for the TRANSPARENT parameter of the GetMap request are TRUE and FALSE. However MapServer only checks for the value TRUE, all other (arbitrary) values evaluate to false.

This causes a test in the INSPIRE validator/ETS repository to fail: https://github.com/inspire-eu-validation/ets-repository/blob/28b66265aab8fc8228672283887475818aade1ab/service/vs-wms-soapui-project.xml#L5084-L5180.

This PR adds a sanity check for the GetMap TRANSPARENT parameter value - allow only TRUE or FALSE and raise a service exception for other values.


jratike80:
That it true by the standard, even so that only upper case TRUE or FALSE are allowed.
It seems that Geoserver accepts `&TRANSPARENT=FOO` and sends a map. Geoserver can be run in a special CITE compliant mode, maybe those who use Geoserver for INSPIRE are using that mode.

tschorr:
The 2 failing checks seem unrelated, at least I cannot find any failing WMS requests in the logs.

jratike80:
Are you sure that your change does not break any existing applications?  It may be a surprise if someone has been sending &TRANSPARENT=NO for years and then after the next upgrade the application stops working.

The change is correct, the standard requires TRUE|FALSE, but the only real advantage is that the INSPIRE validator gets happy. From users perspective things work correctly already. TRUE gives maps with transparency, FALSE gives maps without transparency, the default without &TRANSPARENT= means no transparency.

tschorr:
> Are you sure that your change does not break any existing applications? It may be a surprise if someone has been sending &TRANSPARENT=NO for years and then after the next upgrade the application stops working.

Quite on the contrary I'm sure my changes definitely *will* break existing applications that are sending wrong parameter values for TRANSPARENT. But I assume this could be dealt with with a sensible release policy and mentioning in the release notes.

 > The change is correct, the standard requires TRUE|FALSE, but the only real advantage that the change gives is that the INSPIRE validator gets happy. From users perspective things work correctly already. 
 
The current behaviour doesn't conform to the standard, so I'd say it doesn't work correctly. I can see how this just works fine for many people, although one could argue that e.g. your example from above might lead people into thinking &TRANSPARENT=YES will actually work...

> TRUE gives maps with transparency, FALSE gives maps without transparency, the default without &TRANSPARENT= means no transparency. The OGC test suite does not test this thing and therefore MapServer WMS 1.3.0 is certified https://portal.ogc.org/public_ogc/compliance/product.php?pid=1083
> 
> See https://cite.ogc.org/teamengine/about/wms-client/1.3.0/site/
> 
> > Capabilities not tested
> > GetMap request for transparent picture [7.3.3.9]
> 
> The best option would be if INSPIRE validator would test only &TRANSPARENT=TRUE and &TRANSPARENT=FALSE. Unfortunately by the discussion on that old, closed issue I believe that INSPIRE validator remains strict.

There's maybe missing a link here - what old closed issue are you referring to? But apart from that, the INSPIRE check might be overdoing things a bit but on the whole it is a valid test.

> The second best option might be to make the strict TRUE|FALSE configurable and default to current behavior but that would require more work and is something that Mapserver normally does not support.

I agree this would require more work and most likely would need an RFC to start with. Here I'm proposing a fix for the problem at hand. For others to decide whether this PR will be accepted.

jmckenna:
@tschorr please also include a test for this, as part of your pull request, in /[msautotest ](https://mapserver.org/development/tests/autotest.html)

geographika:
@tschorr thanks for raising this. 

For the OGC Features API compliancy, an additional METADATA item was added, that allowed users to run a Mapfile in a compliant mode. That would be my preference in this case too (with the default set to `false`). 

```
      "oga_compliance_mode" "true"
```

The implementation could work the same way. See https://github.com/MapServer/MapServer/blob/d54aae0311066e116974ef888b82959bcefb8c24/src/mapogcapi.cpp#L1306

The failing tests is likely due to https://github.com/MapServer/MapServer/issues/7279

I'd be very much in favour of adding in a compliant WMS mode, with a test. 

```
  WEB
    METADATA
      ...
      "wms_compliance_mode" "true"
    END
  END
```

tschorr:
> > ```
> >       "oga_compliance_mode" "true"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> >   WEB
> >     METADATA
> >       ...
> >       "wms_compliance_mode" "true"
> >     END
> >   END
> > ```
> 
> Do we want to restrict this to WMS? What about having an "ogc_compliance_mode"?

or rather 'ows_compliance_mode' maybe.

geographika:
> or rather 'ows_compliance_mode' maybe.

Yes - the way it works when looking up metadata is the check would be first made for "wms_compliance_mode" and if missing then "ows_compliance_mode". 
"ows_compliance_mode" could then be set once in the Mapfile to apply to WFS, WMS, WCS etc.

jratike80:
> There's maybe missing a link here - what old closed issue are you referring to?

I meants this link from my previous comment https://github.com/INSPIRE-MIF/helpdesk-validator/issues/171

> Quite on the contrary I'm sure my changes definitely will break existing applications that are sending wrong parameter values for TRANSPARENT. But I assume this could be dealt with with a sensible release policy and mentioning in the release notes.

Such release policy is ok when we have security fixes and real improvements which are backwards incompatible, but implementing the fixes into client applications pays money, and in this case the change is cosmetic. Money for nothing, I would say.

For comparison, we have made a similar change when we started to require `&STYLES=` in Mapserver 8 because it is mandatory. This fails nowadays with an error message:

https://demo.mapserver.org/cgi-bin/msautotest?SERVICE=wms&VERSION=1.3.0&REQUEST=GetMap&LAYERS=world_latlong&FORMAT=image/jpeg&BBOX=-90,-180,90,180&WIDTH=720&HEIGHT=360&CRS=EPSG:4326

```
<ServiceException code="MissingParameterValue"> msWMSLoadGetMapParams(): WMS server error. Missing required parameter STYLES. Note to service administrators: defining the "wms_allow_getmap_without_styles" "true" MAP.WEB.METADATA item will disable this check (backward compatibility with behavior of MapServer < 8.0) </ServiceException>
</ServiceExceptionReport
```

Set empty "STYLES" and request succeeds:

https://demo.mapserver.org/cgi-bin/msautotest?SERVICE=wms&VERSION=1.3.0&REQUEST=GetMap&LAYERS=world_latlong&FORMAT=image/jpeg&BBOX=-90,-180,90,180&WIDTH=720&HEIGHT=360&CRS=EPSG:4326&STYLES=

However, the request should yield an error because `&SERVICE=wms` is wrong. The WMS standard says:

> ...the SERVICE parameter of the request shall have the value “WMS” 
> ...
> Parameter names shall not be case sensitive, but parameter values shall be. 

The INSPIRE validator is nonchalant when it comes to this violation and I think it is good so.

+1 for `I'd be very much in favour of adding in a compliant WMS mode, with a test`

tschorr:
@jmckenna @jratike80 @geographika I've added a test. As for adding the `ows_compliance_mode`, I'm waiting a bit for the discussion to settle/more reactions/a decision before I continue :-)

jmckenna:
great, thanks @tschorr

src/mapwms.cpp
```diff
@@ -1289,6 +1289,13 @@ int msWMSLoadGetMapParams(mapObj *map, int nVersion, char **names,
       map->imagetype = msStrdup(values[i]);
     } else if (strcasecmp(names[i], "TRANSPARENT") == 0) {
       transparent = (strcasecmp(values[i], "TRUE") == 0);
+      if ((!transparent) && (strcasecmp(values[i], "FALSE") != 0)) {
```
jratike80, line 1292: By the WMS standard, using strcasecmp for the value is WRONG. Or am I right that it ignores case in comparison? Accepted values are `TRUE` of `FALSE`. I hope that the INSPIRE validator does not check that, though.

tschorr: There's this statement in section 6.8.1: "Parameter names shall not be case sensitive, but parameter values shall be.". BUT there's also section 6.5 on boolean values having lowercase "true" and "false" (and also allows 1 and 0, not currently checked by MapServer at all). So yes it is confusing.

In any case the INSPIRE validator does not check that (yet).

jratike80: That boolean values section deals with other things. The GetMap parameters are defined in Table 8 — The Parameters of a GetMap request `TRANSPARENT=TRUE|FALSE`.
For my mind not checking the case is a good thing, but accepting `&TRANSPARENT=true` is undoubtedly against the WMS standard.

geographika (approved):
Very nice addition - thanks @tschorr

geographika:
@tschorr I'll try and get the CI updated to use the new version of GDAL so everything is passing again, and you can rebase this on top. 
Out of interest, are you testing MapServer against CITE tests? Getting WMS compliancy was on the development roadmap but has been stalled for some time now.

tschorr:
> Out of interest, are you testing MapServer against CITE tests? Getting WMS compliancy was on the development roadmap but has been stalled for some time now.

@geographika we are using our own installation of the [INSPIRE Reference validator](https://inspire.ec.europa.eu/validator/home/index.html) for testing a couple of WMS/WFS services. In theory this includes OGC WMS compliance testing but we are still evaluating our test setup. So far this issue is the only test failure we could attribute to MapServer.

PR #7282, opened 2025-05-21
Shapefile: Parse CPG values

PatrikSylve:
Continuation on https://github.com/MapServer/MapServer/pull/7274. 

Will parse CPG values in the same fashion as [gdal](https://github.com/OSGeo/gdal/blob/release/3.11/ogr/ogrsf_frmts/shape/ogrshapelayer.cpp#L503). This will help iconv to recognize code pages on the format as listed in: https://support.esri.com/en-us/knowledge-base/read-and-write-shapefile-and-dbase-files-encoded-in-var-000013192 

I added a new autotest together with a bunch of shape files in different encodings. The test is based on the file `encodings_utf-8.shp` which contains points with characters that are valid in specific encodings. I then used `ogr2ogr` to encode it in the different encodings. 

In the mapfile, each layer/encoding will have a similar `EXPRESSION` to make sure Mapserver will decode the attributes properly: 

> EXPRESSION "ÀÁÂÃÄÅÆÇÈÉÊËÌÍ....


jmckenna:
Thanks @PatrikSylve ! (and thanks for including a test)

PR #7284, opened 2025-05-22
INSPIRE compliance: Fix service exception status code when the SERVICE parameter is invalid/missing

tschorr:
[INSPIRE WMS Tests](https://github.com/inspire-eu-validation/ets-repository/blob/28b66265aab8fc8228672283887475818aade1ab/service/vs-wms-soapui-project.xml#L501-L552) expect the service to return a service exception with HTTP status code 200 for the case when the `SERVICE` parameter value is missing or invalid. Although there are currently several possibilities for returning errors in this case, none passes the INSPIRE validator tests:

* configuring `wfs_cite_wfs2` will return an XML service exception with HTTP status code 400
* using `MS_MODE = OWS` will return an HTML error message with HTTP status code 200

(Imho both settings are a bit obscure and maybe could be deprecated).

This PR introduces a service exception according to the CITE/INSPIRE specifications with HTTP status code 200 if `[ows|wms]_compliance_mode` is set to `true`. It's actually mimicking the behaviour of Geoserver for this error case.

This makes the above INSPIRE test pass without producing additional WFS test failures, for me at least.

This PR is based on #7280 so should be applied after merging #7280.


tschorr:
Tests are passing, except for the AppVeyor build that has been cancelled and maybe needs restarting.

geographika (approved):

PR #7286, opened 2025-05-22
Increase default memory for vagrant and remove legacy fix

geographika:
The default 2GB of RAM for vagrant makes builds very slow. It can be overwritten with environment variables, but making the default 4GB makes it run through msautotests much faster. 

The virtualbox-fix.sh no longer seems necessary (or have an effect). VBoxGuestAdditions is now up to version 7.x now (4.3 is referred to in the script). 

Fix added to make it fully automated in https://github.com/MapServer/MapServer/commit/29be7f091310cb7a2f343e15e344dd3e21e352ed

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -2,6 +2,7 @@
 # vi: set ft=ruby :
 
 require 'socket'
+require 'fileutils'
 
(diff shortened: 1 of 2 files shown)


jmckenna:
thanks @geographika !

PR #7295, opened 2025-06-21
[WIP] OGC API/OpenAPI: add vendorSpecificParameters for items paths

tomkralidis:


geographika:
Thanks @tomkralidis! Is it possible to add a test to illustrate its usage?

Link to relevant section in the spec: https://docs.ogc.org/is/17-069r3/17-069r3.html#query_parameters

tomkralidis:
@geographika not sure.  The OpenAPI update means that unknown parameters would be allowed.  I would then suspect that requests like https://github.com/MapServer/MapServer/blob/main/msautotest/api/ogcapi.map#L17 would **not** result in a 400 (https://github.com/MapServer/MapServer/blob/main/msautotest/api/expected/ogcapi_collections_mn_counties_items_unknown_parameter.json.txt).

geographika:
As discussed with @tomkralidis:

When the above code is added, then sending additional querystring parameters to the OGC API requests will be compliant. This will make `oga_compliance_mode` redundant, as the only checks this makes is for additional parameters. `oga_compliance_mode` could then be removed. 

To check this https://cite.opengeospatial.org/teamengine/test.jsp will be used to ensure compliancy.

geographika:
Thanks @tomkralidis. 

As discussed in https://lists.osgeo.org/pipermail/mapserver-dev/2025-September/017280.html additional parameters will be useful for authorization.

PR #7297, opened 2025-06-25
Check wms_onlineresource for OpenLayers output

geographika:
When using the [OpenLayers output](https://mapserver.org/cgi/openlayers.html) for a WMS requests (by adding `&FORMAT=application/openlayers`), the "ows_onlineresource" is used as part of the OpenLayers template:

https://github.com/MapServer/MapServer/blob/e6eba841ae2c4fc92061ac11fa01c7595cbb492d/src/maptemplate.c#L102

However, currently only `ows_onlineresource` is checked for in the metadata. This PR also checks for the  `wms_onlineresource` WEB METADATA key.


jmckenna:
thanks @geographika !

Issue #5091, opened 2015-04-22
FriBidi


sbrunner:
Two question or issue on FriBidi:

On this line:
https://github.com/mapserver/mapserver/blob/branch-7-0/CMakeLists.txt#L105
FriBidi is relay for **left-to-right** languages, not for **right-to-left** languages ?

On this command:
https://github.com/mapserver/mapserver/blob/branch-7-0/scripts/vagrant/mapserver.sh#L26-L31
`WITH_FRIBIDI` is set on tine to 0 and an other to 1 ...

tbonfort:
fixed, thanks

PR #7303, opened 2025-06-28
Update Typo in INSTALL.CMAKE

rockinrobstar:
Fix typo previously fixed in CMakeLists.txt but missed here. See #5091


jmckenna:
thanks @rockinrobstar !

Issue #7177, opened 2024-10-22
OGC API Feature returns complex geometries destroyed with many geometry errors


enatdvmv:
OGC API Feature returns complex geometries destroyed with many geometry errors [Verwaltungsgrenzen MV: Ämter](https://www.geodaten-mv.de/dienste/dvg_laiv_oga/ogcapi/collections/aemter/items?f=json&limit=150). The FME GeometryValidator (all) detects 2.401 geometry errors for this data set.

MapServer WFS 2.0.0 returns the feature without errors [Verwaltungsgrenzen MV: Ämter](https://www.geodaten-mv.de/dienste/dvg_laiv_wfs?SERVICE=WFS&REQUEST=GetFeature&VERSION=2.0.0&Typenames=dvg:aemter), pygeoapi 0.19.0. also.

The map-files for the WFS 2.0.0 and OGC API Feature are 99% identical.

## Operating system and MapServer version
Debian bookworm 12.7 with MapServer 8.0.0-3+b8 and identical errors with
Debian trixie with MapServer 8.2.2-1.

geographika:
@enatdvmv - do you have an example of a complex geometry? Is it supported in GeoJSON?
The GML used by the WFS 2.0.0 request likely supports geometries not available in GeoJSON.

enatdvmv:
@geographika
Here is a multipolygon with correct geometry as a response from pygeoapi and with destroyed geometry as a response from Mapserver.
[out_5353_mapserver.json](https://github.com/user-attachments/files/17619395/out_5353_mapserver.json)
[out_5353_pygeoapi.json](https://github.com/user-attachments/files/17619396/out_5353_pygeoapi.json)

geographika:
@enatdvmv - this looks to be due to different precisions used for the geometry. MapServer defaults to 6 decimal places for the OGC Features API. You can increase this to match pygeoapi by adding the following to your Map file in the WEB METADATA section:

```
  WEB
    METADATA
...
      "oga_geometry_precision" "13"
    END
  END
```

enatdvmv:
@geographika 
Tested with OGA_GEOMETRY_PRECISION "15" without any significant changes.

And MapServer WFS 2.0.0 extended with the output format geojson (DRIVER "OGR/GEOJSON"), returns all geometries in the geojon response without errors.

jratike80:
[...]

PR #7305, opened 2025-07-02
OGC API: fix issue #7177 (feature geometry of multipolygons is returned with errors)

MAX13666:
While working with MapServer we noticed that the OGC API would return weird multipolygon geometries.
As explained by @jratike80 on a comment in issue #7177, the OGC API returns multipolygon geometry coordinates "stacked" in a single element (and then multiplied for the number of different polygons present in the original) instead of separate.
We also noticed that when presented with a multipolygon with an inner ring, the OGC API result seems to completely ignore it, instead returning a polygon without the "hole".

Here are some images and GeoJSON files as an example:

This is the original multipolygon in QGIS:
image

This is how the multipolygon looks in the OGC API view in MapServer - as you can see the holes are gone:
image

This is how the multipolygon looks in MapServer, with the fix made by this pull request:
image

Here are the GeoJSON files for the three polygons, for comparison:
[original.json](https://github.com/user-attachments/files/21016238/original.json)
[OGCAPI-broken.json](https://github.com/user-attachments/files/21016245/OGCAPI-broken.json)
[OGCAPI-fixed.json](https://github.com/user-attachments/files/21016246/OGCAPI-fixed.json)

This behavior does not happen with WFS - the multipolygon geometry is returned correctly.

This problem is caused by a small typo in `mapogcapi.cpp`, which this PR fixes.


geographika:
Thanks for this @MAX13666. Are you able to add a test to msautotest using your JSON example?

ElPerenza:
Hi, I worked with @MAX13666 on this. I added a quick test and it seems to work fine.

geographika:
Thanks @MAX13666 and @ElPerenza for reporting and fixing this.

jmckenna:
thanks @MAX13666 and @ElPerenza, great catch!

PR #7309, opened 2025-07-16
Style: make fallback configurable in CLASS

PatrikSylve:
When applying a SLD style containing 'Else'-filter, the Else-class  will be marked as 'fallback' in Mapserver to ensure this class is used if and only if no other class is applicable. 

This MR enables the fallback property to be configured in the mapfile using:

```
CLASS
...
FALLBACK TRUE
END # CLASS
```

This makes it possible to reproduce the SLD behavior using `FALLBACK`  together with `RENDERMODE=ALL_MATCHING_CLASSES`. 

I also updated the `writeClass` method to write `FALLBACK` and the rendermode directive to keep the SLD behavior intact when mapscript is used to produce mapfiles that have had SLD applied: 

```
map.applySLD(content)
print(map.convertToString())
```


jmckenna:
thanks @PatrikSylve (ignore the "Check CRLF" action error, there is a known problem with its Docker image, I may disable this action shortly)

PatrikSylve:
Realized I had missed to commit maplexer.c. Will fix

geographika:
@PatrikSylve - just to understand further - this PR exposes a property `fallback` that is already used in the codebase for SLD generation?
How is it currently set to `true` when generating SLD?

PatrikSylve:
> @PatrikSylve - just to understand further - this PR exposes a property `fallback` that is already used in the codebase for SLD generation? How is it currently set to `true` when generating SLD?

Correct. 

The `class.isfallback` is set when parsing an SLD ElseFilter, see https://github.com/MapServer/MapServer/blob/main/src/mapogcsld.cpp#L935. 

And the logic on how fallback class is applied is implemented in `msShapeGetNextClass`: 
https://github.com/MapServer/MapServer/blob/main/src/maputil.c#L830

geographika:
@PatrikSylve - thanks for the PR. It will need an accompanying documentation PR in https://github.com/MapServer/MapServer-documentation to the CLASS docs with the new keyword, and ideally https://mapserver.org/ogc/sld.html to get this merged.

geographika:
Unless there are any further comments or objections, I'll merge this in the next couple of days. Thanks @PatrikSylve!

PR #7318, opened 2025-08-08
WMS GetFeatureInfo: add options to precisely identify points through their symbols

rouault:
Currently, when MapServer receives a WMS GetFeatureInfo request on a point layer, it queries the vector data source on a square centered on the value of the X and Y parameters of the query (converted from image pixels to the layer coordinate reference system), with a radius expressed in the LAYER.TOLERANCE MapFile parameter (or its default values if not specified). This behavior works fine as long as symbols used to represent the point are of square shape, and with the symbol being centered on the point. When the symbol is of rectangular shape and/or with a anchor point far from its center, users may get unexpected results (false positives and false negatives) when clicking on features on the map.

This PR adds a new IDENTIFICATION block under LAYER that can be set to

```
  IDENTIFICATION
      CLASSAUTO
  END
```

to use the symbols from the style classes of the incoming WMS GFI request

or

```
  IDENTIFICATION
      CLASSGROUP "name_of_class_group"
  END
```

to use a precise style group when identifying point symbols.

Identification takes into account potential transparency in the symbol.

The existing TOLERANCE and TOLERANCEUNITS are taken into account. Note that for clarity of the mapfile, it is possible to move them under the IDENTIFICATION block.

Note: this enhancement is currently restricted to POINT layers.


jratike80:
Feels like a good improvement.
Do you check STYLE-MIN/MAXSCALEDENOM and make sure that the symbol shows in the corresponding GetMap result?

rouault:
> Do you check STYLE-MIN/MAXSCALEDENOM 

yes, I've added tests for that. That was already checked by the original msQueryByPoint() function

> and make sure that the symbol shows in the corresponding GetMap result?

Hard to answer 100% positive to that given identification and rendering are two different code paths. Checking further I saw that the identification code could potentially return non-drawn feature because of expression evaluation, when using a non-template code path. I've improved that.

rouault:
@sdlime @geographika  Any thought on the proposed mapfile syntax ?

geographika:
Very nice improvement!

Are there any other settings planned for an `IDENTIFICATION` block? Otherwise, would a `PROCESSING` option make more sense to set this for a layer? A whole new block for a single keyword seems a little excessive. 

Question on the `CLASSGROUP "name_of_class_group"` - is there a use-case for using a different `CLASSGROUP` from the active one (set in the Mapfile or WMS `STYLES=myclassgroup`)?

rouault:
> Are there any other settings planned for an `IDENTIFICATION` block?

not yet, but for example I can imagine we could tweak identification to accept transparent pixels inside the symbol. Having a new block seemed cleaner to me than a processing option

> Question on the CLASSGROUP "name_of_class_group" - is there a use-case for using a different CLASSGROUP from the active one 

I can imagine someone could create a symbol that is larger than the displayed one for the purpose of easier clicking

geographika:
As I understand it, OGC Maps API has no equivalent for `GetFeatureInfo` and requests should be through the OGC API – Features API. If this is the case, then maybe this functionality would be better exposed using `METADATA` and `wms_` keywords?
Possibly worth taking to the dev mailing list for other views?

The disadvantage of a new `IDENTIFICATION` block would be there no inbuilt MapScript support (although should be fine to add if required). It also would need a `writeIdentification` function to dump the block back out to a file like the other keywords.

rouault:
> As I understand it, OGC Maps API has no equivalent for `GetFeatureInfo` and requests should be through the OGC API – Features API. If this is the case, then maybe this functionality would be better exposed using `METADATA` and `wms_` keywords?

I'm not sure to follow you. This identification is really tied to map rendering and styling. That cannot be cleanly emulated using OGC API Features (or WFS). What would be the advantage of moving under METADATA wms_ keywords ? That isn't really metadata.

> Possibly worth taking to the dev mailing list for other views?

will do

>  It also would need a writeIdentification function to dump the block back out to a file like the other keywords.

ah yes, will do

geographika:
> > As I understand it, OGC Maps API has no equivalent for `GetFeatureInfo` and requests should be through the OGC API – Features API. If this is the case, then maybe this functionality would be better exposed using `METADATA` and `wms_` keywords?
> 
> I'm not sure to follow you. This identification is really tied to map rendering and styling. That cannot be cleanly emulated using OGC API Features (or WFS). What would be the advantage of moving under METADATA wms_ keywords ? That isn't really metadata.

As I understand it, the functionality in this PR can only ever be used as part of a WMS `GetFeatureInfo` request. It can't be used through the MapServer CGI functionality, nor any (as yet not implemented) MapServer OGC Maps API. So could it make sense to have it as a WMS layer config option, handled similar to the other WMS layer specific settings in https://mapserver.org/ogc/wms_server.html#layer-object-metadata

rouault:
> So could it make sense to have it as a WMS layer config option,

I see your point. Obviously it could. That said maybe some day there will some kind of GetFeatureInfo for OGC API maps. Anyway I'm going to bring that to the mailing list

rouault:
PR updated to:
- rename IDENTIFICATION to IDENTIFY
- added writeIdentify
- tested successfully a mode=query request

rouault:
documentation added in https://github.com/MapServer/MapServer-documentation/pull/1035

PR #7324, opened 2025-08-19
PostGIS: Escape geometry column name consistently

jnystad:
There is a few places where geometry column name is not escaped. This fixes that for `msPostGISLayerGetExtent` and `msPostGISBuildSQLWhere`.

Not sure if `msPostGISLayerTranslateFilter` `case MS_TOKEN_BINDING_SHAPE` should do this as well, though. Any advice?


jmckenna:
thanks @jnystad !

PR #7328, opened 2025-08-29
WFS: add next link if there might be features left

PatrikSylve:
# Summary 
In some cases, GetFeature responses will contain less features than expected. This PR will add `next`-links in such cases to make all features discoverable through WFS-paging.
# Problem

When requesting GetFeature with pagination and BBOX in another SRS than the layer source, the pagination currently reports less results than expected in certain cases. 

I have added one such case in the autotests, and it is visualised in the image below.
  
Screenshot from 2025-08-26 17-26-13

* The purple polygon `bbox4326` is the requested BBOX in EPSG:4326
* The pink rect is the bounds of the requested BBOX in the source projection, EPSG:3978
* The points is the `popplace`-data, where those in green are within requested BBOX.
* The map projection is in source projection

Since the number of points within requested BBOX is 7, a query with `COUNT=7&STARTINDEX=0` is expected to return all 7 matches. Currently it will only report 4 hits and no `next`-link, indicating that there are no more features.

The filtering done in  `msQueryByRect` will first call `msLayerWhichShapes` with the BBOX in source SRS and if paging is enabled, it will limit the result to `COUNT` or `maxfeatures`-limit. If a feature is filtered out in a later stage, the result count will be less than `maxfeatures`-limit, and MapServer will report that there are no more features left.  

## Solution
This PR does not solve original issue, but will add `next`-link if there may be any more features of interest, making the features discoverable. 

To keep track whether a `next`-link might need to be added, I added a `hasnext`-flag to `resultCacheObj`. This will be marked as true if there may be features of interest in the next page. I.e., if any feature has been filtered out, and the total number of features (including skipped ones) equals to the amount of requested features. 

One downside of this is that the user may in some cases need to page through pages containing 0 features. 

## Other alternatives

I am not sure what would be the best way to solve the original issue.  I'd be happy to hear if anyone has an idea. Some ideas I can think of:

### Skip paging in first/driver filtering

If paging was removed in the `msLayerWhichShapes`-step, and handled after when iterating the features, the problem would be solved. But I guess this would not be optimal for large datasets. 

### Support query by shape

If `msLayerWhichShapes` were to support query by a shape, we could just filter on a shape in source coords, matching the requested BBOX. If I am not wrong, it currently only supports filtering by BBOX. 

### Internal pagination

I looked a bit into doing some internal pagination and storing feature indexes in between pages. But if there are no featureIds, calling `msLayerWhichShapes` multiple times may assign 'random'(?) indexes to features, so seems not that reliable. Could this be done another way?

### DIFF
--- a/msautotest/wxs/expected/wfs_ogr_gpkg_pagination_4326.xml
+++ b/msautotest/wxs/expected/wfs_ogr_gpkg_pagination_4326.xml
@@ -1,0 +1,148 @@
+Content-Type: text/xml; subtype="gml/3.2.1"; charset=UTF-8
+
+<?xml version='1.0' encoding="UTF-8" ?>
+<wfs:FeatureCollection
+   xmlns:ms="http://mapserver.gis.umn.edu/mapserver"
+   xmlns:gml="http://www.opengis.net/gml/3.2"
+   xmlns:wfs="http://www.opengis.net/wfs/2.0"
+   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+   xsi:schemaLocation="http://mapserver.gis.umn.edu/mapserver http://localhost/path/to/wfs_simple?SERVICE=WFS&amp;VERSION=2.0.0&amp;REQUEST=DescribeFeatureType&amp;TYPENAME=popplace&amp;OUTPUTFORMAT=application%2Fgml%2Bxml%3B%20version%3D3.2 http://www.opengis.net/wfs/2.0 http://schemas.opengis.net/wfs/2.0/wfs.xsd http://www.opengis.net/gml/3.2 http://schemas.opengis.net/gml/3.2.1/gml.xsd"
+   timeStamp="" numberMatched="4" numberReturned="4"
+   next="http://localhost/path/to/wfs_simple?SERVICE=WFS&amp;VERSION=2.0.0&amp;REQUEST=GetFeature&amp;TYPENAMES=popplace&amp;BBOX=43.0%2C-65.0%2C45.0%2C-62.0&amp;COUNT=7&amp;STARTINDEX=7">
+      <wfs:boundedBy>
+      	<gml:Envelope srsName="urn:ogc:def:crs:EPSG::4326">
+      		<gml:lowerCorner>44.020359 -64.732816</gml:lowerCorner>
+      		<gml:upperCorner>44.971547 -62.513078</gml:upperCorner>
+      	</gml:Envelope>
+      </wfs:boundedBy>
+<!-- WARNING: No featureid defined for typename 'popplace'. Output will not validate. -->
+    <wfs:member>
+      <ms:popplace>
+        <gml:boundedBy>
+        	<gml:Envelope srsName="urn:ogc:def:crs:EPSG::4326">
+        		<gml:lowerCorner>44.911800 -62.513078</gml:lowerCorner>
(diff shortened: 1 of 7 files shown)


geographika:
@PatrikSylve - thanks for this PR and the detailed notes. 

Without considering possible implications or difficulty to implement, the option "Support query by shape" sounds the cleanest solution. 

On the last point, more and more features such as OGC API Features require a unique ID for each feature to work, so it doesn't seem unreasonable to expect WFS services requiring paging to require this too. 

As an aside, I'm not sure why the Windows CI is broken on this PR and working on others. I've tried rerunning a couple of times, but errors seem unrelated to any changes.

jratike80:
I have no exact recommendations, just sharing what I know about paging.
By my experience with GeoServer, StartIndex and Count tend to be unreliable without sortBy. Without sortBy it is not guaranteed that features are coming every time in the same order from the datasource. For really consistent result sets the attribute used in sortBy should be unique. WFS standard allows also a non-consistent paging, but as a user I would not like such behavior:

> 7.7.4.4.2.3 Response paging without transactional consistency 
> Servers that do not declare transactional consistency shall not be required to maintain transactional 
> consistency or state between paging iterations. Thus it is possible for new features to be added, updated or 
> removed from the complete result set between iterations. As a consequence it is possible to have a result set 
> element be skipped or duplicated between iterations.

Discussion about if server shoud sort automatically when paging is requested
https://discourse.osgeo.org/t/geoserver-devel-wfs-and-paging/95933/11
Discussion about sorting making queries slower
https://discourse.osgeo.org/t/geoserver-devel-performance-versus-conformance-in-wfs-2-0-paging/97181/12

I think that GeoServer now makes sorting automatically when the request contains `&startIndex`. This has a side effect. If there is a server side max page size, let's say 1000, then 1000 features returned by plain GetFeature are not necessarily the same than the 1000 features returned when `&startIndex=0`. That can make a surprise for a GeoServer user who wants to get a big dataset and makes GetFeature first without startIndex, and after notising that received only 1000 features, asks more by adding `&startIndex=999`.

ejn:
@jratike80 I think this is a problem of user education: Without SORTBY there is no guaranteed sorting of a result, and nor should there be any such expectation (cf. SQL: without ORDER BY the sorting is theoretically arbitrary, although may be influenced by the internal storage order of the actual records). Therefore if there is any chance that you will be wanting to page results then you must use the identical SORTBY for all requests, including the first.

This could of course cause a problem with offering a "next" link in the results, because without some sorting on the initial request then the next link cannot reliably truly link to the next results (guaranteeing no missing or duplicate results) if there is not either an implicit or explicit stable sorting.

From my point of view a "next" link should therefore only be offered when one of the following is true:
- sorting is internally stable (this may be always true for some drivers, or could be applied at the layer level by defining a default sorting)
- sorting is explicitly specified in the request (i.e. a SORTBY parameter was applied to the request)

ejn:
just to clarify my previous comment: if a user makes a mistake with `sortby`/`startindex` and doesn't get all features and/or gets some duplicates then I think that is their problem, but if "we" supply a `next`-link (or rather a series of `next`-links) which lead to the user not getting all features and/or getting some duplicates then I think that is "our" problem, so "we" should only supply a `next` link in cases where we can guarantee that it is safe to do so

jmckenna:
@PatrikSylve there are some conflicts in mapquery.c to resolve here for this pull request.

jmckenna:
great, thanks @PatrikSylve !

PR #7331, opened 2025-09-02
Update SDTS test to use SXF driver

geographika:
The SDTS driver was removed in GDAL 3.11 (see https://github.com/OSGeo/gdal/pull/11758/commits/9a43fb590166c47ae663b5952b1c61a33df323c5). The test looks like it was already returning an error rather than GetCapabilities response, so removing. 

Now failing test was due to a different error message being returned with the GDAL update:

```bash
 --- ../expected/wfs_ogr_drv_nocreatedatasource_caps.xml	2025-09-02 08:16:24.351347560 +0000
+++ wfs_ogr_drv_nocreatedatasource_caps.xml	2025-09-02 08:22:32.237125642 +0000
@@ -4,5 +4,5 @@
 <BODY BGCOLOR="#FFFFFF">
 loadOutputFormat(): General error message. OUTPUTFORMAT (nocreatedatasource) clause references driver (OGR/SDTS), but this driver isn&#39;t configured.
-msInitDefaultOGROutputFormat(): General error message. OGR `SDTS&#39; driver does not support output.
+msInitDefaultOGROutputFormat(): General error message. No OGR driver named `SDTS&#39; available.
 </BODY></HTML>
```


rouault:
I believe the purpose of this test is to check behavior when specifying a OGR driver without creation capabilities . So instead of removing it completely, I'd witch to the SXF driver that is read-only and always compiled.

rouault (approved):

PR #7333, opened 2025-09-03
Fix use-after-free error in msDrawRasterLayerLowOpenDataset

marisn:
1. Allocation (`mapraster.c:649`): The function `msDrawRasterLayerLowOpenDataset` calls a GDAL function       (`CSLTokenizeString2`) which allocates a block of memory to hold some tokenized strings.
2. De-allocation / Free (`mapraster.c:654`): A few lines later, the code frees that same block of memory by       calling `CSLDestroy`.
3. Invalid Read (`mapraster.c:665-666`): A few lines after that, the code tries to read from the memory it just freed. It performs a string comparison (strcasecmp) on a pointer that is now invalid.

Debugging assistance by gemini-2.5-pro


marisn:
A workaround is to set DEBUG to 0 in MAP file;  MS_DEBUGLEVEL environmental variable et al.

marisn:
Affected versions: >= 8.2.0
Source: https://github.com/MapServer/MapServer/commit/f20110ee1d748e7e7d274ca28fde829a63475f9a

geographika:
@marisn - is there any way to add a test for this? Anything that runs the code should trigger the memory leak that in turn would have caused the CI to fail (so that code path must currently be untested).

marisn:
> @marisn - is there any way to add a test for this? Anything that runs the code should trigger the memory leak that in turn would have caused the CI to fail (so that code path must currently be untested).

It is easy to trigger – you need a WMS source and DEBUG > 0. As I'm on a vacation, I don't know when weather will be bad enough to create a test case ;-)

marisn:
Added a test case that crashes without this PR and passes with it.

geographika:
Thanks @marisn !

PR #7334, opened 2025-09-03
Fix(parser): Prevent crash with non-existent layer in REQUIRES

marisn:
When a REQUIRES or LABELREQUIRES expression contains a reference to a layer that does not exist in the mapfile, the expression parser would treat it as an attribute binding. Since msEvalContext is called without a shape object, this would lead to a NULL pointer dereference and a crash when the parser tried to evaluate the binding. This patch adds a check after tokenizing the expression to ensure no attribute binding tokens are present. If any are found, it indicates a reference to a non-existent layer. An error is reported and the evaluation returns MS_FALSE, preventing the crash.

Fix generated by gemini-2.5-pro


geographika:
@marisn - good catch. I can recreate the error by adding either of the following:

```
LABELREQUIRES "![fake]"
REQUIRES "[fake]
```

Would you be able to update the test map at `mapserver/msautotest/query/context.map` to include some missing options as part of the PR?

marisn:
> Would you be able to update the test map at `mapserver/msautotest/query/context.map` to include some missing options as part of the PR?

Just hunting down crashes in production :-(
I am on a vacation thus have no idea when I'll be able to.

geographika:
@marisn `REQUIRES` has been around since 2003, so likely the bug too. I think you can finish your vacation before worrying about a test case :-)

marisn:
@geographika tests I created fail without this PR and pass with it. Please take a look. We can discuss @ code sprint if more changes are needed.

PR #7101, opened 2024-07-09
install OGCAPI html-bootstrap4 templates

jmckenna:
- related to issue #7100

PR #7338, opened 2025-09-07
Add Bootstrap 5 OGC API template files

geographika:
This PR adds a new `html-bootstrap5` folder containing a [set of template files](https://mapserver.org/ogc/ogc_api.html#template-handling) for the OGC Features API. I have added this as a new folder rather than updating the current files as the folder is named `html-bootstrap4`. We could rename this `html-bootstrap` and do continual updates to these files, but that could break existing installations. 

The new templates are updated copies of the current `html-bootstrap4` templates (with an almost identical look), with the following changes:

- updates to Bootstrap 5.3.8 (from 4.5.3)
- uses the 
- updates to Leaflet 1.9.4 (from 1.7.1)
- updates to Swagger UI 5.28.1 (from 3.51.0)
- removes the JQuery dependency completely (no longer required for Bootstrap, and plain JS now used in other cases)
- uses the CDN `bootstrap.bundle.min.js` avoiding the custom https://files.dnr.state.mn.us/lib/bootstrap4/javascripts/bootstrap.min.js which included `poppr.js` as this is part of the bundle)
- all CDN links remove tracking and check for integrity (as suggested by cdnjs.cloudflare.com) to ensure the files haven't been modified by a rogue CDN
- removed `-moz-scrollbars-vertical` (only required for Firefox pre 2018)


geographika:
Deployment of new template files also discussed in https://github.com/MapServer/MapServer/pull/7101

rouault:
> Deployment of new template files also discussed in #7101

could we add an option like the one I suggested in https://github.com/MapServer/MapServer/pull/7101#issuecomment-2218571276 so they get installed ?

geographika:
Renamed the folder to `html-bootstrap` to avoid pinning to a specific version, and added files to the install. This can be turned off using the new `INSTALL_HTML_BOOTSTRAP` option (as suggested by @rouault). 

See related discussions on the mailing list at https://lists.osgeo.org/pipermail/mapserver-dev/2025-September/017256.html

geographika:
cc @sebastic - note the new `INSTALL_HTML_BOOTSTRAP` build option for 8.6 release.

Issue #7341, opened 2025-09-09
MapServer and Int32 Rasters


geographika:
I've been working with an Int32 raster containing large IDs as cell values and run into some strange off-by-one or off-by-two results.

`114140618` becomes `1.1414062e+08` (`114140620`) with a WMS GetFeatureInfo query, but becomes `114140616` as a raster label. See image below, with the green vector lines labelled with the correct values. The raster values are correct in the data source and when working in QGIS. 

`112667633` becomes `112667630` in a GetFeatureInfo  query, and `112667632` in raster labels.

<img width="575" height="380" alt="Image" src="https://github.com/user-attachments/assets/1a2238cf-3271-48d2-9897-572bf959f0f1" />

I have a few negative values in the raster  (all < -90735 and the Float32 limit of 16777216) and these all display correctly and return correct GetFeatureInfo results.

It looks like MapServer is casting to Float32 values in several places, which causes the issue. 

I'll upload a test case and debug further, but there are many casts to Float32 in the raster code paths, so maybe this is a known issue?

geographika:
I was hoping these were both due to the same issue and could be resolved with one fix. It seems it is the same issue (use of Float32 truncating Int32 values), but different code paths and therefore 2 different fixes required.

jmckenna:
(adding to the 8.6.0 milestone)

geographika:
Fixed with #7343 and #7344.

PR #7342, opened 2025-09-09
Int32 raster test cases

geographika:
Test cases and data for #7341. Increasing to `%.10g` allows for Integers to be output to the GeoJSON (although the value is off-by-two).


geographika:
Will split this into separate PRs. See #7343.

PR #7358, opened 2025-10-06
Correct implementation of "msAlphaBlendPM". Bands are already pre-mul…

ThomasCusson:
At our company we noticed, using the composite "src-over" blending mode (default), that some darkening is happening when opacity is not 0% or 100%. This is a typical problem of pre-multiplied alpha, so I looked into it.

It seems that the `msAlphaBlendPM` (PM stands for Pre-Multiplied) acts as if the bands were not pre-multiplied already, in a fashion similar to `msAlphaBlend`.
It results in a double multiplication by alpha channel, producing colors that are too dark on half-transparent areas. This issue probably arise only when using a change of projection, as it is only called in `mapresample.c`.

I also used "256" values instead of 255, to match the implementation from `msAlphaBlend`, but maybe it's not useful, I'll leave that up to your judgement.


Guillaume0789:
Hey ! 
I work with Thomas and here a screenshot of the problem we encountered. 
weird effect between images

jmckenna:
thanks @ThomasCusson .  Please also include a test for this as part of your pull request (you can see a few other COMPOP tests in /msautotest/renderers/).  Find more info about msautotest at https://mapserver.org/development/tests/autotest.html

PR #7360, opened 2025-10-10
Add ows_extra_params for OGC Features API

geographika:
This PR adds a new `ows_extra_params` metadata entry allows you to append additional query parameters to all URLs generated by MapServer for the OGC Features API. This can be used, for example, to include authentication tokens such as JSON Web Tokens (JWT) in your API requests.

Parameters will be persisted to all URLs includes in the API JSON responses. Parameters can be set at the `WEB` level and overridden at the `LAYER` level if required. 

See initial discussions at https://lists.osgeo.org/pipermail/mapserver-dev/2025-September/017280.html

The HTML templates (plain and Bootstrap) have all been updated to include these parameters in their URLs, and `{{ template.extra_params }}` is available in any custom templates.  

Documentation preview available at https://geographika.github.io/MapServer-documentation/ogc/ogc_api.html#web-object-metadata

PR #7363, opened 2025-10-17
Remove legacy bootstrap4 files

geographika:
With the addition of extra parameters added in #7360, and the updated Bootstrap 5 files added in #7338, leaving the old templates will likely cause confusion. I propose these should be removed for the 8.6 release.


jmckenna:
@geographika makes sense to me.  thanks!

Issue #7364, opened 2025-10-17
Query maps with MS_NORMAL and MS_HILITE styles fail with WFS layers


szekerest:
Query maps with MS_NORMAL and MS_HILITE styles fail with WFS layers. The returned error is:

msWFSLayerWhichShapes() Got HTTP status 0 downloading WFS layer [layername].

This is because msWFSLayerWhichShapes is being executed on a copy of the layer which is not added to the map. 
Therefore, msOWSExecuteRequests will not be able to find that layer from the map->layers collection for updating the layer status with the HTTP result.

PR #7365, opened 2025-10-17
Fix query maps with MS_NORMAL and MS_HILITE styles with WFS layers (#7364)

szekerest:


jmckenna:
thanks @szekerest.  Do you think it is possible to add a test for this in msautotest?  (I notice some similarities in [tomk_workshop.map](https://github.com/MapServer/MapServer/blob/main/msautotest/wxs/tomk_workshop.map) & [rfc36.map](https://github.com/MapServer/MapServer/blob/main/msautotest/misc/rfc36.map) )

szekerest:
Probably a FILTER query would work against wfs_client_100.map. For me it is a bit challenging to create the expected result on Windows, but I'll see what I can do.

jmckenna:
true, yes, thanks (after I posted that I realized it was a WFS client layer that was needed).  thanks again.

jmckenna:
thanks @szekerest !

PR #7366, opened 2025-10-18
Process empty path requests with mapfiles in querystrings

geographika:
While testing the latest `main` branch in a few different environments, I noticed requests using the new `MS_INDEX_TEMPLATE_DIRECTORY` would return the index page for requests with `PATH_INFO` set to `/` but with a valid querystring containing a map e.g. `http://localhost:7000/?map=/etc/mapserver/lines.map&REQUEST=GetMap...`.

This PR should ensure the index page is only returned when no map is loaded, or no parameters are passed to MapServer.


jmckenna:
thanks @geographika.  Can you possibly include an msautotest for this?  (I imagine there should be a way to call this through localhost and check the response with curl etc). thanks again.

geographika:
@jmckenna - tests added. It was possible to recreate using msautotests.

jmckenna:
thanks @geographika !

PR #7368, opened 2025-10-27
WMS Client:  Validate style in wms_style metadata

PatrikSylve:
When a layer with `CONNECTIONTYPE=WMS` have the `wms_style`-metadata set, the style will be published in GetCapabilities as a valid style to request. But when specifying the style in a GetMap request, the following exception is returned: 

```xml
<?xml version='1.0' encoding="UTF-8" standalone="no" ?>
<ServiceExceptionReport version="1.3.0" xmlns="http://www.opengis.net/ogc" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/ogc http://schemas.opengis.net/wms/1.3.0/exceptions_1_3_0.xsd">
<ServiceException code="StyleNotDefined">
msWMSLoadGetMapParams(): WMS server error. Style (MyStyle) not defined on layer.
</ServiceException>
</ServiceExceptionReport>
```

I updated the validation to allow requests with the style specified in `wms_style`. 

I added some tests for it, but since the demo server does not publish any styles, it will pass the validation on client side and return error from remote in the image response.


geographika:
Thanks @PatrikSylve - looks good to me.

jmckenna:
@PatrikSylve thanks again, it looks good here!

Issue #7367, opened 2025-10-20
Layer extent for WMS client is overwritten when zooming in


szekerest:
Since this commit:

https://github.com/MapServer/MapServer/pull/7159/commits/a7ad17c911e4f0ab760eb7211c8bfb1344122516

When we zoom in a map, which have WMS client layer configured, the layer extent is set to the map extent unexpectedly. After that, when we zoom out again, the layer extent reflects the previous zoom, so not the entire layer is displayed on the map.

Probably, the author of the refered commit assumed that each rendering session is happening on a new map object, however there are several mapscript applications which keep the mapObj reference between the rendering sessions. In this case, setting the layer extent during rendering may cause such side effects.

More specifically, in mapwmslayer.c line 779 we can see:

`lp->extent = bbox;`

While the bbox is calculated as the intersection of the previous layer extent and the map extent in line 763:

`msRectIntersect(&bbox, &layer_rect);`

szekerest:
cc @PatrikSylve @jmckenna

PatrikSylve:
Hi, yes I did not consider that the reference to the layer would be kept between rendering sessions. I guess it is something like this that happens in your case? 

```python

map = mapscript.mapObj(map_file); 
map.setExtent(firstMapExtent)
# Layer will get updated extent, l.extent = intersection(l.extent, firstMapExtent)
map.draw() 

map.setExtent(secondMapExtent)
# Will draw intersection(l.extent, secondMapExtent) == intersection(intersection(l.orginalExtent, firstMapExtent), secondMapExtent)
map.draw() 
```

The reason for the change in the first place was to avoid reprojecting the layer extent into the SRS of the map, as this is not always possible to do. 

I think the best solution here would be to not touch the layer extent at all. And in that case, I think we likely need to handle the fact that the image linked to the layer is in mapObj projection and not the layers projection. https://github.com/MapServer/MapServer/blob/main/src/mapwmslayer.c#L1330

Here is the original PR that introduced the change: https://github.com/MapServer/MapServer/pull/7159/files

jmckenna:
@PatrikSylve thanks

@szekerest thoughts?

jmckenna:
@szekerest @PatrikSylve any thoughts on how to handle this issue?  (it would be good to have this included in the 8.6.0-beta, obviously)

PatrikSylve:
[...]

PR #7371, opened 2025-10-28
WMS Client: Do not modify layer extent #7367

PatrikSylve:
Fixes issue with layer extent being modifed in WMS client handling #7367

I removed cases where layer extent and projections were modified in `msBuildWMSLayerURL`. 

I added `pszEPSG` to `httpRequestObj` to store the SRS of the WMS request. This was added to handle cases where the request will be performed in another SRS than the layer. In such cases, the layer will be temporarily assigned the request projection and extent in `msDrawWMSLayerLow`. 

I added a test for mapscript that also verifies that the layer extent and projection is not modified.

### DIFF
--- a/msautotest/mspython/test_wms_client.py
+++ b/msautotest/mspython/test_wms_client.py
@@ -8,7 +8,6 @@ try:
 
     mapscript_available = True
 except ImportError:
-    print("no")
     pass
 
 pytestmark = [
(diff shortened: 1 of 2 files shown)


jmckenna:
thanks @PatrikSylve for all of this effort, it is appreciated!

Issue #7370, opened 2025-10-28
python bindings breakage with pip 25.3


landryb:
just being a messenger here, but apparently mapscript's python bits use a deprecated construct that is removed in pip 25.3, cf openbsd/ports@d991d63fbcf5 for a local fix

i can do a PR adding `--use-pep517` to the 3 pip install invocations in https://github.com/MapServer/MapServer/blob/main/src/mapscript/python/CMakeLists.txt but i'd like to make sure this is backwards compatible with the various pip versions that are used to build mapserver in the ecosystem...

geographika:
I'm not sure that I understand the changes in that patch - the `--use-pep517` is to force use of a `pyproject.toml` (which doesn't exist).
The PR title is *stop using the legacy setup.py bdist_wheel mechanism which was removed* which isn't used (`python -m build --wheel` is used for the build). 
The Python bindings would benefit from moving to `pyproject.toml` but I don't see a breaking issue at the moment.

sthen:
I didn't look into all the details, was just doing a quick build fix, but building with pip 25.2 gives this warning

```
Processing /pobj/mapserver-8.4.1/build-i386/src/mapscript/python
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'done'
Building wheels for collected packages: mapscript
  DEPRECATION: Building 'mapscript' using the legacy setup.py bdist_wheel mechanism, which will be removed in a future
 version. pip 25.3 will enforce this behaviour change. A possible replacement is to use the standardized build interfa
ce by setting the `--use-pep517` option, (possibly combined with `--no-build-isolation`), or adding a `pyproject.toml`
 file to the source tree of 'mapscript'. Discussion can be found at https://github.com/pypa/pip/issues/6334
```

...building with 25.3 as-is fails; building with 25.3 with --use-pep517 --no-build-isolation succeeds.

sthen:
...if I try with 25.3 it goes off to the network (which is blocked in openbsd ports builds) to try and install setuptools:

```
-- Install configuration: "Release"
Processing /usr/obj/ports/mapserver-8.4.1/build-amd64/src/mapscript/python
  Installing build dependencies: started
  Installing build dependencies: finished with status 'error'
  error: subprocess-exited-with-error
  
  × installing build dependencies did not run successfully.
  │ exit code: 1
  ╰─> [7 lines of output]
[...]

PR #7377, opened 2025-11-02
Disable build isolation to not download setuptools dependency.

sebastic:
The Debian package build failed with pip 25.3 without this change.

I'm not 100% convinced it belongs upstream, but other packagers are likely to run into the same issue.

### DIFF
--- a/src/mapscript/python/CMakeLists.txt
+++ b/src/mapscript/python/CMakeLists.txt
@@ -140,7 +140,7 @@ install(
     endif()
 
     execute_process(
-      COMMAND ${Python_EXECUTABLE} -m pip install \${PYTHON_ROOT} \${PYTHON_PREFIX} .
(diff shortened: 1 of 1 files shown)


landryb:
cf #7370 ?

sebastic:
The discussion in #7370 confirms that other packagers run into the same issue.

So I think we should at least merge this.

Adopting pyproject.toml can happen later.

landryb:
from my understanding, that means that having setuptools installed on the system is now a requirement, since it was previously installed by pip. I dont think that's a big deal. `--use-pep517` is another mistery to me :)

sebastic:
You already needed to have setuptools installed in the build environment, whether installed by the distribution package manager or by pip.

`--use-pep517` is not required at this time. Projects that still rely on `setup.py install` should adopt `pyproject.toml` because that functionality is scheduled to be removed from setuptools, but Python MapScript is not one of those projects.

jmckenna:
thanks @sebastic I agree this will help packagers.

Issue #7312, opened 2025-07-29
Error in oracle select, space missing


gkvoelkl:
Hello, 

In mapserver 8.4.0 I am getting this error message

Wed Jul 23 10:15:54 2025].540224 Error: ORA-00907: Rechte Klammer fehlt
 . Query statement: SELECT SDO_AGGR_MBR(geom) AS GEOM from (SELECTgeom FROM (select * from Test )) .Check your data statement.in msOracleSpatialLayerGetExtent()
[Wed Jul 23 10:15:54 2025].540256 msOracleSpatialLayerGetExtent(): OracleSpatial error. Check your data statement and server logs

The statement is generated by mapserver. A space is missing **SELECTgeom**

Please change this.

Best regards Gerhard

jratike80:
Have you set LAYER EXTENT for the Oracle layer? I believe that if it is set, then Mapserver does not generate the `SELECT SDO_AGGR_MBR(geom)` query at all. It is a very slow query with big tables anyway.

gkvoelkl:
After MAP I set EXTEND. I think it will work for all layers in the map file.

jratike80:
Please have a try and set the extent explicitly for your layer https://mapserver.org/mapfile/layer.html.

> The spatial extent of the data. In most cases you will not need to specify this, but it can be used to avoid the speed cost of having MapServer compute the extents of the data. An application can also possibly use this value to override the extents of the map.

jratike80:
Hi @gkvoelkl , did you try the workaround by setting the layer level extent?

jmckenna:
Closing to due inactivity.  Please re-open if there is a problem.

samy-rd:
Hello

I also stumbled uppon the problem while using SWIG Api's `layerObj->getExtent()` function. The error is the same as @gkvoelkl stated.

I've looked into the Mapserver function `osAggrGetExtent ` in `/src/maporaclespatial.c`.
I found that the function is not working properly if the corresponding layer/layerObj has no items (`layerObj->numItems = 0`).
If there is no Items, then the concatinated String is missing a space and the error occurs.

Comparing the function `osAggrGetExtent ` from Mapserver 7.4 to 8.4 you can also see the following:

[MS 7.4 : /maporaclespatial.c : line 1068](https://github.com/MapServer/MapServer/blob/772bd6f3d6d344ca55e893c707763c21e00561a8/maporaclespatial.c#L1068)
[...]

PR #7379, opened 2025-11-07
add extra space in construction of query

msmitherdc:
fixes https://github.com/MapServer/MapServer/issues/7312, add extra space in oracle aggr query to prevent sql error

### DIFF
--- a/src/maporaclespatial.c
+++ b/src/maporaclespatial.c
@@ -1217,7 +1217,7 @@ static char *osAggrGetExtent(layerObj *layer, char *query_str,
   size_t tmpFromSize =
       sizeof(char) * (7 + strlen(geom_column_name) + strlen(table_name) + 1);
   char *tmpFrom = msSmallMalloc(tmpFromSize);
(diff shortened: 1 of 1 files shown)


geographika:
The failing NetCDF tests are unrelated to this (and likely to be caused by a GDAL version bump in the CI - these will probably need to be made more robust). 
Also to note we have no msautotests tests for the Oracle driver.

pelord:
I don't think so. Although, I did not try the fix
 It's seem to be related to another issue. I will try to find a fix next week.

src/maporaclespatial.c
```diff
@@ -1217,7 +1217,7 @@ static char *osAggrGetExtent(layerObj *layer, char *query_str,
   size_t tmpFromSize =
       sizeof(char) * (7 + strlen(geom_column_name) + strlen(table_name) + 1);
```
geographika: I think the 7 should be updated to 8 to account for the extra space here

PR #7385, opened 2025-11-11
Fix: Memory leaks in config and landing page fixes

geographika:
Fix for following asan errors (see #7384):

```
==36744==ERROR: AddressSanitizer: stack-use-after-scope on address 0x70caa8900020 at pc 0x70cab821f96f bp 0x7ffcf1eb86e0 sp 0x7ffcf1eb7e88
READ of size 12 at 0x70caa8900020 thread T0
    #0 0x70cab821f96e in strlen ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:391
    #1 0x70cab2e84613 in pcre2_regexec (/lib/x86_64-linux-gnu/libpcre2-posix.so.3+0x1613) (BuildId: 82e51d62e96715f491183a56806553d062ba132c)
    #2 0x70cab7298236 in ms_regexec /src/src/mapregex.c:93
    #3 0x70cab72a3e28 in msEvalRegex /src/src/mapfile.c:149
    #4 0x70cab72e3c3d in msLoadMap /src/src/mapfile.c:7034
    #5 0x70cab6fae932 in msCGILoadMap /src/src/mapservutil.c:271
    #6 0x654190509ce3 in main /src/src/apps/mapserv.c:278
    #7 0x70cab63031c9  (/lib/x86_64-linux-gnu/libc.so.6+0x2a1c9) (BuildId: 274eec488d230825a136fa9c4d85370fed7a0a5e)
    #8 0x70cab630328a in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2a28a) (BuildId: 274eec488d230825a136fa9c4d85370fed7a0a5e)
    #9 0x654190508974 in _start (/src/build/mapserv+0x3974) (BuildId: 6f13b92201bf317acafea77e6371beb751e6d60f)
```

cc @jmckenna

PR #7386, opened 2025-11-13
CI: Add config to msautotest folder checks

geographika:
This looks like the cause of the CI passing while tests were failing in the `config` folder. I have added this in (and sorted alphabetically), so this should now make the ASAN CI build fail until #7385 is merged.


jmckenna:
thanks @geographika !

Issue #7387, opened 2025-11-14
generated index page contains incorrect links


jmckenna:
- regarding [RFC 140](https://mapserver.org/development/rfc/ms-rfc-140.html)
  - if a MAP is configured named `ttt` in the .conf file, then going to the index page http://127.0.0.1/cgi-bin/mapserv.exe
    - in the "HTML Documentation" section the link goes to a broken result: http://127.0.0.1/cgi-bin/mapserv.exettt/
      - notice the missing `/` after mapserv.exe
      - correct link: http://127.0.0.1/cgi-bin/mapserv.exe/ttt/
    - same issue for the generated JSON link: http://127.0.0.1/cgi-bin/mapserv.exettt/?f=json
  - also, on the resulting landing page, the "OpenLayers Viewer" link points to a broken link: http://127.0.0.1/cgi-bin/mapserv.exe?map=/my/path/ttt.map&?template=openlayers&mode=browse&layers=all
    - notice the extra `?`
    - correct link should be: http://127.0.0.1/cgi-bin/mapserv.exe?map=/my/path/ttt.map&template=openlayers&mode=browse&layers=all

(no issues with the OGCAPI service http://127.0.0.1/cgi-bin/mapserv.exe/ttt/ogcapi )

## Possibly part of the fix/workaround

- for the missing trailing `/`, around `line 327` of [mapserv-index.cpp](https://github.com/MapServer/MapServer/blob/main/src/mapserv-index.cpp#L327)  could be added in the `json createMapSummary()` function:
```
  // make sure that URLs end with '/'
  if (!onlineResource.empty() && onlineResource != "./" && onlineResource.back() != '/') {
    onlineResource += '/';
  }
```

that fixes my trailing slash issue locally.

- for the other OpenLayers Viewer link issue, likely an enhancement is needed in the `getOnlineResource()` function of mapserv-index.cpp (I stopped here and decided to file the issue)

(not sure why this wasn't caught by any CI tests)

PR #7388, opened 2025-11-15
Add additional URL creation logic for Landing Page URLs

geographika:
Fix for #7387. 

The landing page assumed a "clean" root URL, such as http://mapserver.example.com/ and for Mapfile access to be via the keys defined in the CONFIG file e.g. http://mapserver.example.com/mymap/

This PR will append a final backslash to root URLs such as http://127.0.0.1/cgi-bin/mapserv.exe, so the root becomes http://127.0.0.1/cgi-bin/mapserv.exe/.

The landing page wraps the `msOWSGetOnlineResource` used to create GetCapabilities, which uses `"*_onlineresource"` in a Map's metadata, and tries to construct the URL if this is not set. It is recommended to set `"*_onlineresource"` if possible. This PR adds another check and test to see if a querystring is already present in an online resource, so http://127.0.0.1/cgi-bin/mapserv.exe?map=/my/path/ttt.map will add an ampersand rather than a ? to create URLs such as http://127.0.0.1/cgi-bin/mapserv.exe?map=/my/path/ttt.map&template=openlayers&mode=browse&layers=all


jmckenna:
thanks @geographika !

PR #7390, opened 2025-11-19
OGCAPI: return OGC-NumberReturned, OGC-NumberMatched and Link HTTP headers

rouault:
as allowed in the spec (not mandatory for GeoJSON output, but for other output formats that cannot provide that information in the response body)


jmckenna:
thanks @rouault I see them in the response headers locally.

PR #7392, opened 2025-11-21
Output GDAL files: fix from init=epsg:xxxx layers

rouault:
When e.g. exporting a layer trough WCS whole LAYER.PROJETION is init=epsg:xxxx with the xxxx code being a custom one in a legacy 'epsg' text file, loading that projection in msProjectionObj2OGCWKT() and resulted in an output file without SRS definition


jmckenna:
thanks @rouault !

PR #7407, opened 2025-12-15
UVRaster: reproject wind vector from layer projection to map/request projection

rouault:
Before this PR, the angle of the wind vector wasn't modified if
reprojection was involved. With this PR, we now build a small vector
representing the wind direction in the layer projection and reproject it
to the target projection. We use that to correct the vector direction,
while preserving the original vector length.
The newly added test case shows a raster in EPSG:4326 that has a pure
west-oriented wind, and queried in a north polar stereographic
projection, center at the north pole. The clock wise oriented
arrows along circles are the expected result.


jmckenna:
thanks @rouault !

alistaireverett:
Thanks for this!

RousseauLambertLP:
Thanks a lot @rouault ! 

So, if I understand correctly, this fix makes it possible to use U and V vectors from a grid that is not aligned North-South geographically (seems to via a quick test on my side) ? 

See for example the following question on the MapServer-users mailing list: https://www.mail-archive.com/mapserver-users@lists.osgeo.org/msg29171.html ( cc @alistaireverett :) )

rouault:
> So, if I understand correctly, this fix makes it possible to use U and V vectors from a grid that is not aligned North-South geographically (seems to via a quick test on my side) ?

yes

PR #7410, opened 2025-12-29
WMS: rework building of the layer tree to make it faster, and understandable!

rouault:
Avoid quadratic performance in the number of layers.
    
On a 103 MB .map file with 6130 layers and 3 level of nesting, WMS  GetCapabilities response generation goes from 56 seconds to 5 seconds.


jmckenna:
thanks @rouault nice improvement!

PR #7415, opened 2026-01-17
Build: fix build against GDAL 3.13.0dev

rouault:
Those changes are compatible with earlier GDAL versions as well


jmckenna:
thanks @rouault !

PR #7418, opened 2026-01-26
Fix memory allocation for bool

PatrikSylve:
Had an issue where MapServer crashed on Windows (but not linux) when rendering raster with NODATA and `PROCESSING "SCALE=AUTO"` set in the mapfile. 

It crashed when I tested using `map2img` from the conda build ` 8.6.0 py314h835350a_2` and a local windows build. 

It seems like the  `bool` type is 4 bytes in some Windows builds, so I updated the calloc call to ensure proper memory size is allocated.


jmckenna:
thanks for this @PatrikSylve

PR #7423, opened 2026-02-03
Contour: fix 2 reprojection related issues

rouault:
- contour: fix requesting a raster georeferenced in long/lat in EPSG:3857 with a bounding box overlapping the antimeridian
- contour: fix horizontal/discontinuity lines when rendering a rotated pole raster in EPSG:4326/3857


jmckenna:
thanks @rouault !

rouault:
@jmckenna Thanks for the merge. I think it wouldn't hurt to keep regular pull requests like that one (i.e. not ones fixing security issues) opened for at least a couple days to let a chance for potential reviewers to have a look.

geographika:
@rouault - I've not had a chance to try it yet, but would it be possible to recreate the functionality of the `CONNECTIONTYPE CONTOUR` layer by using GDALG pipelines?

rouault:
> I've not had a chance to try it yet, but would it be possible to recreate the functionality of the `CONNECTIONTYPE CONTOUR` layer by using GDALG pipelines?

mostly, except that the GDAL contouring algorithm doesn't take into account the spatial filter set on the output, so that would involve contouring of the whole raster

PR #7432, opened 2026-02-12
OGCAPI Features: restore support for 'oga_compliance_mode' metadata item (8.8 only)

rouault:
and make it default to true.

That parameter was removed per PR #7360 which added support for a extra_params configuration item.

This PR adds the param names from the extra_params as allowed parameters to validate the query parameters. This is a pre-requisite for the to-come support of "Queryables as Query Parameters" of OGC API Features - Part 3 : https://docs.ogc.org/is/19-079r2/19-079r2.html#queryables-query-parameters

### DIFF
--- a/msautotest/api/expected/ogcapi_api_unknown_parameter.json.txt
+++ b/msautotest/api/expected/ogcapi_api_unknown_parameter.json.txt
@@ -1,0 +1,4 @@
+Content-Type: application/json
+Status: 400
+
+{"code":"InvalidParameterValue","description":"Unknown query parameter: unknown"}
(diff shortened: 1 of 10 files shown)


geographika:
@rouault - just to clarify, if a user has set:

```
WEB
  VALIDATION
...
  END
  METADATA
    "ows_extra_params" "token=%TOKEN%&user_id=%USER_ID%"
```

Then this code adds `token` and `user_id` to the `allowedParameters` list (along with `f`,`bbox`,`bbox-crs`,`datetime`,`limit`,`offset`, and `crs`) and any other querystring parameters will throw an error unless `oga_compliance_mode` is explicitly set to `false`. Then any queryable items added for OGC API Features will also be added to this list when querying data.

rouault:
> Then this code adds `token` and `user_id` to the `allowedParameters` list (along with `f`,`bbox`,`bbox-crs`,`datetime`,`limit`,`offset`, and `crs`) and any other querystring parameters will throw an error unless `oga_compliance_mode` is explicitly set to `false`. Then any queryable items added for OGC API Features will also be added to this list when querying data.

exactly

PR #7436, opened 2026-02-24
OGC API Features: implement CQL2 text filtering from Part 3

rouault:
New query parameters: filter, filter-lang and filter-crs

New layer metadata:
- "oga_geometry_name": geometry field name. Defaults to 'geom'
- "oga_geometry_type": among "geometry-point", "geometry-multipoint", "geometry-linestring", "geometry-multilinestring", "geometry-polygon", "geometry-multipolygon", and "geometry-geometrycollection". Or "geometry-any" which is the default


geographika:
Wow! If I'm understanding correctly, this PR also includes a Common Query Language (CQL) parser and conversion to MapServer expressions?

rouault:
> Wow! If I'm understanding correctly, this PR also includes a Common Query Language (CQL) parser and conversion to MapServer expressions?

yes, that's the bulk of the work.  Also working on CQL2 JSON for a follow-up PR, that will use the encoding agnostic code in cql2.h/.cpp

PR #7437, opened 2026-02-24
No longer depend on GDAL's cpl_port.h MIN/MAX/ABS macros

rouault:
They have been renamed in GDAL 3.13dev to avoid poluting the namespace. Use MS_ prefixed macros instead of std::min/max/abs from C++


jmckenna:
@rouault GDAL's `cpl_port.h` is also the source for the `M_PI` macro (see kerneldensity.c, mapmssql2008.c, mappostgis.cpp), although very odd that only I am hitting this, on MSVC.  I can workaround, but, that is what I found.  It might be just my packaging, so you can ignore my comment.

PR #7439, opened 2026-02-24
Minor optimisations for the MSSQL driver

geographika:
This PR has 2 changes:

1. Avoid multiple calls to get the column details from the database. The driver currently calls `select top 0 *...` 3 times for a WFS request, in the `msMSSQL2008LayerGetItems` function (mapped to `vtable->LayerGetItems`). I was hoping to cache `ms_MSSQL2008_layer_info_t` or avoid it being removed, but `msLayerGetItems` calls `msLayerFreeItemInfo` resetting this each time. The only stable object is `msODBCconn_t`, so the column details are cached here. If the next call's SQL matches exactly then the column details from the cache is retrieved. This reduces database calls to a single `select top 0 *...` for a WFS request, and helps when reviewing SQL Server profile logs (there is no noticeable performance difference). 

2. Avoid calling `convert(nvarchar(max)` for all fields when selecting from the database. All data is returned as strings to MapServer, but by using `convert(nvarchar(256)` for smaller SQL Server types (such as int, float etc.) it reduces memory usage on the database server. There is no noticeable different to MapServer performance itself.

The main benefit is that the database server doesn't need to allocate as much memory to the query, which can be important with many concurrent database requests. In a sample query, the profiler notes a drop from RequestedMemory **~143 MB** to RequestedMemory  **~8.8 MB** - a 16x reduction. The query only actually uses 1.9 MB. The memory is then available for other queries. 

```xml
<!-- before -->
<MemoryGrantInfo SerialRequiredMemory="5784" SerialDesiredMemory="146176" RequiredMemory="5784" DesiredMemory="146176" RequestedMemory="146176" GrantWaitTime="0" GrantedMemory="146176" MaxUsedMemory="1912" MaxQueryMemory="3018200" LastRequestedMemory="0" IsMemoryGrantFeedbackAdjusted="No: First Execution" />
<!-- after -->
<MemoryGrantInfo SerialRequiredMemory="5784" SerialDesiredMemory="9048" RequiredMemory="5784" DesiredMemory="9048" RequestedMemory="9048" GrantWaitTime="3" GrantedMemory="9048" MaxUsedMemory="1912" MaxQueryMemory="3004400" LastRequestedMemory="146176" IsMemoryGrantFeedbackAdjusted="Yes: Adjusting" />
```

It is possible to also get the column size, and cache this (as part 1 of this PR), which would also allow smaller sizes to be used for fields such as `ncharvar(240)`, but this requires more substantial changes to the driver. 

Initially, I was trying to avoid the MSSQL driver from running the same query twice for WFS queries (see https://mapserver.org/development/rfc/ms-rfc-52.html). I managed to get the server-side database cursor working, but it actually degraded performance with a database call for every feature in the recordset. Any other attempts at caching were basically the same approach as using `"wfs_features_cache_count" "1000"`, and the attempt was abandoned.

### DIFF
--- a/src/mapmssql2008.c
+++ b/src/mapmssql2008.c
@@ -2979,11 +2979,6 @@ int msMSSQL2008LayerGetItems(layerObj *layer) {
     layerinfo->conn->itemtypes[i] = layerinfo->itemtypes[i];
   }
 
-  if (layer->debug && layerinfo->conn->cache_sql) {
-    msDebug("msMSSQL2008LayerGetItems: cached %d items on connection "
-            "for sql: %s\n",
-            layerinfo->conn->numitems, layerinfo->conn->cache_sql);
-  }
   return msMSSQL2008LayerInitItemInfo(layer);
 }
 


geographika:
@szekerest - any objections to merging the above?

Note the PostGIS driver suffers similar problems, in that a single WFS request queries the datasource schema (and PostGIS version) multiple times, and could reduce logs and database access with a similar caching approach.

szekerest:
@geographika Checked the code and it looks correct to me. You can go ahead and merge it.

PR #7446, opened 2026-02-26
Define _USE_MATH_DEFINES so that M_PI is available when include math.h

rouault:
@jmckenna Can you give this a try?


jmckenna:
that's exactly what I added to the very top of kerneldensity.c and it fixed the error for that file...but I will try this PR now

jmckenna:
thanks @rouault that worked!!!

PR #7449, opened 2026-03-01
Add new `fallback_to_map_extent` to avoid layer extent calculations

geographika:
This PR adds a new `fallback_to_map_extent` metadata item to avoid calculating layer extents from the datasource when they are not explicitly defined. This can be added to the `MAP WEB METADATA`, for example:

```
WEB
  METADATA     
    "oga_fallback_to_map_extent" "true"
  END
END
```

`msOWSGetLayerExtent()` has been updated to use the following logic when determining a layer extent:

1. layer "*_extent" metadata
2. layer->extent if valid
3. **map->extent if valid and the fallback_to_map_extent metadata item is set** - *this is the logic introduced by this PR*
4. msLayerGetExtent() to read extent from the layer

The rationale for this change is related to OGC API Features requests to `/ogcapi/collections`. This endpoint requires that all layers in a Mapfile have either an `EXTENT` defined or an extent metadata item set. If neither is present, MapServer calculates the extent by querying the datasource.

In a Mapfile with many database-backed layers (for example, 50 layers), a single request to `/ogcapi/collections` may trigger 50 extent calculations against the database (or manually/programatically add these extents to the Mapfile). This can significantly impact performance.

In many deployments, precise per-layer data extents are not critical, and using the MAP extent as a fallback is sufficient.

Additionally, if a layer is empty, the extent calculation can fail, resulting in the service returning:  `Unable to get collection bounding box`. 

https://github.com/MapServer/MapServer/blob/a6cf3b47a9743c7d3b3a66d8d3ad0accb5d6c078/src/mapogcapi.cpp#L949-L951


jratike80:
Maybe OK, but for some source data types like GeoPackage msLayerGetExtent() is fast and gives automatically extent that is always up to date. Having `fallback_to_map_extent` on MAP level would leave only fixed hand written LAYER-EXTENT or fixed hand written MAP-EXTENT. In a way the setting would rather belong to LAYER level, but it would require more writing.

On the other hand, it is not always good that the layer extent is changing every time when features are added to or removed from the datasource.
Could in make sense to have an option `LAYER-EXTENT "AUTO"` that would trigger the current default, that is, to query the extent from the datasource, and that would override the new `MAP-fallback_to_map_extent`? Or would it get too complicated"

geographika:
Thanks, @jratike80, for the feedback - good points in the cases where you may want the latest extent for certain layers where calculating extent is almost instant (raster layers would probably fall into this case too).

I think adding `fallback_to_map_extent` as a `LAYER` metadata option to override a `MAP` level default would solve this?

Currently, the extent is always auto-calculated if no manual extent is set, so I'd be reluctant to add something that would change existing Mapfile behaviour. 

I'm also open to suggestions for alternatives to the metadata name `fallback_to_map_extent`.

jratike80:
> I think adding fallback_to_map_extent as a LAYER metadata option to override a MAP level default would solve this?

Yes. But if extent is usually set in LAYER-EXTENT, could it be better to allow some special keyword for EXTENT? Otherwise it will be possible to set one extent with LAYER-EXTENT, and another with LAYER-METADATA-map-extent. That can make confusion.

geographika:
> Yes. But if extent is usually set in LAYER-EXTENT, could it be better to allow some special keyword for EXTENT? Otherwise it will be possible to set one extent with LAYER-EXTENT, and another with LAYER-METADATA-map-extent. That can make confusion.

This is already the case with metadata items and EXTENT:

```
LAYER
  EXTENT 190012 4816648 762254 5472427 # ignored (for OGC features request)
  METADATA
    "ows_extent" "180000 470000 770000 5300000" # ignored (for OGC features request)
    "oga_extent" "190000 480000 760000 5400000" # for an OGC Features request this takes precedence
    "oga_fallback_to_map_extent" "true" # ignored
  ...  
  END
```

Any new EXTENT keyword would also be ignored in the above example as the "oga_extent" takes precedence. 

To fallback to the MAP extent would require several items to be missing:

```
LAYER
  # no EXTENT
  METADATA
    # no "*_extent" metadata
    "oga_fallback_to_map_extent" "true" # ignored
  ...  
  END
```

jratike80:
Ok, so not much additional mess.

Issue #7362, opened 2025-10-15
ORA-00907 : missing right parenthesis


pelord:
*** Seem to be related to https://github.com/MapServer/MapServer/issues/7312 ***
For @jmckenna :  I could not reopen the issue. 

With Mapserver 8.4.1, 

When applying WMS Vendor specific filter  on an Oracle Layer, in insensitive mode, a extra parenthesis seem to be added.

Here the applied filter
```
FILTER=(<Filter xmlns="http://www.opengis.net/ogc"><PropertyIsEqualTo matchCase="false"><PropertyName>DIRECTION</PropertyName><Literal>N</Literal></PropertyIsEqualTo></Filter>)
```

Here a part of debug log (DEBUG 5). I have remove some parts of the SQL, by I've kept the WHERE clause. 

```

SELECT DIRECTION,rownum, GEOM,myid 


 FROM MY_TABLE WHERE  ( REGEXP_LIKE( DIRECTION, '^N$'),'i'  ) )  


AND SDO_FILTER( .....)
```
This query contain an extra ) before the 'i' in the clause. 

It's look like it is relate to the match case property because it is working properly with matchCase set to true.

Here my layer definition. I've tried with a direct acces to the table and with a subselect. Same result
```
ENCODING "WINDOWS-1252"
CONNECTIONTYPE ORACLESPATIAL
CONNECTION "USER/{ENCRYPTED_PWD}@TNSNAME" 
PROCESSING "CLOSE_CONNECTION=DEFER"
DATA "GEOM from  MY_TABLE using unique myid SRID 3798"
```

jmckenna:
sorry for the hassle @pelord I now reopened that other issue

jmckenna:
@pelord does PR https://github.com/MapServer/MapServer/pull/7379 also solve your issue here?

jmckenna:
@pelord any luck in finding the fix for this?

pelord:
This afternoon :)

jmckenna:
that would be good timing, thanks again.

pelord:
I compile it tomorrow morning and I give you some feedback in the morning. Sorry for the delay!

pelord:
seem to be related to this change

https://github.com/MapServer/MapServer/commit/a468ec94918db4260a9ed3061988419f76642ce8

pelord:
@mikitex70  Any test / use case for your change?

pelord:
[...]

PR #7452, opened 2026-03-02
revert(pr7012): issue with case insensitive ogc filters

pelord:
Reverting https://github.com/MapServer/MapServer/pull/7012

according to discussions with the PR author.

https://github.com/MapServer/MapServer/issues/7362#issuecomment-3662025709

### DIFF
--- a/src/maporaclespatial.c
+++ b/src/maporaclespatial.c
@@ -4334,12 +4334,9 @@ int msOracleSpatialLayerTranslateFilter(layerObj *layer, expressionObj *filter,
         native_string = msStringConcatenate(native_string, snippet);
         if (ieq == MS_TRUE) {
           native_string = msStringConcatenate(native_string, "$");
+		  ieq = MS_FALSE;
         }
         native_string = msStringConcatenate(native_string, "'");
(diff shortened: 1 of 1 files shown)


pelord:
at mapserver team! 

You could also revert the PR/commit without merging mine :)

jmckenna:
cc @mikitex70

pelord:
@mikitex70 Any comments to rollback your change?

mikitex70:
You are free to rollback my change, the patch resolves only some of the issues. The big problem is the expression parser which is not so robust, in particular when regex expressions are used. A way to workaround its issues is to wrap expression with extra parenthesis, but sometimes it doesn't work.

pelord:
Go then? :)

jmckenna:
> Go then? :)

Your pull request here was still failing (line# 4337 had tabs instead of spaces at the start of the line), I've pushed a change here now...

pelord:
bump!!  :)

In the next release? I've fixed my local version and I do not think to integrate the next mapserver version soon, but before to forget it!

jmckenna:
@pelord thanks for the much-needed bump.  Merging now..

Issue #7231, opened 2025-02-20
Advertised ESRI:XXXX projection codes fail in WMS


geographika:
When requesting a WMS using an `ESRI:XXXX` projection e.g. `&SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&BBOX=-19948500.0,-6147449.0951699,19688885.757062882,8361000.0&CRS=ESRI:54052`

The following error is returned:

```
 msProcessProjection(): Projection library error. PROJ error "Invalid value for an argument" when instantiating "init=ESRI:54052"
```

Mapfile:

```
    PROJECTION
        'ESRI:54052'
    END #PROJECTION
    WEB
        METADATA
            'ows_srs'             'ESRI:54052 EPSG:4326 EPSG:3857 ESRI:54009'
```

The `init=` is added on the following line:

https://github.com/MapServer/MapServer/blob/12816a4b8c56db6f840bf73270f35ea4c8f57a57/src/mapfile.c#L1360

Note `init=` is added in several places in the codebase: https://github.com/search?q=repo%3AMapServer%2FMapServer+init%3Depsg%3A%25s&type=code

MapServer handles `init=EPSG:XXXX` correctly, and uses the proj.db to lookup the projection, however `init=ESRI:XXXX` causes PROJ to look for a custom projection file in `PROJ_DATA`, even though proj.db contains the ESRI projections. 

A workaround is to create a custom projection file in the  `PROJ_DATA` folder named `ESRI` (note on Linux this is case-sensitive so all projections must be referenced as `ESRI:54009` not `esri:54009`). Then copy in the projection definitions in the following format:

```
<54009> +proj=moll +lon_0=0 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs +type=crs
<54052> +proj=igh +lon_0=0 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs +type=crs
```

Ideally proj.db would be used. See also #7230.

geographika:
For reference: https://lists.osgeo.org/pipermail/mapserver-dev/2023-April/016957.html

> init=epsg:XXXX, as a way of doing a lookup in proj.db, is a legacy 
syntax that was kept only in PROJ >=6 for EPSG. For other authorities, 
PROJ only recognizes "AUTHORITY_NAME:CODE" for lookups in the database, 
which enforces the order axis of the CRS definition (so EPSG:4326 != 
[...]

Issue #6981, opened 2023-12-07
Custom projections not supported via WFS


geographika:
Pull request https://github.com/MapServer/MapServer/pull/6795 allowed custom projections to be requested through WMS requests, however WFS and WCS don't have the same support. 

Use of custom projections currently return the following error:

```xml
<ows:Exception exceptionCode="InvalidParameterValue" locator="srsname">
    <ows:ExceptionText>msWFSGetFeature(): WFS server error. Invalid GetFeature Request:Invalid SRS.  Please check the capabilities and reformulate your request.</ows:ExceptionText>
</ows:Exception>
```

Adding support would allow requests such as the following to be supported:

+ `map=wfs_custom_projection.map&SERVICE=WFS&VERSION=2.0.0&REQUEST=GetFeature&TYPENAME=province&srsname=epsg2:42304`

+ `map=wfs_custom_projection.map&SERVICE=WFS&VERSION=2.0.0&REQUEST=GetFeature&TYPENAME=province&srsname=ESRI:108022`

jratike80:
I would like to see how that custom CRS would appear in the GML3 from WFS 2.0 and in the DescribeCoverage from WCS 2.0.

geographika:
@jratike80 if there was a custom definition such as `epsg2::42304` it would be included in the Mapfile as follows:

```
WEB
  METADATA
    "wfs_title"		   "Test simple wfs"
    "wfs_srs"		   "epsg2:42304 EPSG:4326 EPSG:3857 esri:102001"
```

And would appear in the responses in the following form (for WFS 2.0 or if `wfs_return_srs_as_urn` is set to True for earlier WFS versions):

```
      <wfs:boundedBy>
      	<gml:Envelope srsName="urn:ogc:def:crs:epsg2::42304">
      		<gml:lowerCorner>2253355.750000 -76361.773438</gml:lowerCorner>
      		<gml:upperCorner>2747648.500000 515489.031250</gml:upperCorner>
      	</gml:Envelope>
```

WCS projection code paths are again different from both WMS and WFS so would be addressed separately.

geographika:
Closed with https://github.com/MapServer/MapServer/pull/7457

PR #7457, opened 2026-03-11
Add support for non-EPSG projections for WCS and WFS protocols

geographika:
This pull request adds support for using non-EPSG projection codes in WCS and WFS services. This includes both accessing the data in a non-EPSG projection and advertising the projections in GetCapabilities requests. For example the `wcs_custom_projection_capabilities201.xml` response includes:

```xml
  <crs:CrsMetadata>
    <crs:crsSupported>http://www.opengis.net/def/crs/ESRI/0/54052</crs:crsSupported>
    <crs:crsSupported>http://www.opengis.net/def/crs/EPSG/0/4326</crs:crsSupported>
    <crs:crsSupported>http://www.opengis.net/def/crs/ESRI/0/53009</crs:crsSupported>
    <crs:crsSupported>http://www.opengis.net/def/crs/epsg2/0/42304</crs:crsSupported>
  </crs:CrsMetadata>
```

And any of these projections can be used to request a coverage e.g. `&SERVICE=WCS&VERSION=2.0.1&REQUEST=GetCoverage&COVERAGEID=test&SUBSETTINGCRS=urn:ogc:def:crs:ESRI::53009&SUBSET=x(-3276599.0027238373,-3258457.2498260457)&SUBSET=y(8498088.094072288,8507095.190921152)&SCALESIZE=x(512),y(254)&OUTPUTCRS=urn:ogc:def:crs:ESRI::53009&FORMAT=image/png`

A full test suite for WCS 1.0.0, 1.1.1, and 2.0.1 is added in `wcs_custom_projection.map`. 

As several functions were shared with the WFS code, WFS was also updated, and a full test suite found in `wfs_custom_projection.map`. 

WMS had already been updated to handle custom CRSs in https://github.com/MapServer/MapServer/pull/6795. 

**Note** - OGC Features API was not updated, and does not support handling non-EPSG projections. This can be dealt with in a subsequent PR. 

Other notes:

- The majority of code changes were to `msLoadProjectionString` - which is used by several MapServer service protocols. 
- The various projection formats which can be handled by the updated are tested in a new Python MapScript test file `msautotest/mspython/test_projections.py`. In summary any code in the format AUTH:CODE that is in `proj.db` will be handled, and any authorities not in proj.db can be added using `init=CUSTOMAUTH:CODE` and search `PROJ_DATA` for a file with a Proj string. 
- This PR fixed a few existing tests which used the newer `EPSG:3857` rather than `init=EPSG:3857`, removing CRS warnings in WMS GetCapabilities

This PR fixes the following issues:

- https://github.com/MapServer/MapServer/issues/7231 - Advertised ESRI:XXXX projection codes fail in WMS
- https://github.com/MapServer/MapServer/issues/6981 - Custom projections not supported via WFS

### DIFF
--- a/src/mapwfs.cpp
+++ b/src/mapwfs.cpp
@@ -304,11 +304,10 @@ static void msWFSPrintRequestCap(const char *request, const char *script_url,
 ** (http://www.opengis.net/def/crs/EPSG/0/4326) format.
 */
 static int msWFSLocateSRSInList(const char *pszList, const char *srs) {
-  int nTokens = 0, i = 0;
+  int nTokens = 0;
   char **tokens = NULL;
   int bFound = MS_FALSE;
-  const char *code = NULL;
-  char authority[64];
+  std::string authority, code;
 
   if (!pszList || !srs)
     return MS_FALSE;
@@ -320,11 +319,7 @@ static int msWFSLocateSRSInList(const char *pszList, const char *srs) {
     const char *sep = strchr(p, ':');
     if (sep == NULL)
       return MS_FALSE;
-    /* extract authority */
-    size_t authlen = sep - p;
-    if (authlen >= sizeof(authority))
-      return MS_FALSE;
-    strlcpy(authority, p, authlen + 1);
+    authority = std::string(p, sep - p);
     /* skip version field (may be empty) */
     p = sep + 1;
     sep = strchr(p, ':');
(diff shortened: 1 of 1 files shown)


rouault (commented):
Nice work!

geographika:
Thanks for the review @rouault - much appreciated!
Having gone through lots of the projection code I think there are quite a few changes it would be good to add as part of a MapServer 9.0, including a requirement for PROJ9+. In addition:

- try and get rid of the custom MapServer `+epsgaxis=ne parmaeter` and use PROJ to get axis order
- review and update all the references to `init=` checks in the codebase

Issue #7463, opened 2026-03-26
WFS SortBy only applied within Oracle pages


JamieIdox:
When using Oracle, SortBy is only applied to each page, it is not applied to all the data before paging. Is this deliberate? ie, if I submit a pull request to fix this is it likely to be accepted or rejected?

eg, in the following request note the startIndex and SortBy:

```
<GetFeature xmlns="http://www.opengis.net/wfs/2.0" service="WFS" version="2.0.0" startIndex="1000" count="1000" xsi:schemaLocation="http://www.opengis.net/wfs/2.0 http://schemas.opengis.net/wfs/2.0/wfs.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
	<Query typeNames="wfs:Planning_Application_Polygons" srsName="EPSG:27700" xmlns:wfs="http://mapserver.gis.umn.edu/mapserver">
		<Filter xmlns="http://www.opengis.net/fes/2.0">
			<BBOX>
				<ValueReference>SHAPE</ValueReference>
				<Envelope xmlns="http://www.opengis.net/gml/3.2" srsName="EPSG:27700">
					<lowerCorner>139987.2407923497 -3524.05645833332</lowerCorner>
					<upperCorner>316170.54920765024 157942.01645833335</upperCorner>
				</Envelope>
			</BBOX>
		</Filter>
		<SortBy xmlns="">
			<SortProperty>
				<PropertyName>DATEMODIFIED</PropertyName>
				<SortOrder>DESC</SortOrder>
			</SortProperty>
		</SortBy>
	</Query>
</GetFeature>
```

This results in the following SQL, note that ROWNUM is included in the inner query:

```
SELECT * from (
	SELECT atmp.*, ROWNUM rnum from (
		SELECT "REFVAL", "KEYVAL", "DATEMODIFIED", "ADDRESS", "DESCRIPTION", "OBJECTID", rownum, SHAPE 
		FROM PA_V1_6_DCAPPLICATIONS_POLY 
		WHERE  ISPAVISIBLE = 1  AND SDO_FILTER( SHAPE, MDSYS.SDO_GEOMETRY(2003, :srid, NULL,MDSYS.SDO_ELEM_INFO_ARRAY(1,1003,3),:ordinates ),'querytype=window') = 'TRUE' 
		ORDER BY "DATEMODIFIED" DESC   
	) atmp 
	where ROWNUM<=2000
) 
where rnum >=1001
```

[...]

PR #7464, opened 2026-03-27
Oracle: add subquery to apply sortBy before rownum paging - #7463

JamieIdox:
the main query is contained in the inner btmp sub query, rownum is moved to outside that query so that it is applied after the inner ORDER BY. I have tried to keep this change as simple as possible just by rearranging the existing code. 

I am slightly concerned that we are missing a trailing AND from the removed code. 

I have not been able to test or even build this as I don't have a build environment.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,10 @@ permissions:
 jobs:
   build:
     runs-on: ubuntu-latest
+    
+    defaults:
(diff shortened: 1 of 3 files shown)


JamieIdox:
Thank you! 🙏

msmitherdc (approved):
Sorry I missed this waiting in me. Approved.

PR #7473, opened 2026-04-09
Add guard for invalid projection strings

geographika:
Follow-up to #7457 to add a test and guard for invalid projection strings in a Mapfile.

Fixes: Reference Info: 500975699 mapserver:mapfuzzer: Null-dereference READ in msProcessProjection https://oss-fuzz.com/testcase-detail/6167937613627392

### DIFF
--- a/src/mapproject.c
+++ b/src/mapproject.c
@@ -730,7 +730,7 @@ static int _msProcessAutoProjection(projectionObj *p) {
 int msProcessProjection(projectionObj *p) {
   assert(p->proj == NULL);
 
-  if (p->numargs < 1 || p->args[0] == NULL) {
+  if (p->numargs < 1) {
     msSetError(
         MS_PROJERR,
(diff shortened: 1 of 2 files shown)


src/mapproject.c
```diff
@@ -730,6 +730,14 @@ static int _msProcessAutoProjection(projectionObj *p) {
 int msProcessProjection(projectionObj *p) {
   assert(p->proj == NULL);
 
+  if (p->numargs < 1 || p->args[0] == NULL) {
```
rouault: can we actually have p->numargs == 1 and p->args[0] == NULL ?

geographika: I guess that would indicate another problem elsewhere! I removed the check.

geographika:
Just to add `msProcessProjection` was unmodified by #7457, but maybe the changes allowed the fuzzer to send invalid input to the function (or gave more possible examples of projection strings).

PR #7478, opened 2026-04-15
Add GetFeatureInfo support for CONNECTIONTYPE KERNELDENSITY/IDW layers

### LINKED ISSUE #7477: GetFeatureInfo requests returns empty when layer has connectiontype kerneldensity
When you have a layer using connectiontype kerneldensity like the heat.map in msautotest https://github.com/MapServer/MapServer/blob/main/msautotest/gdal/heat.map a request for GetFeatureInfo returns empty:
(I have modified the heat.map by adding "wms_getfeatureinfo_formatlist" "text/html" in web.metadata and TEMPLATE "heatmap_query.html" for the layer)
```
MAPSERVER_CONFIG_FILE=mapserver-config.conf GDAL_DISABLE_READDIR_ON_OPEN=EMPTY_DIR MS_DEBUGLEVEL=5 CPL_TIMESTAMP=ON CPL_DEBUG=ON GDAL_HTTP_VERBOSE=YES /usr/local/bin/mapserv QUERY_STRING="map=heat.map&&LAYERS=heatmap&BBOX=-82.544323889094,-41.272161944547,82.544323889094,41.272161944547&WIDTH=200&HEIGHT=100&VERSION=1.1.1&FORMAT=image%2Fpng&SERVICE=WMS&REQUEST=GetFeatureInfo&STYLES=&EXCEPTIONS=application%2Fvnd.ogc.se_inimage&SRS=EPSG%3A4326&QUERY_LAYERS=heatmap&INFO_FORMAT=text/html&X=10&Y=10"
Warning 1: Unknown configuration option 'MS_MAP_PATTERN'.
msLoadMap(): 0.009s
CGI Request 1 on process 195225
msRasterQueryByRect(heatmap): entering.
msRasterQueryByRect(heatmap): layer data and tileindex NULL ... doing nothing.msRasterQueryByRect(heatmap): entering.
msRasterQueryByRect(heatmap): layer data and tileindex NULL ... doing nothing.msQueryByPoint(): No matching record(s) found.
Content-Type: text/html

mapserv request processing time (msLoadMap not incl.): 0.010s
msFreeMap(): freeing map at 0x651e1a781d10.
freeLayer(): freeing layer at 0x651e1a7ccb80.
freeLayer(): freeing layer at 0x651e1aaa8f80.
mapserv total execution time: 0.020s
[Wed Apr 15 10:55:30 2026].5311, 0.0000: GDAL: In GDALDestroy - unloading GDAL shared library.

```

Looking at the code this seems expected https://github.com/MapServer/MapServer/blob/main/src/maprasterquery.c#L702-L707 as the function returns MS_SUCCESS of there are not tileindex nor data

I expect the GetFeatureInfo to return the configured data in the given template.

To me it looks like the connectiontype kerneldensity ( and idw) was added for the GetMap, but not GetFeatureInfo.

TAlonglong:
closes #7477 

msRasterQueryByRect() previously returned immediately with no results when layer->data and layer->tileindex were both NULL, which is always the case for KERNELDENSITY and IDW layers. These connection types generate an in-memory GDAL dataset on the fly via msInterpolationDataset().

Changes:
- Bypass the early-return guard for KERNELDENSITY/IDW in msRasterQueryByRect()
- Call msGDALInitialize() unconditionally before the dataset branch so the GDAL MEM driver is always registered
- Compute map->cellsize via msAdjustExtent() when it is 0 (query path does not call msCalculateScale() unlike the draw path), restore it afterwards
- Create and query the interpolation dataset, then clean up as in the draw path

Test: new RUN_PARMS in msautotest/gdal/heat.map exercising WMS GetFeatureInfo against the existing kerneldensity heatmap layer.


jratike80:
I believe that GetFeatureInfo from IDW layers will return meaningful results, for example interpolated height between measured points. But what does the result from kerneldensity layers present? What is the unit?

TAlonglong:
OK, IDW was just copy/pasted from the GetMap part without checking what this does. That can be removed.

For the unit I would assume as the name says, kerneldensity, the density of the feature. But this will vary with the selected value of the radius. So this is a good question which I don't have an answer.

jmckenna:
thanks for including tests @TAlonglong

src/maprasterquery.c
```diff
@@ -746,50 +750,100 @@ int msRasterQueryByRect(mapObj *map, layerObj *layer, rectObj queryRect)
...
+      if (!layer->tileindex) {
+        char **connectionoptions =
+            msGetStringListFromHashTable(&(layer->connectionoptions));
+        hDS = GDALOpenEx(decrypted_path, GDAL_OF_RASTER, NULL,
+                         (const char *const *)connectionoptions, NULL);
+        CSLDestroy(connectionoptions);
+      } else {
+        hDS = GDALOpen(decrypted_path, GA_ReadOnly);
+      }
     }
```
geographika, line 847: Are there any changes here, or is it just reformatting (if so this should be reverted)?

TAlonglong: Its indentation as far as I see it due the added if connectontype/idw else <old stuff>

geographika (approved):

jmckenna:
thanks again @TAlonglong !

PR #7480, opened 2026-04-19
Allow only SRS or CRS parameters for OpenLayers template output

geographika:
Fix for a XSS issue, where unvalidated CRS parameters were added to the OpenLayers HTML template output when WMS 1.1.0 is used. 

The fix checks the WMS version and accepts either SRS or CRS, and ignores the other parameter. MapServer then checks that these values are one of the allowed CRSs in the Mapfile, which prevents unwanted input.

All other template parameters are all validated by the existing WMS code (OUTPUTFORMAT, LAYERS, VERSION etc.).


jmckenna:
thanks @geographika !

PR #7500, opened 2026-05-08
INSPIRE Default Schemas Location: use https instead of plain http

tschorr:
The current default for the INSPIRE schema repository location is http://inspire.ec.europa.org/schemas (plain HTTP).
However a request to this location will actually return a HTTP 301 Moved Permanently response, with the corresponding https address as the actual schema location.
This causes INSPIRE Validator tests to fail because the tests do not follow the redirect.

The behaviour is different for the default OGC schema repository location:
While this is also a plain http address (http://schemas.opengis.net), schema files are returned with HTTP status code 200.

This PR changes the default to https://inspire.ec.europa.org/schemas.


tschorr:
Coveralls failed with a timeout. I'm not allowed to restart the check.

rouault:
> This causes INSPIRE Validator tests to fail because the tests do not follow the redirect.

this PR is fine, but shouldn't that be reported to Inspire Validator too ? Following 301/302 is pretty standard internet behavior...

PR #7504, opened 2026-05-11
mapserv: exit FastCGI loop promptly on SIGTERM/SIGUSR1

Firefishy:
When mapserv is built with FastCGI and is idle inside FCGI_Accept(), SIGTERM only sets finish_process=1 and is masked by the SA_RESTART default of signal(), so the blocking accept()/select() resumes and the process appears hung until the next request arrives.

Call FCGX_ShutdownPending() from the handler (signal-safe; sets libfcgi's internal shutdown flag) and install handlers via sigaction() with sa_flags=0 so the syscall returns EINTR. libfcgi then returns -1 from FCGI_Accept() and the loop exits cleanly.

_Disclosure: this issue write-up and the accompanying patch were produced with the assistance of GitHub Copilot (Claude Opus 4.7); the code change was generated by the AI and reviewed by me before submission._

### DIFF
--- a/src/apps/mapserv.c
+++ b/src/apps/mapserv.c
@@ -60,6 +60,14 @@ static int finish_process = 0;
 static void msCleanupOnSignal(int nInData) {
   (void)nInData;
   finish_process = 1;
+#ifdef USE_FASTCGI
+  /* Ask libfcgi to stop accepting new requests. This causes a blocked
+   * FCGI_Accept() to return -1, so the main loop can exit promptly
+   * instead of waiting for the next request to arrive. Safe to call
+   * from a signal handler: it just sets an internal flag (declared in
+   * fcgiapp.h, which is pulled in via fcgi_stdio.h above). */
+  FCGX_ShutdownPending();
+#endif
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


Firefishy:
I have tested locally and this fixes the SIGTERM issue I was observing. 

Error reproduced with:
```
spawn-fcgi -p 9000 -n -- /usr/local/bin/mapserv 
# New window
pkill mapserv
```
Prior to the fix mapserv would "ignore" the TERM and only shutdown at the next request, which if mapserver was idle is never received.

Firefishy:
The code change from this PR is very similar to this StackOverflow solution: https://stackoverflow.com/a/41374601/2551154

Firefishy:
Feel free to close this PR if an AI contribution is not acceptable, I will not be offended :-)

jmckenna:
thanks @Firefishy, looks fine to me.  Let's give a few more days to let others possibly review this.

src/apps/mapserv.c
```diff
@@ -242,8 +250,21 @@ int main(int argc, char *argv[]) {
...
 #ifndef _WIN32
-  signal(SIGUSR1, msCleanupOnSignal);
-  signal(SIGTERM, msCleanupOnSignal);
+  {
+    /* Install via sigaction without SA_RESTART so that a blocking
+     * accept()/select() inside FCGI_Accept() returns EINTR when the
+     * signal arrives while the process is idle. Combined with
+     * FCGX_ShutdownPending() in the handler, this lets the FastCGI
+     * loop exit on SIGTERM instead of waiting for the next request. */
+    struct sigaction sa;
```
rouault, line 259:
```suggestion
    struct sigaction sa;
    memset(&sa, 0, sizeof(sa));
```

Firefishy:
Requested changes made, squashed and pushed.

rouault (approved):

jmckenna:
thanks again @Firefishy

geographika:
I only got a chance to look at this now. Just wondering if the kill process could be run as part of the CI in https://github.com/MapServer/MapServer/blob/main/.github/workflows/start.sh at the end, just to check it runs correctly in the future. 

```
spawn-fcgi -p 9000 -n -- /usr/local/bin/mapserv 
# New window
pkill mapserv
```

PR #7512, opened 2026-05-22
Fix incorrect JSON error message output on Index Pages

geographika:
A request for a non-existent format type for an index page, such as http://localhost:7000/?f=missing returns malformed JSON:

```json
Content-Type: application/json
Status: 400

{"code":"InvalidParameterValue","description":"Unsupported format requested: missing"}
Content-Type: application/json
Status: 400

{"code":"InvalidParameterValue","description":"Unsupported format requested."}
```

The `msOGCAPIGetOutputFormat` function was called twice, and writes an error message to the output. 

This PR removes the duplicate call, and lets the caller decide the error message to return. 

The invalid format parameter provided by the user is no longer returned, as this is unsanitized input. The new JSON response (as checked for in the test) is:

```json
Content-Type: application/json
Status: 400

{"code":"InvalidParameterValue","description":"Unsupported format requested."}
```


rouault (approved):

pvgenuchten:
Thanx for quick action @geographika

PR #7517, opened 2026-05-29
WMS OpenLayers output: properly escape mapserv_onlineresource

rouault:
Fixes https://github.com/MapServer/MapServer/security/advisories/GHSA-xqj6-vjqr-33vv


.github/workflows/start.sh
```diff
@@ -190,4 +190,11 @@ EOF
 MAPSERVER_CONFIG_FILE=/tmp/mapserver.conf mapserv QUERY_STRING="MAP=wfs_simple.map&SERVICE=WFS&REQUEST=GetCapabilities" > /tmp/res.txt
 cat /tmp/res.txt | grep "Web application error" >/dev/null || (cat /tmp/res.txt && /bin/false)
 
+# Check https://github.com/MapServer/MapServer/security/advisories/GHSA-xqj6-vjqr-33vv
```
geographika, line 193: Do these links become public when the PR is merged? Or is this for org members only in the future?

geographika (approved):

.github/workflows/start.sh
rouault: Past adviosries are publicly available https://github.com/MapServer/MapServer/security/advisories/GHSA-4h8g-378q-r75m . So the link in the PR should also be once the advisory is made public

PR #7522, opened 2026-06-10
Fix SLD generation when using string operators such as `le`

geographika:
If field names contained operators such as `le`, `ge` etc. then these were being picked up as operators.
This PR checks they are surrounded by spaces (and checks for them last after `<=`). 

```
    LAYER
      NAME "Test32"
      TYPE POINT
      CLASS
        EXPRESSION (([field_le] > 0) AND ([field_le] <= 6))
        STYLE
            COLOR 0 176 80
            WIDTH 1
        END
      END
    END
```

Without fix would generate:

```xml
...
<ogc:PropertyIsLessThanOrEqualTo><ogc:PropertyName>field_</ogc:PropertyName><ogc:Literal>] <= 6</ogc:Literal>
```

Bug introduced by me in https://github.com/MapServer/MapServer/pull/7182#issuecomment-2505849392 which fixed one issue, but created another.


rouault (approved):

jmckenna:
thanks @geographika !

PR #7526, opened 2026-06-13
Switch XML mapfile parsing to safe default libxml2 options

geographika:
Disable XML entity expansion and external DTD loading in XML mapfile parsing (see https://mapserver.org/mapfile/xml_mapfile.html). See https://lists.osgeo.org/pipermail/mapserver-users/2026-June/083830.html for further details. 

The following options will then no longer be supported. None of these are used by the default XSL file at https://github.com/MapServer/MapServer/blob/main/src/xmlmapfile/mapfile.xsl. 

```
<!DOCTYPE map [
  <!ENTITY common SYSTEM "common.xml">
]>
<!DOCTYPE Layer [
  <!ATTLIST Layer type CDATA "raster">
]>
<Layer/>

```

### DIFF
--- a/src/mapxml.c
+++ b/src/mapxml.c
@@ -55,7 +55,7 @@ int msTransformXmlMapfile(const char *stylesheet, const char *xmlMapfile,
     goto done;
   }
 
-  doc = xmlReadFile(xmlMapfile, NULL, XML_PARSE_NOENT | XML_PARSE_DTDLOAD);
+  doc = xmlReadFile(xmlMapfile, NULL, 0);
   if (doc == NULL) {
     msSetError(MS_MISCERR, "Failed to load xml mapfile",
                "msTransformXmlMapfile()");


rouault:
I don't think this change is really needed given that the XSL is fully under the control of the administrator. If they voluntary shoot themselves in the foot that's their responsibility :-)

geographika:
Yes, if there is no feedback I'll probably just close this. I've no idea if accessing remote/network XSL is a common workflow. We could also consider dropping XML Mapfile support if it gradually loses parity with new Mapfile features.

rouault:
> I've no idea if accessing remote/network XSL is a common workflow

I'd bet nobody uses that.  That PR is likely *technically* not needed, but on reflection, we might still want to merge it, since that will save us from double ,triple, etc... similar reports in the future... Or the alternative would be to add a comment line before saying that ``XML_PARSE_NOENT | XML_PARSE_DTDLOAD`` is fine in this context of use

rouault:
merging. Use safe parameters will avoid later reports, and it is unlikely to cause regression

PR #7527, opened 2026-06-15
Set OpenLayers resolution on CGI output maps

geographika:
The OpenLayers viewer template uses a `ol.View()`, which defaults to `EPSG:3857` with the min/max resolution range for Web Mercator. 

Layers with extents in other units (e.g. `EPSG:4326` degrees, or small local extents) hit the default `minResolution`, which prevents zooming further in. This PR:

- Calculates minResolution/maxResolution from the map's extent and sets them on the view, so the resolution range matches the data regardless of projection.
- Moves the addProjection/ol.proj.get registration into the shared template so it applies to BROWSE mode as well as WMS requests. This prevents JS errors for unknown projections. 
- For BROWSE mode, gets the projection from the Mapfile using `msOWSGetEPSGProj()` (as there is no CRS/SRS request param).
- Defaults to EPSG:3857 if no projection can be determined (the current behaviour). 
- Updates the test suite

### DIFF
--- a/src/maptemplate.c
+++ b/src/maptemplate.c
@@ -43,6 +43,7 @@
 
 #include <assert.h>
 #include <ctype.h>
+#include <math.h>
 
 static inline void IGUR_sizet(size_t ignored) {
   (void)ignored;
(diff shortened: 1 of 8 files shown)


src/maptemplate.c
```diff
@@ -5013,6 +5018,7 @@ int msReturnOpenLayersPage(mapservObj *mapserv) {
   const char *openlayersCssUrl = olCssUrl;
   char *projection = NULL;
   char *format = NULL;
+  char *epsgCode = NULL;
```
rouault:
```suggestion
```

src/maptemplate.c
```diff
@@ -5045,6 +5051,14 @@ int msReturnOpenLayersPage(mapservObj *mapserv) {
     }
   }
 
+  if (mapserv->Mode == BROWSE) {
+    msOWSGetEPSGProj(&(mapserv->map->projection), NULL, NULL, MS_TRUE,
```
rouault, line 5055:
```suggestion
    char *epsgCode = NULL;
    msOWSGetEPSGProj(&(mapserv->map->projection), NULL, NULL, MS_TRUE,
```

src/maptemplate.c
```diff
@@ -5045,6 +5051,14 @@ int msReturnOpenLayersPage(mapservObj *mapserv) {
...
+                     &epsgCode);
+    if (epsgCode) {
+      projection = epsgCode;
```
rouault, line 5059: (added that silly looking comment because when reviewing I initially though this file was a .cpp one and projection a std::string, and thus a memory leak of epsgCode would occur...)

```suggestion
      // Transfers ownership of epsgCode to projection
      projection = epsgCode;
```

rouault (approved):

jmckenna:
thanks @geographika !

PR #7533, opened 2026-06-19
Option to add GDAL/OGR Feature Id field to LAYER attributes

geographika:
Both GDAL and QGIS show the FID column when opening through GDAL. For example the *fid* column below:

```bash
gdal info .\buildings.gpkg
INFO: Open of `.\buildings.gpkg'
      using driver `GPKG' successful.
Layer name: buildings
Geometry: Multi Polygon
...
FID Column = fid
Geometry Column = geom
osm_id: String (0.0)
name: String (0.0)
building: String (0.0)
```

MapServer however does not make this field available as an attribute. For example:

```
LAYER
  NAME "buildings"
  # interestingly adding fid below also makes the field available, but the "all" option cannot then be used
  # PROCESSING "ITEMS=fid,building"
  # new PROCESSING  keyword added in this pull request
  PROCESSING "OGR_EXPOSE_FID=TRUE"
  CONNECTIONTYPE OGR
  CONNECTION "data/osm/timisoara/buildings.gpkg"
  METADATA
    "ows_title" "Buildings"
    "gml_include_items" "all"
    "gml_featureid" "fid"
    "gml_types" "auto"
    "oga_queryable_items" "all"
END
```

Without the changes in this pull request a query such as http://localhost:7000//timisoara/ogcapi/collections/buildings/items?f=json

will return the following error (as `"gml_featureid" "fid"` is not found):

```
{"code":"ServerError","description":"Error getting feature. Feature id not found."}
```

With the new `PROCESSING "OGR_EXPOSE_FID=TRUE"` the FID is now available, and can be used as the unique Id field, as labels, expressions etc. 

I'm not a fan of adding more and more processing keywords, but adding the FID by default breaks 100s of tests, and therefore possibly live MapServer deployments. 

As I understand it FID columns in OGR are always 64-bit integers, so "Long" can be set as the default type. 

I had a look through the code and I don't think there is an existing way to add these feature Ids so they are available to MapServer  apart from manually adding the field name in `PROCESSING "ITEMS=fid,building"` as shown in the example above.

### DIFF
--- a/src/mapogr.cpp
+++ b/src/mapogr.cpp
@@ -2681,7 +2681,7 @@ static void msOGRPassThroughFieldDefinitions(layerObj *layer,
   if (exposeFID != NULL && CSLTestBoolean(exposeFID)) {
     const char *fidColumn = OGR_L_GetFIDColumn(psInfo->hLayer);
     if (fidColumn != NULL && fidColumn[0] != '\0')
-      +msUpdateGMLFieldMetadata(layer, fidColumn, "Long", "", "", 0);
+      msUpdateGMLFieldMetadata(layer, fidColumn, "Long", "", "", 0);
   }
 
   /* Should we try to address style items, or other special items? */
--- a/src/mapogr.cpp
+++ b/src/mapogr.cpp
@@ -2709,7 +2709,7 @@ static char **msOGRFileGetItems(layerObj *layer, msOGRFileInfo *psInfo) {
   const char *fidColumn = OGR_L_GetFIDColumn(psInfo->hLayer);
   const char *exposeFID = msLayerGetProcessingKey(layer, "OGR_EXPOSE_FID");
   const bool hasFID = (fidColumn != NULL && fidColumn[0] != '\0') &&
-               (exposeFID != NULL && CSLTestBoolean(exposeFID));
+                      (exposeFID != NULL && CSLTestBoolean(exposeFID));
 
   totalnumitems = numitems = OGR_FD_GetFieldCount(hDefn);
 
--- a/src/mapogr.cpp
+++ b/src/mapogr.cpp
@@ -2676,6 +2676,12 @@ static void msOGRPassThroughFieldDefinitions(layerObj *layer,
                              0);
   }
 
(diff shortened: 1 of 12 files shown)


src/mapogr.cpp
```diff
@@ -2698,8 +2706,17 @@ static char **msOGRFileGetItems(layerObj *layer, msOGRFileInfo *psInfo) {
     return NULL;
   }
 
+  const char *fidColumn = OGR_L_GetFIDColumn(psInfo->hLayer);
+  const char *exposeFID = msLayerGetProcessingKey(layer, "OGR_EXPOSE_FID");
+  int hasFID = (fidColumn != NULL && fidColumn[0] != '\0') &&
```
rouault:
```suggestion
  const bool hasFID = (fidColumn != NULL && fidColumn[0] != '\0') &&
```

rouault (commented):
I'd like to see an added test with a WFS FILTER query on the fid field to make sure no other code path needs to be updated

geographika:
WFS filter tests added, which highlight a few things. 

- WFS supports FeatureId queries without having to have the "fid" exposed. See the test with `&REQUEST=GetFeature&TYPENAMES=popplace_gpkg_nofid&FILTER=<fes:Filter+xmlns:fes='http://www.opengis.net/fes/2.0'><fes:ResourceId+rid='popplace_gpkg_nofid.1'/></fes:Filter>" `
- However the `wfs_featureid` needs to be set in the metadata or the following error is returned `<!-- WARNING: No featureid defined for typename 'popplace_gpkg_nofid'. Output will not validate. -->`. 
- Adding `PROCESSING "OGR_EXPOSE_FID=TRUE"` will expose the `fid` as a standard property, but the GML Ids are also in the output. 

When using OGC Features API, the `fid` needs to be available so it can be added to the queryables and sortables. My use case that highlighted this was an OSM export where the `osm_id` is not unique, but GDAL reports there is a unique `fid`, but it is not usable by MapServer. Obviously a workaround is to add a new field, but it seems as though MapServer should be as similar as possible to GDAL (and QGIS).

rouault (approved):

PR #7534, opened 2026-06-19
Add filtering and sorting to columns and fix page size logic in Bootstrap Templates

geographika:
Adds sortable headers to the collection-items template, with arrows to indicate sorting:

image

Also in the PR:

- Fix page size selector (duplicated value of 10) and sync with URL limit
- Fix pagination logic (offset / limit)
- Reset offset value on page size/sort change
- Guard `fitBounds()` for empty GeoJSON


rouault (approved):

geographika:
Templates updated further to allow sorting on multiple columns, and basic filtering. One column can be filtered at a time - the previous filters are not loaded back into the filter boxes (we'd need to parse the CQL in JS..hopefully a library will become available to do that). 

This functionality uses the new OGC Features API Part 3 (filtering) and 5 (sorting) to implement these features in HTML and JS. 

image

PR #7525, opened 2026-06-11
OSS-Fuzz: Add new fuzzer targets ows

arthurscchan:
This PR creates a new OSS-Fuzz fuzzer targets OWS.

PR #7539, opened 2026-06-24
Bug: Fix memory leak on param handling

arthurscchan:
When a parameter is matched, its value is duplicated and stored in the param object, overwriting any existing value without first freeing the previously allocated pointer. This causes a direct memory leak whenever the same parameter name appears more than once in a request.

This PR fixes the leak by always clearing the target pointer before assigning the new value during each parameter's processing. The change is safe even for null pointers, as msFree() accepts a null pointer and is a no-op in that case.

This memory leak is discovered by the new fuzzer introduced in #7525.

### DIFF
--- a/msautotest/wxs/wcs_simple.map
+++ b/msautotest/wxs/wcs_simple.map
@@ -23,7 +23,8 @@
 #
 # Capabilities 1.0.0
 # RUN_PARMS: wcs_cap.xml [MAPSERV] QUERY_STRING="map=[MAPFILE]&SERVICE=WCS&VERSION=1.0.0&REQUEST=GetCapabilities" > [RESULT_DEVERSION] [RESULT_DEMIME]
-# RUN_PARMS: wcs_cap.txt [MAPSERV] QUERY_STRING="map=[MAPFILE]&SERVICE=WCS&VERSION=1.0.0&REQUEST=GetCapabilities" > [RESULT_DEVERSION] [RESULT]
+# Also test duplicated parameter not to cause memory leak
+# RUN_PARMS: wcs_cap.txt [MAPSERV] QUERY_STRING="map=[MAPFILE]&SERVICE=WCS&VERSION=1.0.0&VERSION=1.0.0&REQUEST=GetCapabilities" > [RESULT_DEVERSION] [RESULT]
 #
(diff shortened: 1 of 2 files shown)


geographika:
I checked how this was handled for other services. For WFS a helper function is used:

```cpp
static int msWFSSetParam(char **ppszOut, cgiRequestObj *request, int i,
                         const char *pszExpectedParamName) {
  if (strcasecmp(request->ParamNames[i], pszExpectedParamName) == 0) {
    if (*ppszOut == NULL)
      *ppszOut = msStrdup(request->ParamValues[i]);
    return 1;
  }
  return 0;
}


//        if (msWFSSetParam(&(wfsparams->pszVersion), request, i, "VERSION")) {}
```

This has slightly different logic as it keeps the first parameter supplied in the querystring. We should also add some tests for duplicated parameters.

arthurscchan:
I have changed the logic mirror the msWFSSetParam approach.

rouault:
just pushed in https://github.com/arthurscchan/MapServer/tree/fix-memleak a modified msautotest check exercising duplicated parameters in WCS queries

msautotest/wxs/wcs_simple.map
```diff
@@ -23,7 +23,8 @@
 #
 # Capabilities 1.0.0
 # RUN_PARMS: wcs_cap.xml [MAPSERV] QUERY_STRING="map=[MAPFILE]&SERVICE=WCS&VERSION=1.0.0&REQUEST=GetCapabilities" > [RESULT_DEVERSION] [RESULT_DEMIME]
-# RUN_PARMS: wcs_cap.txt [MAPSERV] QUERY_STRING="map=[MAPFILE]&SERVICE=WCS&VERSION=1.0.0&REQUEST=GetCapabilities" > [RESULT_DEVERSION] [RESULT]
+# Also test duplicated parameter not to cause memory leak
+# RUN_PARMS: wcs_cap.txt [MAPSERV] QUERY_STRING="map=[MAPFILE]&SERVICE=WCS&VERSION=1.0.0&VERSION=1.0.0&REQUEST=GetCapabilities" > [RESULT_DEVERSION] [RESULT]
```
jratike80, line 27: Should we test also `&VERSION=1.0.0&VERSION=2.0.0, if the aim is to keep just the first parameter value?

geographika: Yes probably best to have a test that confirms this. I'm not sure what the spec says about this case.

PR #7545, opened 2026-07-01
validate property value bounds in flatgeobuf_decode_properties

nvxbug:
AddressSanitizer, on a crafted .fgb feature whose string property length field claims more bytes than the property blob holds:

    READ of size 65536 at 0x60200000... thread T0
        #0 __asan_memcpy
        #1 flatgeobuf_decode_properties src/flatgeobuf/flatgeobuf_c.cpp:229
    0x602000... is located 0 bytes after 6-byte region allocated by msSmallMalloc

flatgeobuf_decode_properties walks a feature property blob taken straight from the file (feature->properties()->data()/size()). The loop only checks that the 2-byte column index fits; the value payload is never bounded against size. For fixed-width columns parse_value<T> dereferences up to 8 bytes past the buffer, and for string/datetime the 4-byte len is read with no cap, so msSmallMalloc(len + 1) wraps to 0 when len is 0xFFFFFFFF and the following memcpy runs off the heap.

Pass size into parse_value and bound both the fixed-width read and the string len against the remaining bytes before touching data.


jmckenna:
Thanks @nvxbug.  Please also include a test inside `/msautotest/` as part of your pull request (see steps for msautotest: https://mapserver.org/development/tests/autotest.html ), there is already tests there (in `/msautotest/misc/`) with other `*.fgb` files, it would be good if you can create a tiny `.fgb` to reproduce your issue, and include it in a test.

nvxbug:
Added a test under msautotest/misc/. `data/flatgeobuf-invalid.fgb` is a one-feature file (written with ogr2ogr, then the string property's 4-byte length field patched to claim more bytes than the 8-byte property blob holds), and `flatgeobuf-invalid.map` runs a WFS GetFeature against it. GDAL's own FlatGeobuf reader rejects that same file, so it exercises exactly the path this PR guards.

While wiring it up I found that `flatgeobuf_decode_feature` ignored the return of `flatgeobuf_decode_properties`, so the crafted feature was still handed back with an unset value and segfaulted downstream on the query path. Propagating that failure is in the same commit, and GetFeature now returns a clean service exception (see the expected output). Passes locally via `run_test.py`.

jratike80:
This may interest also @bjornharrtell

bjornharrtell:
Thanks @jratike80, I've made a fix for similar issue in upstream at https://github.com/flatgeobuf/flatgeobuf/pull/505.

PR #7558, opened 2026-07-07
msOracleSpatialLayerTranslateFilter(): avoid SQL injections

rouault:
Fixes https://github.com/MapServer/MapServer/security/advisories/GHSA-m632-88w4-634j

Funded by RGD Savoie Mont-Blanc

Note: this has only been compile-time tested, not runtime tested


rouault:
@msmitherdc That could receive some runtime testing, as I haven't.

PR #7559, opened 2026-07-07
Python mapscript: no longer use Python 2 functions whose compatibility layer has been removed in SWIG master 4.5dev

rouault:


jmckenna:
Thanks @rouault (you're faster than me, I was trying syntax to add at the top of `pyextend.i` to handle both Python2 and 3 SWIG API calls).

For other readers (in case / when this error comes up again), the initial error `missing __init__.py` is often a "red herring", the problem most likely is caused by a problem when trying to load `_mapscript.so` by Python mapscript (once you test that locally, you can see the exact import errors, in this case `undefined symbol: PyInt_FromLong`

PR #7560, opened 2026-07-07
Skip centerline for polygons too simple to compute

### LINKED ISSUE #7058: centerline function causes GEOS errors

loom-agent:
This pull request was prepared autonomously by Loom Agent, an AI agent under
human supervision; feedback is welcome and I will address review comments promptly.

Closes #7058.

**Problem.** `GEOMTRANSFORM (centerline([shape]))` on a polygon that is too simple to
support a Voronoi centerline (e.g. a convex polygon, or one with few vertices) raised
`msGEOSCenterline(): Centerline generation failed, try densifying the shapes.` and
**aborted the entire map image** rather than just omitting that one feature's centerline.

**Root cause.** In `msGEOSCenterline()` (`src/mapgeos.c`), the "no Voronoi nodes" case
(`nodes.numpoints == 0`) and the "no path" case (`!path`) both called `msSetError()` and
returned `NULL`, which propagates as a parser error up through `msGeomTransformShape()` /
`msDrawTransformedShape()` and fails the layer draw.

**Fix.** Per the direction Steve floated in #7058 ("Skipping the features, rather than
throwing an error might be best"), these two runtime-generation branches now skip the
centerline for the offending feature: they return the already-allocated empty line
(`shape2`, reset at line 1320 then typed `MS_SHAPE_LINE`) and emit a `msDebug()` note
pointing the user at `densify()`. Downstream already handles the empty shape:
`msDrawShape` (`mapdraw.c:2375`) and `msDrawLineSymbol` (`maprendering.c:612`) both
early-return on `numlines == 0`, so the feature's other styles still draw and the map
completes. Input-validation errors (non-polygon input, negative densify tolerance,
exercised by `centerline4`/`centerline5`) are intentionally left as hard errors.

This is distinct from the rejected auto-densify PR #7060 (which looped for 10+ minutes on
remote data): it does not fetch or densify anything, it only stops aborting.

**Testing.** Reproduced MapServer's actual CI environment on the CI reference platform
(`ubuntu:24.04`, the base image CI uses, plus the same `ubuntugis/ubuntugis-unstable` PPA
that `ci/ubuntu/setup.sh` installs, giving GEOS 3.12.2, GDAL 3.11.4, PROJ 9.4.1). Built
`WITH_GEOS=ON` clean (exit 0) with CI's exact flags `-Wall -Werror -Wextra` and CMake
`-Werror=dev`; `mapserv -v` reports `SUPPORTS=GEOS`.
```bash
cd msautotest
python -m pytest misc/centerline.map -v --strict_mode        # CI mode: all 6 pass
python -m pytest misc/centerline.map -v --run_under_asan      # memory-safety: 0 ASan/LSan findings
```
`--strict_mode` is the CI default (`AUTOTEST_OPTS?=--strict_mode` in the Makefile). With it,
all 6 centerline tests pass on GEOS 3.12.2, including our changed goldens `centerline3.png`
and `centerline6.png`: those were generated on GEOS 3.12.1 and still match on 3.12.2, which
confirms the too-simple polygons render only the base layer (the centerline is skipped, so the
golden is GEOS-version-stable). Reverting `src/mapgeos.c` makes `centerline3` **and** the new
`centerline6` fail with `AttributeError: 'NoneType'`; restoring passes both. Before the fix
`centerline6` aborts with the GEOS error and yields no image; after, it renders the square's
fill/outline with no centerline stroke and no exception. `centerline4`/`centerline5` keep
failing exactly as today (validation errors unchanged). The full `misc` suite shows no new
failures vs. clean main (the pre-existing failures are environmental: PostGIS/OGR/cluster
tests that need data CI provisions).

**Notes.** `centerline3` flips from an expected exception to a rendered image, so its
`_exception.txt` is removed and a new `centerline3.png` golden is included. Both new goldens
were generated on the CI platform (Ubuntu 24.04, `ci/ubuntu/setup.sh`).


rouault:
Closing. Too chatty PR text. See https://mapserver.org/community/ai_tool_policy.html

rouault:
fixed link: https://mapserver.org/community/ai_tool_policy.html

PR #7564, opened 2026-07-07
AGG: avoid Bezier subdivision hang on non-finite points (fixes #7310)

### LINKED ISSUE #7310: Mapserver hangs in a (high) CPU loop - projection issue msProjectRect
Mapserver hangs in a (high) CPU loop when the below conditions are met:

Mapserver version: 8.4.0 (older versions might also be affected, not tested)

`"QUERY_STRING=SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&BBOX=50.21289999999999765,2.527130000000000098,55.72120000000000317,7.374030000000000307&CRS=EPSG:4326&WIDTH=1&HEIGHT=2&LAYERS=station&STYLES=,,,,,,&FORMAT=image%2Fpng&DPI=120&MAP_RESOLUTION=120&FORMAT_OPTIONS=dpi:120&TRANSPARENT=TRUE"`

- Bounding boxes: 50.21289999999999765,2.527130000000000098,55.72120000000000317,7.374030000000000307 or BBOX=50.21289999999999765,2.527130000000000098,55.72120000000000317,6.374030000000000307
- WIDTH=1
- LAYERS=station
- CRS=EPSG4326 (WGS84) / EPSG:4258

We do not see the issue occur with a slightly larger bounding box:

[Issue.zip.zip](https://github.com/user-attachments/files/21415014/Issue.zip.zip)

50.21289999999999765,2.527130000000000098,55.72120000000000317,8.374030000000000307

Observed issues:

The above results in mapserver hanging in a loop on mapserver::curve3_div::recursive_bezier

Remarkable is that the parameter values x and y of the functions: calc_sq_distance and recursive_bezier are called with inf values.
`(gdb) bt
#0  0x00007ffff7df666a in mapserver::calc_sq_distance (x1=inf, y1=inf, x2=-nan(0x8000000000000),
    y2=-nan(0x8000000000000)) at /usr/local/src/mapserver/src/renderers/agg/src/../include/agg_math.h:69
#1  0x00007ffff7df498e in mapserver::curve3_div::recursive_bezier (this=0x7fffffffdeb8, x1=inf, y1=inf, x2=inf,
    y2=inf, x3=inf, y3=inf, level=32) at /usr/local/src/mapserver/src/renderers/agg/src/agg_curves.cpp:213
#2  0x00007ffff7df4a99 in mapserver::curve3_div::recursive_bezier (this=0x7fffffffdeb8, x1=inf, y1=inf, x2=inf,
    y2=inf, x3=inf, y3=inf, level=31) at /usr/local/src/mapserver/src/renderers/agg/src/agg_curves.cpp:225
#3  0x00007ffff7df4a43 in mapserver::curve3_div::recursive_bezier (this=0x7fffffffdeb8, x1=inf, y1=inf, x2=inf,
    y2=inf, x3=inf, y3=inf, level=30) at /usr/local/src/mapserver/src/renderers/agg/src/agg_curves.cpp:224
#4  0x00007ffff7df4a99 in mapserver::curve3_div::recursive_bezier (this=0x7fffffffdeb8, x1=inf, y1=inf, x2=inf,`
(This behaviour looks somewhat similar to https://github.com/MapServer/MapServer/issues/4908)

Even though there are no features at the location of the given bounding box, Mapserver is trying to write labels. This is caused by an incorrect search extent.
We see the search extent 50.21289999999999765, 2.527130000000000098, 55.72120000000000317, 7.374030000000000307 (EPSG:4326) (a blue bbox, overlapped by a yellow bbox) being reprojected to -49159.289932862273, 246832.71246760915, 296784.56947335071, 863479.57403289562 (EPSG:28992) (the red bbox in the image below) in: 
[https://github.com/MapServer/MapServer/blob/7d24be7d59c7bbf6ddc173b6ca80af0b9d6aec9a/src/mapproject.c#L1867](https://github.com/MapServer/MapServer/blob/7d24be7d59c7bbf6ddc173b6ca80af0b9d6aec9a/src/mapproject.c#L1867). These two BBOXes do not overlap. It looks as if the axis of the WGS84 bbox becomes flipped while reprojected; the EPSG:28992 BBOX is the same as WGS84 BBOX 2.527130000000000098, 50.21289999999999765, 7.374030000000000307, 55.72120000000000317 reprojected to EPSG:28992. For the yellow bbox in the image below 50.21289999999999765, 2.527130000000000098, 55.72120000000000317, 8.374030000000000307 this does not occur.

Image
Visual representation of the bounding boxes 

Blue  = Bounding box related to the issue for which the search extent becomes flipped.
Yellow = Greater bounding box. Issue is NOT triggered.
Red = Extent of the reprojected search extent and inverse of the blue bounding box.

loom-agent:
**Issue:** A degenerate WMS GetMap (`WIDTH=1&HEIGHT=2` with a reprojected search
extent) feeds `inf`/`-nan` into the AGG Bezier subdivision
(`curve3_div`/`curve4_div`). Non-finite control points defeat every early-exit
in `recursive_bezier`, so it recurses to the level cap on every branch and never
converges - the 100% CPU hang reported in #7310.

**Fix:** Add an `isfinite()` guard next to the existing recursion limit in both
`curve3_div` and `curve4_div`.

**Why:** A non-finite control point has no meaningful subdivision, so the only
correct action is to stop; all finite input takes the original path unchanged.

**Test:** `tests/unit/test.cpp` feeds inf/nan points into both subdividers and
asserts termination; without the guard `unit_test` hangs (caught by a
`ctest TIMEOUT 60`).

Fixes #7310

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1234,4 +1234,7 @@ if(BUILD_DYNAMIC AND BUILD_TESTING)
     add_executable(unit_test tests/unit/test.cpp)
     target_link_libraries(unit_test PRIVATE mapserver)
     add_test(NAME unit_test COMMAND unit_test)
+    # Bounds the recursion-regression case (testAggCurvesNonFiniteControlPoints),
+    # which would otherwise spin if the guard in agg_curves.cpp is removed.
+    set_tests_properties(unit_test PROPERTIES TIMEOUT 60)
 endif()
--- a/src/renderers/agg/src/agg_curves.cpp
+++ b/src/renderers/agg/src/agg_curves.cpp
@@ -137,12 +137,21 @@ namespace mapserver
     }
 
     //------------------------------------------------------------------------
-    void curve3_div::recursive_bezier(double x1, double y1, 
-                                      double x2, double y2, 
+    void curve3_div::recursive_bezier(double x1, double y1,
+                                      double x2, double y2,
                                       double x3, double y3,
                                       unsigned level)
     {
-        if(level > curve_recursion_limit) 
+        if(level > curve_recursion_limit)
+        {
+            return;
+        }
+
+        // Non-finite control points defeat every early-exit test below, so the
+        // subdivision runs to the recursion limit on every branch (up to 2^32
+        // calls for a quadratic) and never converges (#7310).
+        if(!isfinite(x1) || !isfinite(y1) || !isfinite(x2) || !isfinite(y2) ||
+           !isfinite(x3) || !isfinite(y3))
         {
             return;
         }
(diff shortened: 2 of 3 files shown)


geographika:
See https://mapserver.org/community/ai_tool_policy.html

PR #7565, opened 2026-07-07
Make msRasterizeArc buffer size non-static for thread safety

### LINKED ISSUE #6507: App crash in msRasterizeArc
The stack trace is looking like:

```
mapserver!msMVTWriteTile+0x57027
mapserver!msMVTWriteTile+0x5cfe6
mapserver!msMVTWriteTile+0x56dc4
mapserver!msMVTWriteTile+0x407eb
mapserver+0x272f
mapserver!msRasterizeArc+0x9bf
mapserver!msAddLabelGroup+0x329
mapserver!msValueToRange+0xf64
mapserver!msDrawShape+0x7b5
mapserver!msDrawVectorLayer+0x804
mapserver!msDrawLayer+0x3a3
mapserver!msDrawMap+0x415
mapscript!CSharp_mapObj_draw+0xd
```

and

```
mapserver!msRasterizeArc+0x9ea
mapserver!msPolylineLabelPath+0x125
mapserver!msUnionLayerInitializeVirtualTable+0x1646
mapserver!msDrawShape+0x80d
mapserver!msDrawVectorLayer+0x804
mapserver!msDrawLayer+0x3a3
mapserver!msDrawMap+0x415
mapscript!CSharp_mapObj_draw+0xd
```

I guess this is a multithreading problem since msRasterizeArc uses a static variable (allocated_size).

loom-agent:
**Issue:** `msRasterizeArc()` keeps its working-buffer capacity in a `static int allocated_size`, so concurrent callers (e.g. threaded C# MapScript under load) race on it. A thread can read a size grown by another thread, skip its own realloc, and write past its per-call buffer, which is the heap corruption/crash reported in #6507 (stack traces in `msRasterizeArc` / `msPolylineLabelPath`).

**Fix:** Make `allocated_size` a plain local. The buffer it sizes (`line->point`) is already allocated per call, so the `static` only cached the integer across calls; dropping it makes the function reentrant. This is the fix @szekerest prescribed in the issue.

**Why:** Each call now manages its own capacity and there is no shared mutable state. The only cost is that a large arc no longer leaves a larger starting buffer for the next call, which is negligible and is the only correct behavior under threads.

**Test:** `tests/unit/test.cpp::testRasterizeArcThreadSafety` renders arcs large enough to grow the buffer from 8 threads. Under ASan the pre-fix code aborts with a heap-buffer-overflow in `msRasterizeArc` (mapgeomutil.cpp); under TSan it reports a data race on `allocated_size`. With this change both are clean. `unit_test` runs in CI through `ctest`, including the ASan job.

Fixes #6507


geographika:
See https://mapserver.org/community/ai_tool_policy.html

PR #7571, opened 2026-07-08
msDBFOpenVirtualFile(): validate field extents against record length

nvxbug:
## What does this PR do?

AddressSanitizer, reading the first attribute of a shapefile whose `.dbf` header declares a field wider than its records:

    ==ERROR: AddressSanitizer: heap-buffer-overflow ... READ of size 10
        #1 msDBFReadAttribute mapxbase.c:647
        #2 msDBFReadStringAttribute mapxbase.c:715
    0x... is located 0 bytes after 10-byte region
        allocated by msDBFOpenVirtualFile mapxbase.c:178

`msDBFOpenVirtualFile()` takes `nRecordLength` from the header (bytes 10-11) to size `pszCurrentRecord`, but each field's `panFieldSize`/`panFieldOffset` come from the field descriptors with no check that they fit inside the record. `msDBFReadAttribute()` later `strncpy`s `panFieldSize` bytes from `pabyRec + panFieldOffset`, so a header whose fields do not fit the record reads past the record buffer on any attribute access (WFS GetFeature, joins, attribute-driven labels).

Reject the file at open time when the last field's offset+size exceeds `nRecordLength`, the same layout check shapelib applies. Valid `.dbf` files satisfy `offset+size == nRecordLength` for the last field, so they are unaffected. The added unit test covers both the rejected and the accepted case.

None.

## AI tool usage

## Tasklist

 - [x] Add test case(s)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments


nvxbug:
any update?

jmckenna:
@nvxbug please include a test for this in /msautotest (use an existing map or create one), follow https://mapserver.org/development/tests/autotest.html

(it seems I have to keep repeating this to you)

nvxbug:
Added misc/dbf-invalid.map, modeled on flatgeobuf-invalid.map: a WFS GetFeature against a shapefile whose .dbf declares a 250-byte field in a 10-byte record, expecting the rejection at layer open. Ran it through the pytest harness locally: it fails on an unpatched build (the response contains attribute bytes read past the record buffer) and passes with the fix. Noted for future PRs, will include an msautotest case up front.

jmckenna (approved):

jmckenna:
thanks @nvxbug !

jratike80:
@rouault 

GDAL does not notice anything special in dbf-invalid.shp. Should it?

```
gdal info dbf-invalid.shp --features
INFO: Open of `dbf-invalid.shp'
      using driver `ESRI Shapefile' successful.

Layer name: dbf-invalid
Geometry: Point
Feature Count: 1
Extent: (15.000000, 15.000000) - (15.000000, 15.000000)
Layer Coordinate Reference System: none
OGRFeature(dbf-invalid):0
  POINT (15 15)
```

rouault:
> GDAL does not notice anything special in dbf-invalid.shp. Should it?

it silently ignores it, executing the same check as in this PR. warning added in https://github.com/OSGeo/gdal/pull/15123

PR #7572, opened 2026-07-09
msautotest: adjust result for wms_openlayers_xss test

jmckenna:
- try to fix recent `branch-8-6` test failures ( sample failing [run ](https://github.com/MapServer/MapServer/actions/runs/28909805971/job/85764650558))


jmckenna:
@MarkLee131 do you see a reason why `branch-8-6` result for `wms_openlayers_xss.map` differs from the result in `main`?

(related to #7544 )

jmckenna:
It could be from changes in `maputil.c` in `main`, that are not in `branch-8-6`.

PR #7575, opened 2026-07-11
Fix WCS 2.0 GetCoverage DoS from non-positive RESOLUTION (GHSA-6jr5-rc9c-p3cj)

sdjasj:
## What does this PR do?
Fix:
1. msWCSParseResolutionString20(): After parsing, reject non-finite values and values <= 0 with InvalidParameterValue. This mirrors the existing validation in msWCSParseScaleString20() which already rejects scale <= 0.
2. msWCSGetCoverage20(): Add a defensive guard that rejects width <= 0 or height <= 0 before the existing MAXSIZE check. This catches any edge case where a non-positive dimension reaches this point.

Tests added in msautotest/wxs/wcs_simple.map:
- Negative RESOLUTION → InvalidParameterValue (was exit(1))
- Zero RESOLUTION → InvalidParameterValue (was raw alloc error)

https://github.com/MapServer/MapServer/security/advisories/GHSA-6jr5-rc9c-p3cj
## AI tool usage

 - [x] AI supported my development of this PR. See our [policy about AI tool use](https://mapserver.org/community/ai_tool_policy.html). Use of AI tools, if any, *must* be indicated.

## Tasklist

 - [x] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [ ] ADD YOUR TASKS HERE

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -13,6 +13,11 @@ https://mapserver.org/development/changelog/
 
 The online Migration Guide can be found at https://mapserver.org/MIGRATION_GUIDE.html
 
+8.6.6 development
+-----------------
+
+- security: WCS 2.0 GetCoverage: reject non-positive RESOLUTION values to prevent denial of service (GHSA-6jr5-rc9c-p3cj)
+
 8.6.5 release (2026-07-10)
 --------------------------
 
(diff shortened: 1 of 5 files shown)


jratike80:
Sorry for the late response, but as documented in https://mapserver.org/ogc/wcs_server.html

> The following request parameters are not part of the WCS 2.0 Core or any extension thereof, and are kept for backward compatibility. They originate from a time where the most vital standard extensions were not specified and the WCS core was not quite useful without. They are deprecated.
> 
> SIZE=axis(value): This parameter sets the size of the desired axis to the desired value (pixels).
> 
> RESOLUTION=axis(value): This parameter sets the resolution of the desired axis to the desired value (pixels/unit).

An alternative and perhaps better fix would have been to remove the support for these non-standard parameters. Even if the standard scaling options (scaleByFactor, scaleAxisByFactor, scaleToSize, and scaleToExtent) do not directly support setting the desired resolution, but user needs first check the native scale and calculate the scaling factor to apply.

PR #7577, opened 2026-07-13
size msClipPolygonRect buffer for 3 emitted points per edge

nvxbug:
## What does this PR do?

`msClipPolygonRect` sizes each ring's output buffer for `2 * numpoints` points, but the Liang-Barsky inner loop can emit three points for a single edge: the clip-rect corner, the entry intersection, and the exit/endpoint. Over `numpoints - 1` edges plus the closure point that reaches `3 * numpoints - 2` writes, so a ring whose edges repeatedly cross the clip-rect corners runs past the allocation. The clipped geometry comes from the layer data source and this sits on the ordinary render path (GetMap in `mapdraw`, MVT tiles in `mapmvt`, query templating in `maptemplate`, hit testing), so an attacker-supplied polygon is enough to reach it.

Size the buffer for the real worst case instead. Keeping the bound inside the clipper means callers don't each have to guess how many points an edge can produce.

`msautotest/misc/clip_polygon_corners.map` renders two polygons whose edges cut across the corners of the map extent, so `map2img` reaches the clipper through `msDrawShape` the same way a GetMap request does. Their rings are 6 and 7 points and clip to 14 and 16, past the old `2 * numpoints` bound: against the previous allocation the mapfile aborts under AddressSanitizer (`heap-buffer-overflow` WRITE at `mapprimitive.cpp:765`, allocation at `mapprimitive.cpp:690`), and renders clean afterwards. The rest of the `misc` suite is unaffected by the change.

None.

## AI tool usage

## Tasklist

 - [ ] Add documentation
 - [ ] Review
 - [x] Adjust for comments


nvxbug:
any update?

geographika:
@nvxbug this pull request would ideally include a test case showing how this issue is thrown and how these changes fix it.

nvxbug:
Added one in `tests/unit/test.cpp`. It clips two rings whose edges cut across the corners of the clip rect, which is the case where a single edge emits three points instead of two.

The overflow itself only shows up under a sanitizer, but that works out here: `make cmakebuild` ends in `ctest -V`, and the ASAN job reaches `cmakebuild` through `asan_compatible_tests`, so the new test runs under ASAN in CI. Against the old allocation it aborts:

```
==2967==ERROR: AddressSanitizer: heap-buffer-overflow
WRITE of size 8 at 0x6140000019c0 thread T0
    #0 msClipPolygonRect mapprimitive.cpp:765
    #1 testClipPolygonRect() test.cpp:178
    #2 main test.cpp:200

0x6140000019c1 is located 0 bytes after 385-byte region [0x614000001840,0x6140000019c1)
allocated by thread T0 here:
    #1 msSmallMalloc maputil.c:2544
    #2 msClipPolygonRect mapprimitive.cpp:690
```

With the patch it runs clean. So that a plain non-ASAN build still catches a regression, the test also asserts the output point count exceeds the old bound: the 6-point ring clips to 14 points against the old capacity of 2*numpoints = 12, and the 7-point ring to 16 against 14. It checks the clipped points land inside the rect too, so the sizing change isn't quietly altering output.

geographika:
@nvxbug MapServer uses it's own test suite detailed at https://mapserver.org/development/tests/autotest.html the test.cpp is for very specific cases. Any bug should have a Mapfile and data which can reproduce the issue, showing it is a bug that could be encountered in normal usage.

nvxbug:
Understood, moved it over. Dropped the test.cpp case and added `msautotest/misc/clip_polygon_corners.map` instead, with the expected PNG.

It's a plain polygon layer rendered with map2img, no special API use. The two rings have edges that cut across the corners of the clip rectangle, which is the case where one edge contributes three points instead of two, so they clip to 14 and 16 points from 6 and 7. Against the old `2 * numpoints` allocation that walks off the end:

```
==98879==ERROR: AddressSanitizer: heap-buffer-overflow
WRITE of size 8 at 0x6140000021c0 thread T0
    #0 msClipPolygonRect mapprimitive.cpp:765
    #1 msDrawShape mapdraw.c:2555
    #2 msDrawVectorLayer mapdraw.c:1390
    #3 msDrawLayer mapdraw.c:984
    #4 msDrawMap mapdraw.c:508
    #5 main map2img.c:363

0x6140000021c1 is located 0 bytes after 385-byte region [0x614000002040,0x6140000021c1)
allocated by thread T0 here:
    #1 msSmallMalloc maputil.c:2544
    #2 msClipPolygonRect mapprimitive.cpp:690
```

So `misc-testcase` fails it in the Address_sanitizer job (that job runs with `--strict_mode --run_under_asan`), and passes with the patch. I ran the test both ways locally to check that.

nvxbug:
Redone as an msautotest case, `msautotest/misc/clip_polygon_corners.map`, and the test.cpp one is gone. It's two polygons whose edges cut across the corners of the map extent, so map2img reaches the clipper through msDrawShape the same way a GetMap does. The rings are 6 and 7 points and clip to 14 and 16, past the old 2*numpoints bound.

Against the previous allocation the mapfile aborts under ASAN, so the Address_sanitizer job catches it:

```
==30188==ERROR: AddressSanitizer: heap-buffer-overflow
WRITE of size 8 at 0x6140000021c0 thread T0
    #0 msClipPolygonRect mapprimitive.cpp:765
    #1 msDrawShape mapdraw.c:2555
    #2 msDrawVectorLayer mapdraw.c:1390
    #3 msDrawLayer mapdraw.c:984
    #4 msDrawMap mapdraw.c:508
    #5 main map2img.c:364

0x6140000021c1 is located 0 bytes after 385-byte region [0x614000002040,0x6140000021c1)
allocated by thread T0 here:
    #1 msSmallMalloc maputil.c:2544
    #2 msClipPolygonRect mapprimitive.cpp:690
```

With the patch it renders clean and matches the expected png. The rest of misc is unchanged: locally I get the same set of unrelated failures with and without the patch (pdf, postgis and font cases my build doesn't cover), and clip_polygon_corners is the only test that flips.

I kept the geometry inline, following line_offset_clipped_to_multiline.map, so the coordinates that trigger it are readable in the mapfile. Happy to move it to a shapefile under misc/data if you'd rather have it come from a real data source.

jmckenna:
thanks @nvxbug !

PR #7578, opened 2026-07-15
bound WMS exception response copy to result_size in msDrawWMSLayerLow

nvxbug:
When a cascaded WMS server returns an XML exception, msDrawWMSLayerLow copies the response into a fixed szBuf with strlcpy(szBuf, result_data, MS_BUFFER_LENGTH). result_data is assembled in msHTTPWriteFct by memcpy of exactly result_size bytes and is never NUL-terminated (the +10000 slack from msSmallMalloc is uninitialized heap). strlcpy scans the source for a NUL to compute its length, so it walks past result_size into uninitialized or adjacent heap, and those bytes then land in the msSetError/msDebug message. A malicious or compromised upstream server sending a NUL-free text/xml body on the default in-memory path triggers it.

Copy result_size bytes and terminate explicitly, the way the sibling result_data consumers a few lines down already do. Keeping the length bound at the copy site is the only place that knows the real byte count, since the buffer past result_size is not part of the response.

### DIFF
--- a/src/mapwmslayer.c
+++ b/src/mapwmslayer.c
@@ -1416,8 +1416,12 @@ int msDrawWMSLayerLow(int nLayerId, httpRequestObj *pasReqInfo, int numRequests,
       } else {
         strlcpy(szBuf, "(Failed to open exception response)", sizeof(szBuf));
       }
+    } else if (pasReqInfo[iReq].result_data) {
+      int nSize = MS_MIN(pasReqInfo[iReq].result_size, MS_BUFFER_LENGTH - 1);
+      memcpy(szBuf, pasReqInfo[iReq].result_data, nSize);
+      szBuf[nSize] = '\0';
     } else {
-      strlcpy(szBuf, pasReqInfo[iReq].result_data, MS_BUFFER_LENGTH);
+      szBuf[0] = '\0';
     }
 
     if (lp->debug)


geographika:
@nvxbug similar to the other pull request an msautotest with a XML response triggering this issue should be added.

nvxbug:
Added one in `msautotest/mspython/test_wms_client.py`. It starts a stub server in a subprocess that answers the cascaded GetMap with a `text/xml` body (it has to be a subprocess rather than a thread, since mapscript holds the GIL for the whole `draw()`).

Two cases:

* a `ServiceExceptionReport` body, where the test asserts the reported error ends exactly where the response body ends, so nothing past `result_size` gets copied in;
* an empty `text/xml` body, which leaves `result_data` NULL and segfaults on the old `strlcpy`. That one fails hard without the patch, which is the deterministic half of the test.

The first case only shows the leaked bytes when the tail of the download buffer isn't zeroed, so on a fresh heap it can pass either way. It's still the right invariant to pin down, and it's what trips under a sanitizer.

Ran the mspython suite locally, only pre-existing failures left are the two epsg2 projection ones from the custom PROJ file not being installed here.

geographika:
@nvxbug - that is a lot of additional (and verbose) Python code. I think I prefer the original commit with just the fix, so maybe roll back to that.

nvxbug:
Makes sense, that test was carrying a lot for what it actually pinned down. Rolled it back, so the PR is just the mapwmslayer.c fix again.

nvxbug:
Makes sense, rolled it back to just the fix commit so the diff is only the mapwmslayer.c bound now.

PR #7579, opened 2026-07-20
Build fixes for CMake 4.4

geographika:
Recent CI builds are failing due to an update to CMake from 4.3.4 to 4.4.0. 

```
2026-07-10T20:51:55.2129877Z cmake version 4.3.4
...
2026-07-20T09:55:31.8651477Z cmake version 4.4.0
```

This caused the following errors:

```
2026-07-20T09:56:55.2318365Z CMake Error (install-absolute-destination) at CMakeLists.txt:1151 (install):
2026-07-20T09:56:55.2319341Z   INSTALL command given absolute DESTINATION path:
```

Looking at the [release notes](https://cmake.org/cmake/help/latest/release/4.4.html#diagnostics):

> The CMD_INSTALL_ABSOLUTE_DESTINATION diagnostic category was added to diagnose install() commands that specify an absolute DESTINATION. 

And the [install](https://cmake.org/cmake/help/latest/command/install.html#command:install) command:

> Specify the directory on disk to which a file will be installed. <dir> should be a relative path. An absolute path is allowed, but not recommended.

The CMakeLists.txt has been updated to use a relative path rather than absolute. Using absolute can break some installs/packaging setups as it doesn't allow the install tree to be moved. In 4.3 this was allowed but throws the error in 4.4.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,6 +72,13 @@ else()
 endif()
 set(INSTALL_CMAKE_DIR ${DEF_INSTALL_CMAKE_DIR} CACHE PATH "Installation directory for CMake files")
 
+# Keep a relative copy for install(EXPORT ...)
+if(IS_ABSOLUTE "${INSTALL_CMAKE_DIR}")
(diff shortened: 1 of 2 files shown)


jmckenna:
thanks @geographika !

Issue #7430, opened 2026-02-11
[Bug] SVG symbols scaling issues during rasterization since version 8.6.0


linea974:
**Description:** Since the installation of MapServer version 8.6.0, SVG symbol processing no longer yields the expected results when rasterized.

**Observed Behavior:**

- The anomaly is visible during rendering via **OpenLayers** and when **exporting to image formats (JPEG)**.
- Interestingly, when the map is exported to **PDF format**, the symbols are processed correctly.
- It appears that the symbol is not being resized correctly during the rasterization process.

**Environment & Versioning:** Tests were performed on the same machine (**Ubuntu 24.04**) using compiled versions of MapServer. Rolling back to version 8.4.1 restores the expected behavior.

**Output of** mapserv -v:

- **Version 8.4.1 (Working)**: MapServer version 8.4.1 PROJ version 9.4 GDAL version 3.11 OUTPUT=PNG OUTPUT=JPEG SUPPORTS=PROJ SUPPORTS=AGG SUPPORTS=FREETYPE SUPPORTS=CAIRO SUPPORTS=SVG_SYMBOLS SUPPORTS=RSVG SUPPORTS=ICONV SUPPORTS=WMS_SERVER SUPPORTS=WMS_CLIENT SUPPORTS=WFS_SERVER SUPPORTS=WFS_CLIENT SUPPORTS=WCS_SERVER SUPPORTS=OGCAPI_SERVER SUPPORTS=FASTCGI SUPPORTS=THREADS SUPPORTS=GEOS INPUT=JPEG INPUT=POSTGIS INPUT=OGR INPUT=GDAL INPUT=SHAPEFILE INPUT=FLATGEOBUF
[...]

PR #7462, opened 2026-03-24
mapcairo.c: allow build with librsvg < 2.46

ness31:
Allow building with librsvg < 2.46 (restoring branch 8.4 behavior)

PR #7583, opened 2026-07-22
mapcairo.c: avoid double scaling of SVG symbols (fixes: #7430)

johanez:
Since 8.6, `rsvg_handle_render_cairo()` (marked as deprecated) was replaced with the newer `rsvg_handle_render_document()`, however this caused scaling issues (#7430). ~This PR implements a simple workaround by reverting back to the legacy function, solving the issues.~
This PR removes a duplicate application of scale when rendering svgs to raster with `librsvg`.

## What does this PR do?

Path symbol size directly to `rsvg_handle_render_document()` without scaling as the scaling is already passed with the viewport parameter, while mainting options to use the deprecated `rsvg_handle_render_cairo()` vor `librvg` versions before v2.46.

- Issue #7430: 
- [PR #7462](https://github.com/MapServer/MapServer/pull/7462) does not solve the issue, but introduces the option to sue older `librsvg` versions with `rsvg_handle_render_cairo()`.

## AI tool usage

 - [ ] ~~AI supported my development of this PR. See our [policy about AI tool use](https://mapserver.org/community/ai_tool_policy.html). Use of AI tools, if any, *must* be indicated.~~

## Tasklist

 - [X] Make sure code is correctly formatted (cf [pre-commit configuration](https://mapserver.org/development/dev_practices.html#commit-hooks))
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [x] All CI builds and checks have passed


johanez:
This fails two CI checks, because the use of the deprecated function `svg_handle_render_cairo()` (which is the base of this workaround).
 - maybe I can add exemption from throwing a build error?
 - else proper a solution with `rsvg_handle_render_document()` might be needed

jmckenna:
@johanez there are actually 2 different failing errors in your tests (click on each of the 2 runs, and you will see what I mean).  For example: https://github.com/MapServer/MapServer/actions/runs/29905134730/job/88913132480?pr=7583

johanez:
@jmckenna I found another possible solution, using the non-deprecated `rsvg_handle_render_document()` and _not_ using `cairo_scale()` for `rsvg` rasterization, as the scaling is already done by setting the viewport.

jmckenna:
@johanez can you join the discussion in #7430 to propose your change /  give feedback for the other proposed change?

jmckenna:
@johanez please also update the title of this pull request, thanks.

johanez:
I updated the test files. As svg rendering was not working correctly, the coordinates used in the test point-identify with anchor test were not correct.

johanez:
Update from my side: I used the solution suggested by @bhark, as it more easily allows to be compatibel with older librsvg versions (<2.43) as well.

Some test still fail, I think mostly because the `expected` results are made with broken svg 

However, it's quite a task for me to go through this (skill and capacity wise, I have a very un-efficient test setup now), so if anyone want to have a look, more than welcome!

jmckenna:
@johanez thanks.  At this point, the issue isn't fixing the tests in this pull request, the issue is that people in #7430 have confirmed a different fix, than this one.  I really want everyone to agree on one fix...I'll give this a few days more to await everyone's feedback here.  Thanks for pushing this forward @johanez very appreciated!

bhark:
@jmckenna The fix implemented here is the same as i suggested in #7430, and seems functional to me. #7462 just adds the guard, which is already in this PR. 

With regards to the failing tests @johanez the original commit that introduced this issue also regenerated the reference pictures, so they're currently checking the correct rendering against a faulty baseline. I think we just need to regenerate the references (all 6) and call it a day. I'll open a PR on @johanez fork if i get the time later today.

jmckenna:
@johanez @bhark I've updated the expected test results here, and the rendering looks a lot better (see 87144bb2b61ef11295eb3c9c835e85c7252b3533 )

johanez:
Thanks @jmckenna and @bhark . I indeed changed it into using the same fix as @bhark suggested, as it's simpler to keep the older librsvg version guard.

Looks like this is ready to merge and great if you backport it to 8.6.6!

jmckenna:
thanks again @johanez and @bhark !

PR #7591, opened 2026-08-07
Add SLD export support custom vector symbols and hatch fills

geographika:
This pull request allows more MapServer vector styles to be accessed as SLD through WMS `GetStyles` requests. Styles in a Mapfile can then used by both MapServer, and by a client web application.

## Background

SLD can be generated through MapServer URLs such as https://demo.mapserver.org/cgi-bin/wms?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetStyles&layers=country_bounds which in this example returns the SLD of the `country_bounds` layer. This SLD can be read, for example (and for my main use case), by using the [GeoStyler SLD Parser](https://github.com/geostyler/geostyler-sld-parser/) in a browser and applying it to a vector layer in OpenLayers using the [GeoStyler OpenLayers parser](https://github.com/geostyler/geostyler-openlayers-parser/).

MapServer currently supports exporting the 6 WellKnownName names in the core SLD specification - square, circle, triangle, star, cross, x. If these are used in a MapServer layer then the `GetStyles` request returns SLD such as:

```xml
<se:WellKnownName>circle</se:WellKnownName>
```

### Hatches

This pull request adds support for hatch fills which are supported by [GeoServer](https://docs.geoserver.org/main/en/user/styling/sld/extensions/pointsymbols/#shape-symbols), and [GeoStyler](https://github.com/geostyler/geostyler-sld-parser/blob/48f761c812d5d04df44cfa09f3969b5c78778131/src/SldStyleParser.ts#L1373).

A Mapfile containing:

```
SYMBOL
	NAME "hatch"
	TYPE HATCH
END
	
LAYER
	NAME "polygon_snap_to_slash"
	TYPE POLYGON
	FEATURE
		POINTS
			0 0
			0 10
			10 10
			10 0
			0 0
		END
	END
	CLASS
		STYLE
			COLOR 0 0 255
			SYMBOL "hatch"
			ANGLE 30
			WIDTH 0.5
			SIZE 6 # this won't be output to the SLD
		END
	END
END
```

Will now produce the following SLD:

```xml
<se:Mark>
	<se:WellKnownName>shape://slash</se:WellKnownName>
	<se:Stroke>
		<se:SvgParameter name="stroke">#0000ff</se:SvgParameter>
		<se:SvgParameter name="stroke-width">0.5</se:SvgParameter>
	</se:Stroke>
</se:Mark>
```

The code will calculate the closest matching angle, so 80° will display as 90°, and output the following WellKnownNames:

- ~0°	shape://horline
- ~45° shape://slash
- ~90° shape://vertline
- ~135° shape://backslash

### Custom Vector Symbols

This PR also adds support for exporting any generic `SYMBOL` to SLD, by exporting it as an SVG, and then encoding it (to make it smaller in size), using GDAL's `CPLBase64Encode` function (GDAL is already a required MapServer dependency). Embedding the SVG in the SLD file avoids external markers and files, and issues around paths and security.

For example a Mapfile with the following:

```
SYMBOL
	NAME "pentagon"
	TYPE VECTOR
	FILLED TRUE
	POINTS
		0.5 0
		1 0.38
		0.81 1
		0.19 1
		0 0.38
		0.5 0
	END
END


LAYER
	NAME "pentagon"
	TYPE POINT
	FEATURE
		POINTS
			0 0
		END
	END
	CLASS
		STYLE
			SYMBOL "pentagon"
			COLOR "#156082"
			OUTLINECOLOR 255 0 0
			SIZE 16
		END
	END
END
```
	
Will produce a `GetStyles` response of:

```xml
<se:PointSymbolizer>
	<se:Graphic>
		<se:ExternalGraphic>
			<se:OnlineResource xlink:type="simple" xlink:href="data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHdpZHRoPSIxOCIgaGVpZ2h0PSIxOCIgdmlld0JveD0iMCAwIDE4IDE4Ij4KPHBhdGggZD0iTSA5LDEgTCAxNyw3LjA4IEwgMTMuOTYsMTcgTCA0LjA0LDE3IEwgMSw3LjA4IEwgOSwxIFoiIGZpbGw9IiMxNTYwODIiIHN0cm9rZT0iI2ZmMDAwMCIgc3Ryb2tlLXdpZHRoPSIxIi8+Cjwvc3ZnPgo="/>
			<se:Format>image/svg+xml</se:Format>
		</se:ExternalGraphic>
		<se:Size>16</se:Size>
	</se:Graphic>
</se:PointSymbolizer>
```

Ellipse symbols are also handled (by a separate function):

```
SYMBOL
	NAME "oval"
	TYPE ELLIPSE
	FILLED TRUE
	POINTS
		2 1
	END
END
```

Initially, I only added support for symbols which matched some of the GeoServer conventions such as `shape://hexagon`, but a generic SYMBOL to SVG function handles both "named" shapes and any other custom symbols, including 
those with gaps using the `-99 -99` breaks in the `POINTS` block. The older approach can be seen in earlier commits (e,g, https://github.com/MapServer/MapServer/commit/517321653a9c34419c303576be0d61d9e26174c6). 

Code and examples all written by me, with a review of the C++ by Claude.


geographika:
A few more commits after further testing with various symbols and properties.

- The bounding box of a SYMBOL always starts at 0,0, which allows symbols with padding around them. This is now supported (and a test added)
- `ANGLE` is now supported for custom symbols. When testing this raised an existing bug - MapServer ANGLE is anti-clockwise (see https://mapserver.org/mapfile/style.html#mapfile-style-angle) whereas as SLD's `Rotation` is clockwise. This has been fixed for existing symbols. 
- Opacity settings now taken into account for custom vector symbols
- Vector symbol sizes now take into account any outline widths to avoid clipping at the edge of rendered tiles

geographika:
Documentation PR also added. I'll leave these open for a week if anyone has any comments or questions before merging.

PR #7593, opened 2026-08-10
bound WKB point reads to the Z/M byte stride in mappostgis

nvxbug:
## What does this PR do?

AddressSanitizer, reading a Z LineString whose point count clears the old check:

    READ of size 8 at 0x6070000000d4 thread T0
        #0 wkbReadPointP
    0x6070000000d4 is located 0 bytes after 68-byte region

`wkbReadLine` bounds the point count against the bytes left divided by 16, but `wkbReadPointP` reads 24 bytes per point for a Z or M geometry and 32 for ZM, so a Z/M linestring or polygon ring can read past the WKB buffer returned by a PostGIS query. `arcStrokeCircularString` had no such check at all and read its whole odd point count unbounded. Both now derive the per-point stride from the Z/M flags before reading. `wkbReadLine` also clears the `lineObj` on the reject path, since `msAddLineDirectly` takes ownership of `line->point` and the early return previously left it uninitialized.

None.

## AI tool usage

## Tasklist

 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments


jmckenna:
@nvxbug please include a test for this in `/msautotest` (there are existing PostGIS tests there), follow https://mapserver.org/development/tests/autotest.html

nvxbug:
Added tests in msautotest/mspython/test_postgis.py: M and ZM linestrings plus 2D and Z circular strings, which cover the point strides this PR bounds, with the matching tables in create_postgis_test_data.sh/.bat. Verified locally against PostGIS 3.4.

One note on scope: the over-read itself needs WKB whose point count disagrees with its byte length, and a live server won't hand that back since the geometry column goes through ST_AsBinary. So these act as regression coverage that valid M/Z/ZM geometries and stroked arcs still parse correctly with the tighter checks; the malformed-input case is what the ASan repro in the description exercises.

PR #7596, opened 2026-08-16
fix out-of-bounds read in SLD arithmetic operator parsing

nvxbug:
msSLDParseOgcExpression() classifies an OGC element as an arithmetic operator with strstr(ops, psRoot->pszValue), where ops is the literal "Add+Sub-Mul*Div/" and pszValue is the element name from an untrusted SLD document (SLD= / SLD_BODY= on a WMS/WFS request). Since the untrusted name is the strstr needle, any substring match qualifies, and the following *(strstr(...) + 3) assumes the match is one of the three-letter names sitting at a fixed offset. An element like <ogc:v> matches near the end of ops, so the +3 reads one byte past the literal (AddressSanitizer reports a global-buffer-overflow read).

Match the element name against the four operator names exactly before deriving the operator character, so strstr only lands on Add/Sub/Mul/Div and the +3 stays in bounds. This also keeps unrelated names such as <ogc:ub-> from being misread as operators. Doing the check where the operator is resolved covers every SLD path that reaches this parser.

### DIFF
--- a/src/mapogcsld.cpp
+++ b/src/mapogcsld.cpp
@@ -1594,8 +1594,11 @@ int msSLDParseOgcExpression(CPLXMLNode *psRoot, void *psObj, int binding,
       exprBindings[binding].type = MS_EXPRESSION;
       (*nexprbindings)++;
       status = MS_SUCCESS;
-    } else if (strstr(ops, psRoot->pszValue) && psRoot->psChild &&
-               psRoot->psChild->psNext) {
+    } else if (psRoot->psChild && psRoot->psChild->psNext &&
+               (strcmp(psRoot->pszValue, "Add") == 0 ||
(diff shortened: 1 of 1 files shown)


geographika:
Most of the code in this file uses `strcasecmp` (even though the spec might specify case-sensitivity). IMO probably best to match the rest of the parsing code and use `strcasecmp`.

nvxbug:
Makes sense, switched to strcasecmp. That also meant dropping the "Add+Sub-Mul*Div/" lookup string entirely: strstr is case-sensitive so it couldn't resolve the operator char for mixed-case names, and the char is now derived directly from which name matched. Builds clean and Add/Sub/Mul/Div in any case still map to +/-/*//.

PR #7602, opened 2026-08-21
Fix heap out-of-bounds write in interpolation layers (GHSA-59gr-4vvx-5f56)

MarkLee131:
## Summary

`msInterpolationDataset()` (`src/interpolation.c`) sized the `xyz_values` buffer by the output pixel count (`WIDTH*HEIGHT`) but filled it with 3 floats per in-extent sample vertex. The number of vertices is not bounded by the pixel count, so an unauthenticated WMS `GetMap` over an interpolation layer (`IDW`, `KERNELDENSITY`, or `KRIGING`) with a grid smaller than the sample count wrote past the end of the allocation — a heap out-of-bounds write (CWE-787).

This addresses security advisory GHSA-59gr-4vvx-5f56.

## Fix

Grow `xyz_values` on demand as vertices are appended, so every sample is stored inside the allocation. When the output grid is already large enough (the common case) the buffer is never grown, so behaviour is unchanged.

## Test

Adds an msautotest regression test, following https://mapserver.org/development/tests/autotest.html :

- `msautotest/gdal/interpolation_oob.map` renders a 30×30 grid over the existing 1000-point `data/pnts.shp` fixture via a WMS `GetMap` on an `idw` layer.
- `msautotest/gdal/expected/interpolation_oob.png` is the expected output.

Before the fix the test aborts under AddressSanitizer (`heap-buffer-overflow`, `WRITE of size 4` in `msInterpolationDataset`); after the fix it renders a valid image. Verified locally under both the regular test run and `--run_under_asan --strict_mode`, and the existing `idw` / `heat` / `kriging` tests still pass unchanged.


msautotest/gdal/interpolation_oob.map
geographika, line 1: This appears to be almost identical to https://github.com/MapServer/MapServer/blob/main/msautotest/gdal/idw.map. Can't the test just be added to the existing Mapfile?

geographika (changes requested):

msautotest/gdal/interpolation_oob.map
MarkLee131: thanks for reminder. Will amend the commit accordingly.

src/interpolation.c
```diff
@@ -223,6 +224,16 @@ int msInterpolationDataset(mapObj *map, imageObj *image,
           if (x >= 0 && y >= 0 && x < im_width && y < im_height) {
             float *value = values + y * im_width + x;
             (*value) += weight;
+            /* xyz_values stores 3 floats (x, y, accumulated value) for every
+             * in-extent vertex. The number of vertices is not bounded by the
+             * output pixel count, so grow the buffer as needed rather than
+             * writing past its end (GHSA-59gr-4vvx-5f56). */
+            if ((size_t)length + 3 > xyz_capacity) {
+              while ((size_t)length + 3 > xyz_capacity)
```
geographika, line 236: Needs a guard here to stop an infinite loop? `if (xyz_capacity == 0) xyz_capacity = 3;` although image sizes should never be 0 in theory.

geographika (approved):

jmckenna:
thanks @MarkLee131 !

PR #7603, opened 2026-08-22
mapogcfilter: limit filter operation count

sdjasj:
## What does this PR do?

Limits parsed OGC filters to 256 supported operations before constructing the internal `FilterEncodingNode` tree. Logical operators with many sibling operands are currently converted to a right-deep binary tree, so bounding the operation count before conversion keeps the later recursive validation, expression generation, and cleanup passes from exhausting the process stack. WMS uses its existing invalid-filter path to return `InvalidParameterValue`.

 Adds a WMS GetMap regression test to the existing `msautotest/wxs/wms_filter.map` using its `RUN_PARMS` POST mechanism. The filter contains 256 sibling `<PropertyIsEqualTo>` clauses under a single `<Or>` (257 supported operations in total), exceeds the 256-operation limit, and expects the existing blank WMS exception image, confirming that the request is rejected without rendering or crashing.

Follow-up to the security report and the request from @jmckenna to open a regular pull request.

## AI tool usage

 - [x] AI supported my development of this PR. See our [policy about AI tool use](https://mapserver.org/community/ai_tool_policy.html). Use of AI tools, if any, *must* be indicated.

## Tasklist

 - [x] Confirm the 9,500-clause reproducer returns `InvalidParameterValue` with both 8 MB and 1 MB stack limits
 - [ ] Review
 - [ ] Adjust for comments


jmckenna:
@sdjasj please do not modify `run_test.py` and instead modify an existing `.map` (or create a new one) for your tests.

jmckenna:
thanks @sdjasj !

PR #7606, opened 2026-08-24
Avoid `GetMetadata` request memory leaks

geographika:
Same issue as https://github.com/MapServer/MapServer/pull/7539. 
If duplicate parameters are used in the querystring (in this case for a `GetMetadata` request) then there is a memory leak.
The same approach is taken for WCS and WFS (`msWFSSetParam`) - in this case the first parameter value will be used. I may look at using one common function for these if it arises again. 
A test has also been added for this case. 

Fixes oss-fuzz 5362884063068160

```
==250==ERROR: LeakSanitizer: detected memory leaks
	
	Direct leak of 1 byte(s) in 1 object(s) allocated from:
	    #0 0x57a515db0dc4 in ___interceptor_malloc /src/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:67:3
	    #1 0x57a515e0f369 in msStrdup MapServer/src/mapstring.cpp:2324:35
	    #2 0x57a515f6abfc in msMetadataParseRequest MapServer/src/mapmetadata.c:970:45
	    #3 0x57a515f6abfc in msMetadataDispatch MapServer/src/mapmetadata.c:851:7
	    #4 0x57a515dfa8b3 in msOWSDispatch MapServer/src/mapows.cpp:341:16
```


geographika:
Removing backport, as this could change functionality - currently if there are duplicate parameters (which is technically invalid, and therefore unknown behaviour) then the last parameter is used. This PR changes it to the first to match WFS, but probably best to include as part of a new release / version.

PR #7608, opened 2026-08-24
mapquery: reject out-of-range layer index in loadQueryResults

nvxbug:
## What does this PR do?

`loadQueryResults()` reads back a saved query file (the `queryfile` CGI parameter, through `msLoadQuery`) and checks each layer index it reads from the file with `j > map->numlayers`. That bound is off by one: `j == map->numlayers` passes, and `GET_LAYER(map, j)` then indexes `map->layers` one slot past the last valid layer. `map->layers` is grown in `MS_LAYER_ALLOCSIZE` (64) chunks, so when the number of layers fills a chunk the array is packed and `map->layers[numlayers]` reads a pointer past the allocation, which is then dereferenced to store `resultcache`.

AddressSanitizer, crafted `.qy` file with the layer index set to `numlayers` over a full 64-layer chunk:

    ==ERROR: AddressSanitizer: heap-buffer-overflow ... READ of size 8
        #0 msLoadQuery mapquery.cpp:613
      0x... is located 0 bytes after 512-byte region [0x...100,0x...300)
      allocated by thread T0 here:
        #1 msGrowMapLayers mapfile.c:6446

The comparison should be `>=`, which is what every other layer-index check in this file already uses. An msautotest case in `msautotest/query` loads a crafted `.qy` file whose layer index equals `numlayers` through `mode=nquery&queryfile=` and expects the "Invalid layer index loaded from query file" error output.

None.

## AI tool usage

## Tasklist

 - [x] Add test case (msautotest in `msautotest/query`)
 - [ ] Review
 - [ ] Adjust for comments


jmckenna:
@nvxbug please include a test for this in /msautotest (use an existing `map` or create one), follow https://mapserver.org/development/tests/autotest.html

nvxbug:
Done. Added an msautotest case in msautotest/query: a new queryfile.map (two inline layers) plus a crafted data/queryfile_index_oob.qy whose layer index equals numlayers, loaded through mode=nquery&queryfile=. Without the fix mapserv segfaults on it; with the fix the run produces the expected "Invalid layer index loaded from query file" output (expected/queryfile_test001.txt). Verified both ways locally with the pytest harness, and the query suite still passes. Kept the unit test as well.

geographika:
With the msautotest there is no longer a need to also have `‎tests/unit/test.cpp`, which is just more code to try and understand and maintain.

nvxbug:
Fair enough, the msautotest covers the same path. Dropped the unit test, so the PR is now just the one-character fix plus the msautotest case.

jmckenna:
@thanks for the quick change @nvxbug

PR #7612, opened 2026-08-25
mapobject.c: clear cached reprojectors in msRemoveLayer() (fixes #7611)

### LINKED ISSUE #7611: msRemoveLayer() leaves stale cached reprojectors on the removed layer
`msRemoveLayer()` detaches a layer but does not clear `layer->reprojectorLayerToMap` / `layer->reprojectorMapToLayer`. Those cached reprojectors hold a raw pointer to the old map's projection in their `out` field, so once that map is freed the layer carries a dangling pointer. Moving a layer to another map with `removeLayer()` / `insertLayer()` and then freeing the original map gives a use-after-free the next time the layer is drawn or reprojected.

`freeLayer()` already destroys both cached reprojectors (`mapfile.c:4106-4107`) and `initLayer()` initialises them to NULL (`mapfile.c:3960`). `msRemoveLayer()` (`mapobject.c:597`) is the only one of the three that does not touch them, which is what makes this look like an oversight rather than a deliberate choice.

Reachable from MapScript, where `removeLayer()` and `insertLayer()` are public API, so an application doing nothing unusual can hit it.

**Steps to reproduce**

`poc.map`:

```
MAP
  NAME "poc"
  SIZE 256 256
  EXTENT -20037508 -20037508 20037508 20037508
  UNITS METERS
  PROJECTION "init=epsg:3857" END
  LAYER
    NAME "pts"
    TYPE POINT
    STATUS ON
    FEATURE POINTS 0 0 END END
    CLASS STYLE SIZE 4 COLOR 255 0 0 END END
  END
END
```

`repro.c`, the C equivalent of MapScript `removeLayer()` / `insertLayer()`:

```c
#include "mapserver.h"
#include <stdio.h>

int main(void) {
  msSetup();

  mapObj *A = msLoadMap("poc.map", NULL, NULL);
  layerObj *L = GET_LAYER(A, 0);

  /* Populate the cache. reprojector->out is &A->projection. */
  msLayerGetReprojectorToMap(L, A);

  /* Move the layer to a second map. */
  msRemoveLayer(A, 0);
  printf("after msRemoveLayer, L->reprojectorLayerToMap = %p\n",
         (void *)L->reprojectorLayerToMap);

  mapObj *B = msLoadMap("poc.map", NULL, NULL);
  msRemoveLayer(B, 0);
  msInsertLayer(B, L, -1);

  msFreeMap(A);                     /* &A->projection is now freed */

  msLayerGetReprojectorToMap(L, B); /* revalidation reads the freed projection */

  msCleanup();
  return 0;
}
```

Build against an ASan build of libmapserver and run it from the directory holding `poc.map`:

```
clang -fsanitize=address -g -O1 -I<mapserver>/src repro.c -lmapserver -o repro
ASAN_OPTIONS=detect_leaks=0 ./repro
```

Result on 8.6.5 (`d25c319`):

```
after msRemoveLayer, L->reprojectorLayerToMap = 0x50c000000040

==11==ERROR: AddressSanitizer: heap-use-after-free on address 0x525000001620
READ of size 2 at 0x525000001620 thread T0
    #0 msProjectIsReprojectorStillValid  src/mapproject.c:1246:25
    #1 msLayerGetReprojectorToMap        src/maplayer.c:2098:8
    #2 main                              repro.c

freed by thread T0 here:
    #1 main repro.c            (msFreeMap(A))

previously allocated by thread T0 here:
    #1 msLoadMap src/mapfile.c:7042:19
```

The read is `reprojector->out->generation_number` at `mapproject.c:1246`, where `out` still points into the freed `mapObj`.

**Version**

Reproduced on rel-8-6-5 (`d25c31936f3f67911a56560e95ed2c8a7cb8ab4f`). `msRemoveLayer()` on current main has the same gap.

**Suggested fix**

Destroy and NULL both cached reprojectors in `msRemoveLayer()`, mirroring what `freeLayer()` already does. Happy to send a PR.

MarkLee131:
## What does this PR do?

`msRemoveLayer()` detached a layer without clearing `layer->reprojectorLayerToMap` / `reprojectorMapToLayer`. Those cached reprojectors hold a raw pointer to the map the layer was removed from, which has two consequences:

- A layer moved to another map keeps reprojecting to the **old** map's projection, so it is drawn in the wrong place or not at all.
- If the old map is freed while the layer lives on, the next `msLayerGetReprojectorToMap()` reads `reprojector->out->generation_number` from freed memory (heap-use-after-free under ASan).

`freeLayer()` already destroys both cached reprojectors and `initLayer()` initialises them to NULL, so this brings `msRemoveLayer()` in line with the other two places that manage these fields.

The msautotest case moves a layer from a Web Mercator map to a UTM map and checks that it still renders. Without the fix the polygon is projected to the first map's coordinates, falls outside the second map's extent, and nothing is drawn.

Fixes #7611

## Tasklist

 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/msautotest/mspython/test_bug_check.py
+++ b/msautotest/mspython/test_bug_check.py
@@ -303,3 +303,65 @@ def test_bug_6896():
 
     os.unlink(ref_filename)
     os.unlink(test_filename)
+
+
+###############################################################################
+# Verify that a layer moved to another map no longer carries the cached
+# reprojector of the map it was removed from.
+#
+# https://github.com/MapServer/MapServer/issues/7611
+
+
+def test_bug_7611():
+    def add_polygon_layer(map):
+        layer = mapscript.layerObj(map)
+        layer.name = "poly"
+        layer.type = mapscript.MS_LAYER_POLYGON
+        layer.status = mapscript.MS_ON
+        layer.setProjection("init=epsg:4326")
+
+        shape = mapscript.shapeObj(mapscript.MS_SHAPE_POLYGON)
+        ring = mapscript.lineObj()
+        for x, y in [
+            (-117.4, 42.8),
+            (-116.6, 42.8),
+            (-116.6, 43.2),
+            (-117.4, 43.2),
+            (-117.4, 42.8),
+        ]:
+            ring.add(mapscript.pointObj(x, y))
+        shape.add(ring)
+        layer.addFeature(shape)
+
+        style = mapscript.styleObj(mapscript.classObj(layer))
+        style.color = mapscript.colorObj(255, 0, 0)
+        return layer
+
+    def renders(map, layer):
+        drawn = map.draw().getBytes()
+        layer.status = mapscript.MS_OFF
+        blank = map.draw().getBytes()
+        layer.status = mapscript.MS_ON
+        return drawn != blank
+
+    # Map A is Web Mercator, the layer is geographic. Drawing caches a
+    # reprojector on the layer whose "out" is map A's projection.
+    map_a = mapscript.mapObj()
+    map_a.setSize(100, 100)
+    map_a.setProjection("init=epsg:3857")
+    map_a.setExtent(-13100000, 5250000, -12950000, 5380000)
+    layer = add_polygon_layer(map_a)
+    assert renders(map_a, layer)
+
+    # Move the layer to a map that uses a different projection.
+    map_a.removeLayer(0)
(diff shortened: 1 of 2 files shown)


geographika (approved):

jmckenna:
thanks @MarkLee131 !

PR #7615, opened 2026-08-26
mapproject.c: check msProjectCreateReprojector() result in msProjectRect() (fixes #7613)

### LINKED ISSUE #7613: msProjectRect() does not check msProjectCreateReprojector() for NULL
In the "polar stereographic to another projected system" branch of `msProjectRect()`, the result of `msProjectCreateReprojector()` is passed straight to `msProjectRectAsPolygon()`:

https://github.com/MapServer/MapServer/blob/d25c31936f3f67911a56560e95ed2c8a7cb8ab4f/src/mapproject.c#L2051-L2072

`msProjectCreateReprojector()` returns NULL when `createNormalizedPJ()` cannot build a transformation, and `msProjectRectAsPolygon()` dereferences `reprojector->out` on its first line:

https://github.com/MapServer/MapServer/blob/d25c31936f3f67911a56560e95ed2c8a7cb8ab4f/src/mapproject.c#L1807

The other six calls to `msProjectCreateReprojector()` in this file all check the result (890, 1568, 1661 return `MS_FAILURE`; 2009 and 2044 guard with `&&`; 2144 uses a conditional). Only this one does not, which is what makes it look like an oversight.

There is a second problem in the same block: `reprojector` is created inside the `for (sign ...)` loop but only destroyed on the two `return` paths. If neither rect condition matches, the second iteration overwrites the pointer and the function exits at the bottom destroying only `reprojectorToLongLat`, so the reprojector leaks.

**Steps to reproduce**

```python
import mapscript

# North polar stereographic: (0,0) in this CRS is the pole, which selects the
# branch at mapproject.c:2040.
pin = mapscript.projectionObj(
    "+proj=stere +lat_0=90 +lat_ts=70 +lon_0=-45 +k=1 +x_0=0 +y_0=0 "
    "+datum=WGS84 +units=m +no_defs")

# Not a CRS, so proj_create_crs_to_crs_from_pj() fails and
# msProjectCreateReprojector() returns NULL.
pout = mapscript.projectionObj("+proj=noop")

rect = mapscript.rectObj(-1000000, -1000000, 1000000, 1000000)  # encloses (0,0)
rect.project(pin, pout)
```

`rectObj.project()` maps to `msProjectRect()`. On an ASan build:

```
AddressSanitizer: SEGV on unknown address 0x0000000000000008
    #0 msProjectRectAsPolygon  src/mapproject.c:1807:20
    #1 msProjectRect
    #2 _mapscript.so
```

`0x8` is `NULL` plus the offset of `reprojectionObj.out`. Without a sanitizer it is a plain segfault.

**Version**

Reproduced on rel-8-6-5 (`d25c31936f3f67911a56560e95ed2c8a7cb8ab4f`); current main has the same code.

`+proj=noop` is just the shortest way to make `msProjectCreateReprojector()` fail on demand. In a normal deployment the NULL comes from PROJ being unable to build the transformation, for example when grid files are missing.

**Suggested fix**

Check the result before use, and destroy the reprojector when neither branch returns, matching what the rest of the file does. Happy to send a PR.

MarkLee131:
## What does this PR do?

`msProjectRect()` passed the result of `msProjectCreateReprojector()` to `msProjectRectAsPolygon()` without checking it. `msProjectCreateReprojector()` returns NULL when a transformation cannot be built, and `msProjectRectAsPolygon()` dereferences `reprojector->out` immediately, so the process segfaults.

The six other calls to `msProjectCreateReprojector()` in this file all check the result; this one was the exception.

The same block also leaked the reprojector when neither rect condition matched, because it is created inside the `for (sign ...)` loop but destroyed only on the two `return` paths. This adds the missing destroy.

Fixes #7613

## AI tool usage

## Tasklist

 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/msautotest/mspython/test_bug_check.py
+++ b/msautotest/mspython/test_bug_check.py
@@ -365,3 +365,25 @@ def test_bug_7611():
     # projected to map A's Web Mercator coordinates, lands outside map B's
     # extent, and nothing is drawn.
     assert renders(map_b, layer)
+
+
+###############################################################################
+# msProjectRect() used the result of msProjectCreateReprojector() without
+# checking it for NULL, which segfaults when a transformation cannot be built.
+# The polar branch is selected because (0,0) in this CRS is the pole and the
+# rectangle encloses it.
+
+
+def test_msprojectrect_null_reprojector():
+
+    polar = mapscript.projectionObj(
+        "+proj=stere +lat_0=90 +lat_ts=70 +lon_0=-45 +k=1 +x_0=0 +y_0=0 "
+        "+datum=WGS84 +units=m +no_defs"
+    )
+    # Not a CRS, so no transformation can be built from it.
+    not_a_crs = mapscript.projectionObj("+proj=noop")
+
+    rect = mapscript.rectObj(-1000000, -1000000, 1000000, 1000000)
+
+    # Only has to return rather than crash.
+    rect.project(polar, not_a_crs)
--- a/src/mapproject.c
+++ b/src/mapproject.c
@@ -2052,6 +2052,8 @@ int msProjectRect(projectionObj *in, projectionObj *out, rectObj *rect) {
           fabs(p.y - 90) < 1e-8) {
         in->is_polar = 1;
         reprojector = msProjectCreateReprojector(in, out);
+        if (reprojector == NULL)
+          break;
         /* Is the pole in the rectangle ? */
         if (0 >= rect->minx && 0 >= rect->miny && 0 <= rect->maxx &&
             0 <= rect->maxy) {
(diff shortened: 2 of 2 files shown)


geographika (approved):

geographika:
@jmckenna - not sure why there is an additional merge added to this PR. It removes the following test: https://github.com/MapServer/MapServer/blob/42263072bf6534fc1fdcc95f4168db334c3053f8/msautotest/mspython/test_bug_check.py#L315 - is this intentional?

jmckenna:
@geographika hi, what happened was that this pull request was showing a "Conflict", and I had tried to remove the conflict.  I apologize if I made a mistake.

geographika:
@jmckenna ok I think the branch should be ok now. Will let CI run.

jmckenna:
thanks @geographika

PR #7616, opened 2026-08-26
mappool.c: only raise connectionMax after a successful realloc (fixes #7614)

### LINKED ISSUE #7614: msConnPoolRegister() writes past the connection array after a failed realloc
`msConnPoolRegister()` raises `connectionMax` before the `realloc()` that is supposed to deliver the new capacity, and does not put it back when the allocation fails:

https://github.com/MapServer/MapServer/blob/d25c31936f3f67911a56560e95ed2c8a7cb8ab4f/src/mappool.c#L205-L218

After one failed grow, `connectionMax` is permanently 10 ahead of the capacity `connections` actually has. Because the guard is an equality test, it never fires again, so the next ten registrations each write a whole `connectionObj` past the end of the array:

https://github.com/MapServer/MapServer/blob/d25c31936f3f67911a56560e95ed2c8a7cb8ab4f/src/mappool.c#L223-L227

`connectionObj` is 64 bytes on x86-64 and contains `conn->connection` (a `char *`), `conn->conn_handle` and the `conn->close` function pointer, so this is a heap buffer overflow rather than a benign overwrite.

This is a leftover from f4286d624, which fixed the realloc misuse here. Before that commit a failed realloc left `connections` NULL and the next registration crashed on a near-NULL write; afterwards `connections` correctly keeps pointing at the old buffer, which turns the same `connectionMax` desync into a silent overflow.

**Steps to reproduce**

`msConnPoolRegister()` is not exposed to MapScript and its callers are the database drivers, so this is easiest to show with a small C program linked against libmapserver. AddressSanitizer is used to make the pool realloc fail on demand: the array grows ten `connectionObj` at a time, so it eventually exceeds `max_allocation_size_mb` while every other allocation still succeeds.

```c
#include "mapserver.h"
#include <stdint.h>
#include <stdio.h>

int main(void) {
  layerObj layer;
  memset(&layer, 0, sizeof(layer));
  layer.name = (char *)"poc";
  layer.connection = (char *)"poc-connection-string";
  layer.connectiontype = MS_POSTGIS;

  for (long i = 0; i < 2000000; i++) {
    errorObj *err;
    msConnPoolRegister(&layer, (void *)(uintptr_t)(i + 1), NULL);
    err = msGetErrorObj();
    if (err && err->code == MS_MEMERR) {
      printf("pool realloc failed at registration %ld\n", i);
      msResetErrorList();
      msConnPoolRegister(&layer, (void *)(uintptr_t)0xdeadbeef, NULL);
      printf("no overflow reported\n");
      return 1;
    }
  }
  return 2;
}
```

```
ASAN_OPTIONS=allocator_may_return_null=1:max_allocation_size_mb=1 ./poc
```

```
==12==WARNING: AddressSanitizer failed to allocate 0x100180 bytes
pool realloc failed at registration 16380

==12==ERROR: AddressSanitizer: heap-buffer-overflow
WRITE of size 4 at 0x76d7b2294700
0x76d7b2294700 is located 0 bytes after 1048320-byte region

    #0 msConnPoolRegister  src/mappool.c:227:24
  allocated by:
    #0 msConnPoolRegister  src/mappool.c:210:54
```

16380 entries x 64 bytes is the 1048320-byte region; the failed request of 16390 entries is 1048960 bytes, which is the `0x100180` in the warning.

**Version**

Reproduced on rel-8-6-5 (`d25c31936f3f67911a56560e95ed2c8a7cb8ab4f`); current main has the same code.

**Suggested fix**

Only raise `connectionMax` once the realloc has succeeded, and make the guard `>=` so that a count which has somehow passed the capacity still takes the growth path. Happy to send a PR.

MarkLee131:
## What does this PR do?

This moves the increment after the successful realloc and changes the guard to `>=`.

It completes f4286d624: that commit stopped the failed realloc from leaking and from nulling `connections`, but left the `connectionMax` desync in place, which turned the old near-NULL crash into a silent heap overflow.

Fixes #7614

## Tasklist

 - [ ] Add test case(s) in `/msautotest` (follow steps in [Regression Testing](https://mapserver.org/development/tests/autotest.html))
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

I could not find a way to write this one as an msautotest case. Triggering it needs the pool array's `realloc()` to fail, and the suite has no way to inject an allocation failure; `msConnPoolRegister()` is also not exposed to MapScript, and its callers are the database drivers, so reaching it from the test suite would need a live backend as well. The C reproducer in the issue covers it instead. Glad to add a test if you see an angle I missed.

### DIFF
--- a/src/mappool.c
+++ b/src/mappool.c
@@ -205,16 +205,20 @@ void msConnPoolRegister(layerObj *layer, void *conn_handle,
   /* -------------------------------------------------------------------- */
   msAcquireLock(TLOCK_POOL);
 
-  if (connectionCount == connectionMax) {
-    connectionMax += 10;
+  if (connectionCount >= connectionMax) {
+    const int newConnectionMax = connectionMax + 10;
     connectionObj *newConnections = (connectionObj *)realloc(
-        connections, sizeof(connectionObj) * connectionMax);
+        connections, sizeof(connectionObj) * newConnectionMax);
     if (newConnections == NULL) {
+      /* Leave connectionMax matching the capacity we actually have, so a
+         later registration still takes this growth path instead of writing
+         past the end of the existing array. */
       msSetError(MS_MEMERR, NULL, "msConnPoolRegister()");
       msReleaseLock(TLOCK_POOL);
       return;
     }
     connections = newConnections;
+    connectionMax = newConnectionMax;
   }
 
   /* -------------------------------------------------------------------- */


jmckenna (approved):

jmckenna:
thanks @MarkLee131 !

PR #7624, opened 2026-08-28
Fix integer overflows in WCS SCALEEXTENT parsing

geographika:
Fix for mapserver:owsfuzzer: Integer-overflow in msWCSParseRequest20 553179917

- Handle integer overflows for `SCALEEXTENT` parameters
- Also fixes for the same issue when using POST/XML
- Update `msStringParseInteger` function. This previously had a `MS_DONE` return parameter, but it was never used so I have removed it. The docstring comment also referred to setting the return value to `MS_WCS20_UNBOUNDED` but this was never implemented. The function now checks the value is in a valid integer range. There are probably GDAL CPL functions we could use for much of these value conversions. This is a local function so the impact is restricted to `mapwcs20.cpp`.
- Fixed an incorrect copied and pasted error message. 
- Removed `.` from the integer error messages as it made the values look like messed-up floats 
- Added tests for the new error paths

### DIFF
--- a/msautotest/wxs/wcs_simple.map
+++ b/msautotest/wxs/wcs_simple.map
@@ -261,6 +261,10 @@
 # ScalingError
 # RUN_PARMS: wcs_20_exception_scaling.xml [MAPSERV] QUERY_STRING="map=[MAPFILE]&SERVICE=WCS&VERSION=2.0.1&REQUEST=GetCoverage&COVERAGEID=grey&FORMAT=image/tiff&RESOLUTION=x(5)&SIZE=x(10)" > [RESULT_DEMIME]
 #
+# ScaleExtent exceptions
+# RUN_PARMS: wcs_20_exception_scaleextent.xml [MAPSERV] QUERY_STRING="map=[MAPFILE]&SERVICE=WCS&VERSION=2.0.1&REQUEST=GetCoverage&COVERAGEID=grey&SCALEEXTENT=x(6666611106:366666666)" > [RESULT_DEMIME]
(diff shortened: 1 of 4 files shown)


jmckenna:
thanks @geographika !

msautotest/wxs/wcs_simple.map
```diff
@@ -261,6 +261,10 @@
 # ScalingError
 # RUN_PARMS: wcs_20_exception_scaling.xml [MAPSERV] QUERY_STRING="map=[MAPFILE]&SERVICE=WCS&VERSION=2.0.1&REQUEST=GetCoverage&COVERAGEID=grey&FORMAT=image/tiff&RESOLUTION=x(5)&SIZE=x(10)" > [RESULT_DEMIME]
 #
+# ScaleExtent exceptions
+# RUN_PARMS: wcs_20_exception_scaleextent.xml [MAPSERV] QUERY_STRING="map=[MAPFILE]&SERVICE=WCS&VERSION=2.0.1&REQUEST=GetCoverage&COVERAGEID=grey&SCALEEXTENT=x(6666611106:366666666)" > [RESULT_DEMIME]
```
jratike80, line 265: There are two errors in the SCALEEXTENT:
- 6666611106 is invalid
- min is bigger than max

Maybe it would be better to test with `SCALEEXTENT=x(6666611106:6666611116)` even it has two errors as well because both min and max are invalid.

geographika: In that case the code still errors out at the min value. I guess a test for a valid min and invalid max could be added to test that error path too.

PR #7625, opened 2026-08-28
maperror: size the error image within MAXSIZE (GHSA-qcjf-q672-q63w)

MarkLee131:
## What does this PR do?

`msWMSLoadGetMapParams()` stores `WIDTH`/`HEIGHT` into `map->width`/`map->height` as it parses them, but the `MAXSIZE` check runs about 360 lines later, with ten `return msWMSException(...)` sites in between. An in-image exception raised at any of them renders the error image at the unvalidated dimensions, so `WIDTH=100000&HEIGHT=100000` asks for a 40 GB buffer. No valid layer name and no knowledge of the mapfile is needed. When the allocation is refused rather than granted, the worker segfaults instead, because `msWriteErrorImage()` does not check `msImageCreate()` for NULL.

The clamp at `mapwms.cpp:1779` already resets to 400x300 "in case errors INIMAGE are used", but only for the exception it raises itself.

Fixed in `msWriteErrorImage()` rather than in the parser: the map dimensions are used only when within `map->maxsize` (the same predicate as `mapdraw.c:330`). That covers all ten paths and the non-WMS callers, and does not change which error any existing request reports. `msImageCreate()` is now NULL-checked.

Unpatched, the added test returns a 5000x5000 error image and fails; patched, 400x300. I ran the `wxs`, `misc`, `config`, `renderers`, `query`, `sld`, `gdal` and `api` suites against both builds and the new test is the only difference.

Fixes https://github.com/MapServer/MapServer/security/advisories/GHSA-qcjf-q672-q63w

## Tasklist

 - [x] Add documentation — n/a, no user-visible behaviour change
 - [ ] Review
 - [ ] Adjust for comments


geographika:
@MarkLee131 - the new null check for the image prevents a crash:

```
  if (img == NULL) {
    /* There is nothing to draw the message on, and the error that brought us
       here is already on the error list. */
    if (format->refcount == 0)
      msFreeOutputFormat(format);
    msFree(errormsg);
    return;
  }
```

But it will return without running the `isreported` code below or setting headers. I'm not sure what a server would return in this case?

```
  /* the errors are reported */
  while (error && error->code != MS_NOERR) {
    error->isreported = MS_TRUE;
    error = error->next;
  }
```

MarkLee131:
That's deliberate: skipping the `isreported` loop is what makes the fallback work.

`msCGIWriteError()` (`mapserv.c:336`) returns early when the error is already marked reported (`mapservutil.c:75`), so setting `isreported` here is what would leave the client with an empty response. Leaving it unset lets that function emit its own error page. No headers have been sent at that point either, since `msWriteErrorImage()` only sends them once `msImageCreate()` has succeeded.

Tested with `MAXSIZE 200000` in the mapfile so the new clamp passes, then requesting 100000x100000 with ASan refusing the allocation. Before, `SEGV on unknown address 0x000000000028`. After:

```
Status: 400 Bad Request
Content-Type: text/html
...
msImageCreate(): Memory allocation error. Unable to create new image object.
msWMSLoadGetMapParams(): WMS server error. Invalid layer(s) given in the LAYERS parameter. ...
```

So the client gets the usual MapServer error page and the worker survives. I've pushed a comment saying so.

geographika:
@MarkLee131 - I've been testting this patch locally, thanks for this. I did manage to get a divide by 0 at `nLines = (int)ceil((double)nTextLength / (double)nMaxCharsPerLine);` causing a result of `inf` with a request where HEIGHT and WIDTH were 10 - `&WIDTH=10&HEIGHT=10&LAYERS=no_such_layer&EXCEPTIONS=INIMAGE`. 

This would have been an issue prior to this PR, but it would be good to add a guard against this too.

Something like:

```
    if (map->width >= (nMargin * 2) + charWidth && map->width <= map->maxsize &&
       map->height > 0 && map->height <= map->maxsize) {

```

Currently the width has to be greater than 15 to avoid the issue.

jmckenna:
@MarkLee131 I'd like to include this in the MapServer 8.6.6 release.  Can you address the previous question, as soon as possible? thanks.

MarkLee131:
Sry for the delayed reply. Have pushed a fix using your condition: 
It's a bit nastier than the inf suggests: (int)inf is UB, and on x86-64 it happens to give INT_MIN, so the loops get skipped and you just see the bad line count. On aarch64 the conversion saturates to INT_MAX instead — malloc(INT_MAX * 8) and then a NULL deref on the next line.
I didn't add a separate guard on nMaxCharsPerLine: with this in place width is either the 400 default or >= 15, so the divisor is never below 1. 

Checked widths 1 to 5000 under -fsanitize=float-cast-overflow, no reports. Also added a WIDTH=10 case to wms_simple.map.
 @jmckenna should be good for 8.6.6.

jmckenna:
thanks @MarkLee131 !

PR #7630, opened 2026-08-29
mapkmlrenderer: bound coordinate and extent formatting with snprintf

nvxbug:
## What does this PR do?

KmlRenderer::addCoordsNode formats each vertex into a fixed 128-byte lineBuf with an unbounded sprintf. KML output runs with MS_TRANSFORM_NONE, so raw data-source coordinates and Z values reach the renderer without clipping, and the third value comes straight from atof() on a feature attribute when kml_elevation_attribute is set. A coordinate or elevation whose magnitude does not fit the buffer (a shapefile vertex or attribute on the order of 1e300) formats to a few hundred bytes and overruns the stack.

Switch that write, and the two sibling sprintf sites in the same file that format request- or attribute-controlled doubles into small buffers (the LatLonBox extent in createGroundOverlayNode, the line width in lookupPlacemarkStyle), to snprintf bounded by sizeof. Keeping the bound at the formatting site covers every caller of the renderer without reaching into the geometry paths.

None; found while auditing the KML output path.

## AI tool usage

## Tasklist

 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/msautotest/wxs/wms_kml_bigcoord.map
+++ b/msautotest/wxs/wms_kml_bigcoord.map
@@ -1,0 +1,57 @@
+#
+# Test KML output with an out-of-range coordinate.
+#
+# A vertex whose value does not fit the fixed coordinate buffer used by the
+# KML renderer must not overflow it. See addCoordsNode() in mapkmlrenderer.cpp.
+#
+# REQUIRES: INPUT=GDAL OUTPUT=KML SUPPORTS=WMS
+
+# RUN_PARMS: wms_kml_bigcoord.kml [MAPSERV] QUERY_STRING="map=[MAPFILE]&SERVICE=WMS&VERSION=1.1.0&REQUEST=GetMap&SRS=EPSG:4326&BBOX=-1e307,-1e307,1e307,1e307&FORMAT=kml&WIDTH=100&HEIGHT=100&STYLES=&LAYERS=bigline" > [RESULT_DEMIME]
+
+MAP
+  NAME KML_BIGCOORD
+  STATUS ON
+  SIZE 100 100
+  EXTENT -1e307 -1e307 1e307 1e307
+  UNITS DD
+  IMAGECOLOR 255 255 255
+
+  PROJECTION
+    "init=epsg:4326"
+  END
+
+  WEB
+    IMAGEPATH "/tmp/ms_tmp/"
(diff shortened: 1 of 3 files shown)


msautotest/wxs/wms_kml_bigcoord.map
```diff
@@ -0,0 +1,57 @@
...
+# REQUIRES: INPUT=GDAL OUTPUT=KML SUPPORTS=WMS
+
+# RUN_PARMS: wms_kml_bigcoord.kml [MAPSERV] QUERY_STRING="map=[MAPFILE]&SERVICE=WMS&VERSION=1.1.0&REQUEST=GetMap&SRS=EPSG:4326&BBOX=-1e307,-1e307,1e307,1e307&FORMAT=kml&WIDTH=100&HEIGHT=100&STYLES=&LAYERS=bigline" > [RESULT_DEMIME]
+
+MAP
+  NAME KML_BIGCOORD
+  STATUS ON
+  SIZE 100 100
+  EXTENT -1e307 -1e307 1e307 1e307
+  UNITS DD
```
jratike80, line 16: If projection is EPSG:4326, perhaps extent like this should yield also an error,

nvxbug: Arguably yes, but nothing validates the extent or BBOX against the CRS today, and KML output runs with MS_TRANSFORM_NONE so the raw values reach the renderer either way. Even with such a check, the elevation path formats a value taken straight from a feature attribute (kml_elevation_attribute), which extent validation wouldn't catch. So the formatting needs a bound regardless; rejecting out-of-range extents for geographic CRS sounds reasonable but feels like a separate change.

jmckenna: I agree @nvxbug, that could be a separate pull request.

jmckenna:
@nvxbug I assume that this pull request can be backported to `branch-8-6`, if not please give me a yell here.

jmckenna:
thanks again @nvxbug !

nvxbug:
Yes, backporting to branch-8-6 should be fine, the change is self-contained to mapkmlrenderer.cpp plus the test. Thanks for merging.

jmckenna:
Great, thanks @nvxbug

PR #7636, opened 2026-09-01
OSS-Fuzz: Adding new fuzzer targets sld

arthurscchan:
This PR adds a new fuzzer targets sld, together with preparation of seeds for the new fuzzer.


geographika:
@arthurscchan - what is the reasoning behing adding a new `sldfuzzer.c` rather than just building on #7525, and updating the Mapfile to allow OWS requests with SLD?

PR #7642, opened 2026-09-04
mapcontext: fix out-of-bounds read on short BoundingBox SRS

nvxbug:
## What does this PR do?

`msLoadMapContextGeneral()` reads `General.BoundingBox.SRS` from a Web Map Context document, and after ruling out the `AUTO:` prefix it builds the projection with `sprintf(pszProj, "init=epsg:%s", pszValue + 5)`, assuming the value starts with a 5-character `EPSG:` prefix. The context document is untrusted (the CGI `context=` parameter loads a local or remote WMC, e.g. through `CGI_CONTEXT_URL`), so an SRS shorter than five characters such as `SRS="X"` makes `pszValue + 5` point past the end of the value and the copy reads out of bounds.

The strip now happens only when the value actually begins with `EPSG:`, mirroring the `AUTO:` branch right above it; anything else is passed through unchanged so an invalid SRS is rejected by the projection layer instead of walking off the buffer. Under AddressSanitizer a GetCapabilities request with a short-SRS context aborts with a heap-buffer-overflow at `mapcontext.c:642` before the change and runs clean after it.

None; found while reading the context loader.

## AI tool usage

## Tasklist

 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments

### DIFF
--- a/msautotest/wxs/expected/ows_context_bad_srs_caps.xml
+++ b/msautotest/wxs/expected/ows_context_bad_srs_caps.xml
@@ -1,0 +1,128 @@
+Content-Type: application/vnd.ogc.wms_xml; charset=UTF-8
+
+<?xml version='1.0' encoding="UTF-8" standalone="no" ?>
+<!DOCTYPE WMT_MS_Capabilities SYSTEM "http://schemas.opengis.net/wms/1.1.1/WMS_MS_Capabilities.dtd"
+ [
+ <!ELEMENT VendorSpecificCapabilities EMPTY>
+ ]>  <!-- end of DOCTYPE declaration -->
+
+<WMT_MS_Capabilities version="1.1.1">
+<Service>
+  <Name>OGC:WMS</Name>
+<!-- WARNING: Mandatory metadata "wms_title" or "ows_title" was missing in this context. -->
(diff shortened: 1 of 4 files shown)


jratike80:
All valid CRSs do not begin with `EPSG:` or `AUTO:`, for example `CRS:84`, that is kind of default in WMS 1.3.0.

But I have never before seen GetCapabilities request like
`&CONTEXT=ows_context_with_layer_list.xml&SERVICE=WMS&VERSION=1.1.1&REQUEST=GetCapabilities"`

There is indeed a standard about Web Map Context https://www.ogc.org/standards/wmc, but is anybody using it, or does any other WMS server support it?

By the documentation https://mapserver.org/ogc/mapcontext.html
>GetContext is disabled by default because it could be considered a security issue: it could publicly expose the URLs of WMS layers used (cascaded) by a mapfile.

I suggest an alternative fix: remove the support for Web Map Context totally from next MapServer versions.

jmckenna:
This change seems fair to me.

(removing mapcontext support is one of the topics discussed for MapServer 9.0 indeed)

jmckenna:
thanks @nvxbug !

PR #7645, opened 2026-09-07
flatgeobuf: bound geometry ring ends to the xy coordinate count

nvxbug:
## What does this PR do?

`GeometryReader` in `src/flatgeobuf/geometryreader.cpp` builds a `shapeObj` from an `.fgb` feature's coordinates. For a polygon the per-ring end indices come from the geometry's `ends()` array in the file, and `readLineObj()` used them to walk `m_xy` (and the parallel `z`/`m` arrays) without checking them against the number of coordinate pairs actually present in `xy()`. A ring end index past the coordinate count reads past the buffer; `readPoint()` had the same unchecked indexing. The flatbuffer is not verified, so those sizes are file-controlled.

The reader now records the available coordinate-pair count when `xy` is read and rejects any ring range that does not fit it (and any `z`/`m` array shorter than the requested points) before allocating or copying. Keeping the check in `readLineObj`/`readPoint` covers point, line, polygon and the multipoint/multiline/multipolygon variants that all funnel through the same code.

Same neighbourhood as #7545, which bounded the property blob; this covers the geometry-decoding side.

## AI tool usage

## Tasklist

 - [ ] Add test case(s) in `/msautotest` (follow steps in [Regression Testing](https://mapserver.org/development/tests/autotest.html))
 - [x] Add a regression test (`tests/unit/test.cpp`: decodes a polygon whose ring end exceeds its coordinate count; fails under ASan before the fix, passes after)
 - [ ] Add documentation
 - [ ] Review
 - [ ] Adjust for comments
 - [x] All CI builds and checks have passed

### DIFF
--- a/src/flatgeobuf/geometryreader.cpp
+++ b/src/flatgeobuf/geometryreader.cpp
@@ -5,14 +5,21 @@ using namespace FlatGeobuf;
 
 void GeometryReader::readPoint(shapeObj *shape)
 {
+    if (m_offset >= m_xy_count) {
+        msSetError(MS_FGBERR,
+                   "Corrupt FlatGeobuf geometry: point offset out of bounds",
+                   "GeometryReader::readPoint");
+        return;
+    }
+
     lineObj *l = (lineObj *) malloc(sizeof(lineObj));
     pointObj *p = (pointObj *) malloc(sizeof(pointObj));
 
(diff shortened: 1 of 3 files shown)


jmckenna:
cc @bjornharrtell

